PR #18, opened 2018-05-23
010-astarte_in_5_minutes: add trigger installation

rbino:


doc/pages/tutorials/010-astarte_in_5_minutes.md
```diff
@@ -76,6 +76,23 @@ $ curl -X GET http://localhost:4000/v1/test/interfaces -H "Authorization: Bearer
 
 Our Astarte instance is now ready for our devices.
 
+## Install a trigger
+
+We will also test Astarte's push capabilities with a trigger. This will send a POST to a URL of our choice every time that the value generated by `stream_test` is above 0.6.
```
drf: "the value" rather than "that the value"

drf: Add a paragraph:

"Due to how triggers work, it is fundamental to install the trigger before a device connects. Doing otherwise will cause the trigger to kick in at a later time, and as such no events will be streamed for a while"

PR #59, opened 2019-01-30
010-astarte_in_5_minutes: fix Python JWT package

rbino:
There are two packages in pip: jwt and PyJWT. The second one is the one we want.


rbino:
If you install both the pip package and the system version it works just fine, the conflict was between `apt-get install python3-jwt` (which installs PyJWT) and `pip3 install jwt` (which installs `jwt`, that is another thing)

rbino:
I'll add a troubleshooting part in case the user already has `jwt` installed

PR #73, opened 2019-05-31
Add astarte_import tool

bettio:


tools/astarte_import/lib/astarte/import.ex
```diff
@@ -0,0 +1,136 @@
...
+         state
+       ) do
+    state
+  end
+
+  defp xml_event({:startElement, _uri, _l_name, {_prefix, 'interface'}, attributes}, _loc, state) do
+    with {:ok, name} <- get_attribute(attributes, 'name'),
+         {:ok, major_string} <- get_attribute(attributes, 'version_major'),
+         {major, ""} <- Integer.parse(major_string),
+         {:ok, minor_string} <- get_attribute(attributes, 'version_minor'),
```
rbino: Use `minor_version` to match how we store things in the `Astarte.Core.Interface` struct

tools/astarte_import/lib/astarte/import.ex
```diff
@@ -0,0 +1,136 @@
...
+
+  defp xml_event(:startDocument, _location, state) do
+    state
+  end
+
+  defp xml_event(:endDocument, _location, state) do
+    state
+  end
+
+  defp get_attribute(attributes, attribute_name) do
```
rbino: Since this returns `{:ok, result}`/`{:error, reason}` it should probably be called `fetch_attribute`

tools/astarte_import/lib/astarte/import.ex
```diff
@@ -0,0 +1,136 @@
...
+    attribute_tuple =
+      Enum.find(attributes, fn attribute ->
+        case attribute do
```
rbino: You can pattern match directly on the `fn` head since you're using `case` as outermost expression:
```
... fn
{_uri, _prefix, ^attribute_name, _attribute_value} -> true
_ -> false
end)
```

tools/astarte_import/lib/astarte/import.ex
```diff
@@ -0,0 +1,136 @@
...
+          _ ->
+            false
+        end
+      end)
+
+    with {_uri, _prefix, _attribute_name, attribute_value} <- attribute_tuple do
+      {:ok, to_string(attribute_value)}
+    else
+      nil ->
+        {:error, :missing_attribute, attribute_name}
```
rbino: Wrap the "reason" in a tuple (`{:error, {:missing_attribute, attribute_name}}`) so that we're consistent with Elixir's APIs

tools/astarte_import/lib/astarte/import.ex
```diff
@@ -0,0 +1,136 @@
...
+  defp xml_event(:endDocument, _location, state) do
+    state
+  end
+
+  defp get_attribute(attributes, attribute_name) do
+    attribute_tuple =
+      Enum.find(attributes, fn attribute ->
+        case attribute do
+          {_uri, _prefix, ^attribute_name, _attribute_value} ->
+            true
```
rbino: If you return `attribute_value` instead of `true` you can use `Enum.find_value` so you don't have to match on the tuple again below

PR #77, opened 2019-06-13
make astarte_import usable

bettio:


tools/astarte_import/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# Set the locale
+ENV LANG C.UTF-8
+
+# We need SSL
+RUN apt-get -qq install libssl1.1
+
+WORKDIR /app
+COPY --from=builder /app/_build/prod/rel/astarte_import .
+
+#CMD ["./bin/astarte_import", "eval", "Astarte.Import.CLI.main()", ${realm}, ${file}]
```
rbino: Remove this

tools/astarte_import/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+ENV REALM "default"
```
rbino: I'd remove the default realm and raise in `main` if the realm is not passed

tools/astarte_import/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM elixir:1.8.2-slim as builder
```
drf: Can we take this chance to move away from slim images in builders? They make compilation time longer due to more packages being installed, and they don't make any sense as they're only builders.

tools/astarte_import/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+WORKDIR /app
+COPY --from=builder /app/_build/prod/rel/astarte_import .
+
+#CMD ["./bin/astarte_import", "eval", "Astarte.Import.CLI.main()", ${realm}, ${file}]
+
+ENV REALM "default"
+ENV XML_FILE "/files/data.xml"
+
+CMD ["sh", "-c", "./bin/astarte_import eval \"Astarte.Import.CLI.main()\" ${REALM} ${XML_FILE}"]
```
drf: I would consider an ENTRYPOINT with a bash script here, especially in case we need more complicated logic in the future and/or pass custom stuff to the command.

tools/astarte_import/rel/plugins/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.*
+!*.exs
+!.gitignore
```
drf: newline

tools/astarte_import/rel/config.exs
```diff
@@ -0,0 +1,62 @@
...
+# when running `mix release`, the first release in the file
+# will be used by default
+
+release :astarte_import do
+  set version: current_version(:astarte_import)
+  set applications: [
+    :runtime_tools
+  ]
+end
+
```
drf: extraline

PR #92, opened 2019-11-12
doc/architecture: remove has_metadata

bettio:
has_metadata is not implemented and not planned yet.


drf:
Fine with this, but let's add a ticket to the board first.

Issue #86, opened 2019-11-05
Provide a set of ready to use interfaces


bettio:
Astarte should provide a set of ready to install and user interfaces that can be used as an example and can be without further effort:

```
{
    "interface_name": "org.astarte-platform.standard-interfaces.GenericSensor",
    "version_major": 1,
    "version_minor": 0,
    "type": "datastream",
    "ownership": "device",
    "mappings": [
        {
            "endpoint": "/%{sensor_id}/value",
            "type": "double"
        }
    ]
}
```

```{
    "interface_name": "org.astarte-platform.standard-interfaces.AvailableSensors",
    "version_major": 1,
    "version_minor": 0,
    "type": "property",
    "ownership": "device",
    "mappings": [
        {
            "endpoint": "/%{sensor_id}/name",
            "type": "string"
        },
        {
            "endpoint": "/%{sensor_id}/unit",
            "type": "string"
        }
    ]
}
```

PR #95, opened 2019-11-27
Add some standard interfaces

bettio:
Provide some ready to use interfaces.

Fix https://github.com/astarte-platform/astarte/issues/86

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [pairing_api] Add Prometheus instrumenters and exporters
 - [realm_management_api] Add Prometheus instrumenters and exporters
 - [housekeeping_api] Add Prometheus instrumenters and exporters
+- Add standard interfaces for generic sensors.
 
 ### Changed
 - [realm_management] Handle hyphens in `interface_name`. ([#96](https://github.com/astarte-platform/astarte/issues/96))
--- a/standard-interfaces/org.astarte-platform.genericsensors.AvailableSensors.json
+++ b/standard-interfaces/org.astarte-platform.genericsensors.AvailableSensors.json
@@ -1,0 +1,23 @@
+{
+    "interface_name": "org.astarte-platform.genericsensors.AvailableSensors",
+    "version_major": 0,
+    "version_minor": 1,
+    "type": "properties",
+    "ownership": "device",
+    "description": "Describes available generic sensors.",
+    "doc": "This interface allows to describe available sensors and their attributes such as name and sampled data measurement unit. Sensors are identified by their sensor_id. See also org.astarte-platform.genericsensors.AvailableSensors.",
+    "mappings": [
+        {
+            "endpoint": "/%{sensor_id}/name",
+            "type": "string",
+            "description": "Sensor name.",
+            "doc": "An arbitrary sensor name."
+        },
+        {
+            "endpoint": "/%{sensor_id}/unit",
+            "type": "string",
+            "description": "Sample data measurement unit.",
+            "doc": "SI unit such as m, kg, K, etc..."
+        }
+    ]
+}
--- a/standard-interfaces/org.astarte-platform.genericsensors.SamplingRate.json
+++ b/standard-interfaces/org.astarte-platform.genericsensors.SamplingRate.json
@@ -1,0 +1,25 @@
+{
+    "interface_name": "org.astarte-platform.genericsensors.SamplingRate",
+    "version_major": 0,
+    "version_minor": 1,
+    "type": "properties",
+    "ownership": "server",
+    "description": "Configure sensors sampling rate and enable/disable.",
+    "doc": "This interface allows to set generic sensors sampling rate and enable/disable policies for each sensor. Sensors are identified by their sensor_id. See also org.astarte-platform.genericsensors.AvailableSensors.",
+    "mappings": [
+        {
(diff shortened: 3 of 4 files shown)


standard-interfaces/cloud.astarte.SamplingRate.json
```diff
@@ -0,0 +1,22 @@
...
+            "type": "boolean",
+            "allow_unset": true,
+            "description": "Enable/disable sensor data transmission.",
+            "doc": "When true sampled data transmission is always on, otherwise when false is off. When unset data transmission policy is up to the sensor."
+        },
+        {
+            "endpoint": "/%{sensor_id}/samplingPeriod",
+            "type": "integer",
+            "description": "Sensor sample transmission period.",
+            "doc": "Send a sampled value every samplingPeriod seconds. When unset sampling period is up to the sensor."
```
rbino: `allow_unset` is missing here

standard-interfaces/cloud.astarte.GenericSensors.json
```diff
@@ -0,0 +1,13 @@
+{
+    "interface_name": "cloud.astarte.GenericSensors",
+    "version_major": 0,
+    "version_minor": 1,
+    "type": "datastream",
+    "ownership": "device",
+    "mappings": [
+        {
+            "endpoint": "/%{sensor_id}/value",
+            "type": "double"
```
rbino: Missing description and documentation

standard-interfaces/cloud.astarte.GenericSensors.json
```diff
@@ -0,0 +1,13 @@
+{
+    "interface_name": "cloud.astarte.GenericSensors",
```
rbino: cloud.astarte.genericsensors.Values

standard-interfaces/cloud.astarte.AvailableSensors.json
```diff
@@ -0,0 +1,20 @@
+{
+    "interface_name": "cloud.astarte.AvailableSensors",
```
rbino: cloud.astarte.genericsensors.AvailableSensors

standard-interfaces/cloud.astarte.SamplingRate.json
```diff
@@ -0,0 +1,22 @@
+{
+    "interface_name": "cloud.astarte.SamplingRate",
```
rbino: cloud.astarte.genericsensors.SamplingRate

drf:
Shouldn't we use org.astarte-platform as the prefix?

standard-interfaces/cloud.astarte.GenericSensors.json
drf: This should be strongly typed, and we should have a set. E.g.: GenericDoubleSensors

Otherwise (which I believe could be even better), we should have several strongly typed endpoints (e.g.: /doubleValue, /intValue)

bettio: At the end of the day sensors measure real values, and every integer or long integer can be safely converted to a double (with a minor loss of precision).
Measures have also an amount of error/noise, so I think that we don't need too much precision.
Special requirements might be covered with new interfaces (or custom ones).

bettio:
> Shouldn't we use org.astarte-platform as the prefix?

I agree, we need to fix https://github.com/astarte-platform/astarte/issues/96 before.

standard-interfaces/cloud.astarte.genericsensors.SamplingRate.json
```diff
@@ -0,0 +1,25 @@
...
+            "allow_unset": true,
+            "description": "Enable/disable sensor data transmission.",
+            "doc": "When true sampled data transmission is always on, otherwise when false is off. When unset data transmission policy is up to the sensor."
+        },
+        {
+            "endpoint": "/%{sensor_id}/samplingPeriod",
+            "type": "integer",
+            "allow_unset": true,
+            "description": "Sensor sample transmission period.",
+            "doc": "Send a sampled value every samplingPeriod seconds. When unset sampling period is up to the sensor."
```
drf: probably millisecond and longinteger?

bettio: We are trying to encourage people to use sane sampling rates, such as 5 seconds, instead of 5 milliseconds. If this doesn't fit some user requirements the other 2 interfaces are still usable and this interface can be replaced with a custom one.

standard-interfaces/cloud.astarte.genericsensors.Values.json
```diff
@@ -0,0 +1,17 @@
+{
+    "interface_name": "cloud.astarte.genericsensors.Values",
+    "version_major": 0,
+    "version_minor": 1,
+    "type": "datastream",
+    "ownership": "device",
+    "description": "Generic sensors sampled data.",
+    "doc": "This interface reports sensors sampled data datastreams. Each datastream is idenitified using a sensor_id.",
```
drf: > This interface reports sensors sampled data datastreams

wtf.

Something like "Values allows generic sensors to stream samples. It is usually used in combination with AvailableSensors, which makes the device aware of what sensors and what unit of measure they are reporting. sensor_id represents an unique identifier for an individual sensor, and should match sensor_id in AvailableSensors when used in combination."

bettio: Ok, however I've changed `which makes the device aware ` to `which makes API client aware `.

PR #127, opened 2019-12-20
docs: More tools deprecations

drf:
Basically what was left out of the previous PR


rbino:
Since we're using astartectl in the examples now I would add a brief section about installing it (potentially just linking at the appropriate section of the README)

rbino:
Also, while we're at it, I'd use `astartectl` also for the key generation, since we had problems with OpenSSL on some platforms in the past.
The two openssl lines can be replaced with
```
astartectl utils gen-keypair test
```
just make sure to adjust the keyname around (`test_realm.key -> test_private.pem` and `test_realm.key.pub -> test_public.pem`)

PR #183, opened 2020-01-10
mix.exs: disable autogenerated API page

Pavinati:
By default ex_doc generates a "API refenrece" page. This will remove it and its relative entry from the sidebar menu


Pavinati:
> Please, open it against release-0.11 (not master).

Rebased from release-0.11

Issue #96, opened 2019-11-28
[meta] org.astarte-platform.InterfaceName is not a valid interface name


bettio:
`-` are not allowed in interface names right now, they should be allowed if we want to allow `org.astarte-platform.InterfaceName`.

I suggest following change:
- `-` is allowed in any domain components (e.g. ` org.astarte-platform`).
- `-` is disallowed in the last interface name token (e.g.  `InterfaceName`).

This will keep the interface name programmer friendly (e.g. interface name can be always mapped to a class or module name), while this limitation might be removed in future releases.

Related issues:
- [x] https://github.com/astarte-platform/astarte_core/issues/7
- [x] https://github.com/astarte-platform/astarte-dashboard/issues/7

bettio:
Fixed.

PR #188, opened 2020-01-16
Handle interface name to table name

matt-mazzucato:
Given the possibility to include hyphens into interface_name, handle the mapping of interface_name to table name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [0.11.0-beta.2] - Unreleased
+### Changed
+- [realm_management] Handle hyphens in `interface_name`. ([#96](https://github.com/astarte-platform/astarte/issues/96))
 
 ## [0.11.0-beta.1] - 2019-12-26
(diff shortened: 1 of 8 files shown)


matt-mazzucato:
Address #96

zanettea:
If hyphens are simply replaced with spaces, then the poor user will end up striving to understand why the service refuses to install two interfaces with different names saying the interface already exists.

Either you should improve the error or (better imho) provide a transparent mapping which does not fail in corner cases.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [0.11.0-beta.2] - Unreleased
+### Changed
+- [realm_management] Handle hyphens in `interface_name`. ([96](https://github.com/astarte-platform/astarte/issues/96))
```
rbino: Missing # before the issue number

matt-mazzucato:
@zanettea Thanks for the input. I improved the error message.
Unfortunately it's not possible to employ a bijective function while mapping interface_name to table name: this is due to Cassandra's constraints as discussed in [this issue](https://github.com/astarte-platform/astarte_core/issues/7#issuecomment-567896895)

Issue #192, opened 2020-01-17
Add /metrics to Astarte Housekeeping API


matt-mazzucato:
Add prometheus /metrics endpoint.

PR #193, opened 2020-01-17
Add Prometheus instrumenters and exporters

matt-mazzucato:
Add Prometheus instrumenters and exporters to housekeeping api


matt-mazzucato:
Address #192

PR #246, opened 2020-02-18
standard-interfaces: add missing explicit_timestamp

bettio:
Set org.astarte-platform.genericsensors.Values.json/%{sensor_id}/value
explicit_timestamp to true.


drf:
@bettio target 0.11 please

PR #247, opened 2020-02-19
Added React example for Astarte Channel. 

### LINKED ISSUE #102: Implement sensors viewer WebSocket + JS example
Implement a JavaScript example which allows to choose a sensor and to display real time data using a WebSocket.

The example should use `cloud.astarte.genericsensors.Values` properties interface and `cloud.astarte.genericsensors.AvailableSensors` (when available). It would be useful to implement it using React as well.

See also https://github.com/astarte-platform/astarte/issues/86

shubham11107:
Use WebSocket to connect Astarte Channel Displaying device status,
real time status and data 
Only "org.astarte-platform.genericsensors.Values" interface is used
Fix #102

Signed-off-by: shubham wadhwa <shubham.wadhwa@ispirata.com>

### DIFF
--- a/examples/sensor-channels/package.json
+++ b/examples/sensor-channels/package.json
@@ -1,0 +1,40 @@
+{
+  "name": "sensor-channels",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "axios": "^0.19.2",
+    "bootstrap": "^4.4.1",
+    "named-urls": "^2.0.0",
+    "phoenix": "^1.4.12",
+    "react": "^16.12.0",
+    "react-bootstrap": "^1.0.1",
+    "react-dom": "^16.12.0",
+    "react-scripts": "3.4.1"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject",
+    "format-js": "prettier --write \"src/**/*.js\""
+  },
+  "eslintConfig": {
+    "extends": "react-app"
+  },
+  "browserslist": {
+    "production": [
+      ">0.2%",
+      "not dead",
+      "not op_mini all"
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
+  },
+  "devDependencies": {
+    "prettier": "^1.19.1"
+  }
+}
--- a/examples/sensor-channels/src/components/SensorDataHandler.jsx
+++ b/examples/sensor-channels/src/components/SensorDataHandler.jsx
@@ -1,0 +1,108 @@
+import React, { Component } from "react";
+
+import { SensorData } from "./SensorData";
+
+function getLocaleFormat(timestamp) {
+  const event = new Date(timestamp);
+  const locale = window.navigator.language;
+  return event.toLocaleString(locale);
+}
+
+class SensorDataHandler extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      sensors: {},
+      device: {
+        connection_type: false
+      },
+      connected: false,
+      alerts: []
+    };
+  }
+
+  setDeviceStatus = () => {
+    const { device, astarte } = this.props;
+    astarte.getDeviceDataById(device).then(response => {
+      this.setState({ device: { connection_type: response.connected } });
+    });
+  };
+
+  onInComingData = response => {
+    if (response.event === "new_event") {
+      const { sensors } = this.state;
+      const time = response.payload.timestamp;
+      const event = response.payload.event;
+      if (event.path) {
+        sensors[event.path] = {
+          value: event.value,
+          // We have path in event '/${device_name}/value'
+          // We need extract 'device_name' from path
+          // So we skip first "/" using event.path.split("/")[1]
+          name: event.path.split("/")[1],
+          timestamp: getLocaleFormat(time)
+        };
+      }
+      this.setState({ sensors });
+    }
+  };
+
+  onSocketError = () => {
+    const { alerts } = this.state;
+    alerts.unshift({ type: "warning", msg: "An error occurred" });
+    this.setState({ alert });
+  };
+
+  onSocketClose = () => {
+    const { alerts } = this.state;
+    this.onDisconnectHandler();
+    alerts.unshift({ type: "info", msg: "Closed connection to Astarte." });
+    this.setState({ alert });
+  };
+
+  onSocketOpen = () => {
+    const { alerts } = this.state;
+    alerts.unshift({ type: "success", msg: "Open connection to Astarte." });
+    this.setState({ alert });
+  };
+
+  setConnected(value) {
+    const { device, astarte } = this.props;
+    if (value) {
+      astarte.connectSocket({
+        device,
+        interfaceName: "org.astarte-platform.genericsensors.Values",
+        onInComingData: this.onInComingData,
+        onOpenConnection: this.onSocketOpen,
+        onCloseConnection: this.onSocketClose,
+        onErrorConnection: this.onSocketError
+      });
+    } else {
+      astarte.disconnectSocket();
+    }
+  }
+
+  onConnectHandler() {
+    this.setConnected(true);
+  }
+
+  onDisconnectHandler() {
+    this.setConnected(false);
+  }
+
+  componentDidMount() {
+    this.onConnectHandler();
+    this.setDeviceStatus();
+  }
+
+  componentWillUnmount() {
+    this.onDisconnectHandler();
+  }
+
+  render() {
+    const { sensors, device, alerts } = this.state;
+    return <SensorData sensors={sensors} device={device} alerts={alerts} />;
+  }
+}
+
+export default SensorDataHandler;
--- a/examples/sensor-channels/package-lock.json
+++ b/examples/sensor-channels/package-lock.json
@@ -1,0 +1,13819 @@
+{
+  "name": "sensor-channels",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
+      "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
+      "requires": {
+        "@babel/highlight": "^7.8.3"
+      }
+    },
+    "@babel/compat-data": {
+      "version": "7.9.6",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.9.6.tgz",
+      "integrity": "sha512-5QPTrNen2bm7RBc7dsOmcA5hbrS4O2Vhmk5XOL4zWW/zD/hV0iinpefDlkm+tBBy8kDtFaaeEvmAqt+nURAV2g==",
+      "requires": {
+        "browserslist": "^4.11.1",
+        "invariant": "^2.2.4",
+        "semver": "^5.5.0"
+      },
+      "dependencies": {
+        "semver": {
+          "version": "5.7.1",
+          "resolved": "https://registry.npmjs.org/semver/-/semver-5.7.1.tgz",
+          "integrity": "sha512-sauaDf/PZdVgrLTNYHRtpXa1iRiKcaebiKQ1BJdpQlWH2lCvexQdX55snPFyK7QzpudqbCI0qXFfOasHdyNDGQ=="
+        }
+      }
+    },
+    "@babel/core": {
+      "version": "7.9.0",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.9.0.tgz",
+      "integrity": "sha512-kWc7L0fw1xwvI0zi8OKVBuxRVefwGOrKSQMvrQ3dW+bIIavBY3/NpXmpjMy7bQnLgwgzWQZ8TlM57YHpHNHz4w==",
+      "requires": {
+        "@babel/code-frame": "^7.8.3",
+        "@babel/generator": "^7.9.0",
+        "@babel/helper-module-transforms": "^7.9.0",
+        "@babel/helpers": "^7.9.0",
+        "@babel/parser": "^7.9.0",
+        "@babel/template": "^7.8.6",
+        "@babel/traverse": "^7.9.0",
+        "@babel/types": "^7.9.0",
+        "convert-source-map": "^1.7.0",
+        "debug": "^4.1.0",
+        "gensync": "^1.0.0-beta.1",
+        "json5": "^2.1.2",
+        "lodash": "^4.17.13",
+        "resolve": "^1.3.2",
+        "semver": "^5.4.1",
+        "source-map": "^0.5.0"
+      },
+      "dependencies": {
+        "debug": {
+          "version": "4.1.1",
+          "resolved": "https://registry.npmjs.org/debug/-/debug-4.1.1.tgz",
+          "integrity": "sha512-pYAIzeRo8J6KPEaJ0VWOh5Pzkbw/RetuzehGM7QRRX5he4fPHx2rdKMB256ehJCkX+XRQm16eZLqLNS8RSZXZw==",
+          "requires": {
+            "ms": "^2.1.1"
+          }
+        },
+        "ms": {
+          "version": "2.1.2",
+          "resolved": "https://registry.npmjs.org/ms/-/ms-2.1.2.tgz",
+          "integrity": "sha512-sGkPx+VjMtmA6MX27oA4FBFELFCZZ4S4XqeGOXCv68tT+jb3vk/RyaKWP0PTKyWtmLSM0b+adUTEvbs1PEaH2w=="
+        },
+        "semver": {
+          "version": "5.7.1",
+          "resolved": "https://registry.npmjs.org/semver/-/semver-5.7.1.tgz",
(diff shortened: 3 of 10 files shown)


examples/sensor-websocket/src/components/SensorWebSocket.jsx
```diff
@@ -0,0 +1,187 @@
...
+const socketDataHandler = response => {
+  if (response.event === "new_event") {
+    const { sensors } = this.state;
+    const time = response.payload.timestamp;
+    const event = response.payload.event;
+    if (!event.path) {
+      this.setState({
+        device: {
+          connection_type: event.type,
+          timestamp: moment(time).format("MMMM Do YYYY, h:mm:ss a")
```
Pavinati: let's use locale aware time formats

examples/sensor-websocket/src/components/SensorWebSocket.jsx
```diff
@@ -0,0 +1,187 @@
...
+        }
+      });
+    } else {
+      sensors[event.path] = {
+        value: event.value,
+        name: event.path.split("/")[1],
```
Pavinati: this is pretty unclear for people reading this for the first time.
add a comment explaining why you need the second item of that split list, or rewrite this.

This is what is happening behind the curtains
`"/${device_id}/value".split("/") => [ "", "${device_id}", "value" ]`

examples/sensor-websocket/src/components/SensorWebSocket.jsx
```diff
@@ -0,0 +1,187 @@
...
+    </Container>
+  );
+};
+
+const errorSocket = () => {
+  return <p>Socket is giving an Error</p>;
+};
+
+const closeSocket = () => {
+  return <p>Socket is Closed</p>;
```
Pavinati: The user doesn't, and shoudn't know that the data is retrieved using a socket. All the user cares is getting his device info. Let's show meaningful error messages when we can't accomplish  what the user asked

examples/sensor-websocket/src/components/SensorWebSocket.jsx
```diff
@@ -0,0 +1,206 @@
...
+  socketDataHandler = response => {
+    if (response.event === "new_event") {
+      const { sensors } = this.state;
+      const time = response.payload.timestamp;
+      const event = response.payload.event;
+      if (event.path) {
+        sensors[event.path] = {
+          value: event.value,
+          name: event.path.split("/")[1],
+          //"/${device_id}/value".split("/") => [ "", "${device_id}", "value" ]
```
Pavinati: this is a copy paste from my previous comment.
Make this developer friendly.

You can find more on that topic with a simple Google search "code comments best practices", add "javascript" for more specific examples
https://blog.risingstack.com/javascript-clean-coding-best-practices-node-js-at-scale/

examples/sensor-websocket/src/components/SensorWebSocket.jsx
```diff
@@ -0,0 +1,206 @@
...
+    phoenixSocket.onClose(this.onSocketClose);
+    phoenixSocket.onOpen(this.onSocketOpen);
+    phoenixSocket.onMessage(this.socketDataHandler);
+    phoenixSocket.connect();
+    this.channelConnect(phoenixSocket, deviceID);
+  };
+
+  channelConnect = (phoenixSocket, deviceID) => {
+    const channel_token = getChannelToken();
+    const channel = phoenixSocket.channel("rooms:example:test", {
```
Pavinati: if all your rooms have the same name, when you open the same app in 2 browser tabs, all the triggers will interfere with each other, and of course this is not intended for this particular App.

examples/sensor-websocket/src/components/SensorWebSocket.jsx
```diff
@@ -0,0 +1,206 @@
...
+  getDeviceDataById,
+  getWSUrl,
+  socketParams,
+  value_trigger_payload
+} from "../apiHandler";
+
+import { Alert, Col, Container, ListGroup, Row } from "react-bootstrap";
+import moment from "moment";
+
+require("moment/min/locales.min");
```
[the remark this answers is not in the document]
shubham11107: We use because we need to find user language, 
according to your reference `event.toLocaleString('en-GB', { timeZone: 'UTC' })`
How I will get 'en-GB' because I think we need to dynamic, that's why I use above library for this to get user language and make locale datetime format in that language.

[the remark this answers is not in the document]
Pavinati: What about navigator.language?
https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language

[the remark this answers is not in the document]
shubham11107: Yes, We are using navigator.language.
This is the code 
```
require("moment/min/locales.min");

function getLocaleFormat() {
    const locale = window.navigator.language;
    moment.locale(locale);
    return moment.localeData().longDateFormat("LLL");
}
```
In Moment js, we can't update locale without this js `moment/min/locales.min` 
mention in docs https://momentjs.com/docs/#/use-it/browserify/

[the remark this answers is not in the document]
Pavinati: Then you have the answer to your previous question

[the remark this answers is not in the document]
shubham11107: Yes, I'm already using this in all examples.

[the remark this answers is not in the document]
Pavinati: How's that required when you can simply do as follows?
```
const locale = window.navigator.language;
event.toLocaleString(locale, { timeZone: 'UTC' })
```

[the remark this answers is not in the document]
Pavinati: You can also retrieve the timezone by the way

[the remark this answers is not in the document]
shubham11107: Okay let me do this.

.gitignore
```diff
@@ -44,5 +44,6 @@ deps
 .rebar3
 apps/*/_checkouts/
 
+# Sensor Example
```
Pavinati: This comment can be removed

examples/sensor-websocket/src/apiHandler.js
```diff
@@ -0,0 +1,203 @@
...
+  GET(url, params) {
+    return axios.get(url, { headers: this.getHeaders(), params: params });
+  }
+
+  PUT(url, params) {
+    return axios.put(url, params, { headers: this.getHeaders() });
+  }
+
+  isDeviceId(value) {
+    const expression = new RegExp(/[a-z]?[A-Z]?[0-9]?-?_?/i);
```
Pavinati: This regexp is no good.
- You have lower case and upper case groups, but then you add the case insensitive flag ( `/i` ). Chose one
- Groups added like that match a sequence.
- To make sure you match the whole string, you must match start to finish `/^ ... $/`

examples/sensor-websocket/src/components/CredentialsModal.jsx
```diff
@@ -0,0 +1,104 @@
+import React, { Component } from "react";
+import { Col, Form, Modal } from "react-bootstrap";
+
+class CredentialsModal extends Component {
```
Pavinati: when extending react components you must make sure to init the internal component too
```
class MyClass extends React.Component {
  constructor(props) {
    super(props);

    // my init here
  }

  render() {
  }
}
```

examples/sensor-websocket/src/components/CredentialsModal.jsx
```diff
@@ -0,0 +1,104 @@
...
+              Submit
+            </button>
+          </Form>
+        </Modal.Body>
+      </Modal>
+    );
+  }
+}
+
+export default CredentialsModal;
```
Pavinati: missing new line

examples/sensor-websocket/src/components/SensorData.jsx
```diff
@@ -0,0 +1,84 @@
...
+      {alerts.map((alert, index) => {
+        return (
+          <Alert key={index} variant={alert.type}>
+            {alerts.length - index}. {alert.msg}
+          </Alert>
+        );
+      })}
+    </Container>
+  );
+}
```
Pavinati: Missing new line

examples/sensor-websocket/src/components/SensorDataHandler.jsx
```diff
@@ -0,0 +1,104 @@
...
+    this.onDisconnectHandler();
+  }
+
+  render() {
+    const { sensors, device, alerts } = this.state;
+    return <SensorData sensors={sensors} device={device} alerts={alerts} />;
+  }
+}
+
+export default SensorDataHandler;
```
Pavinati: Missing new line

examples/sensor-websocket/src/components/SensorDisplay.jsx
```diff
@@ -0,0 +1,146 @@
...
+        <CredentialsModal
+          setCredentials={this.setCredentials}
+          visible={visible}
+        />
+      </Container>
+    );
+  }
+}
+
+export default SensorDisplay;
```
Pavinati: missing new line

examples/sensor-websocket/src/apiHandler.js
```diff
@@ -0,0 +1,203 @@
...
+    return axios.get(url, { headers: this.getHeaders(), params: params });
+  }
+
+  PUT(url, params) {
+    return axios.put(url, params, { headers: this.getHeaders() });
+  }
+
+  isDeviceId(value) {
+    const expression = new RegExp(/^[a-zA-Z0-9-_]{22}$/g);
+    return value.length === 22 && expression.test(value)
```
Pavinati: the previous regular expression already checks for length, no need to check it again

examples/sensor-websocket/src/apiHandler.js
```diff
@@ -0,0 +1,203 @@
...
+    const URL = this.getAPIUrl("interface_id_path", {
+      device_id: id,
+      interface: interfaces,
+      sensor_id: sensor_id,
+      key: key
+    });
+    return this.GET(URL, params);
+  }
+
+  setSensorSamplingById(id, interfaces, sensor_id, key, params = {}) {
```
Pavinati: sensor sampling?

examples/sensor-websocket/src/components/SensorData.jsx
```diff
@@ -0,0 +1,84 @@
+import React from "react";
+import { Alert, Col, Container, ListGroup, Row } from "react-bootstrap";
+
+function ListItem(props) {
+  return (
+    <ListGroup.Item
+      key={props.index}
+      className="temperature-list px-3 border-0 pb-0 py-2 font-weight-normal bg-transparent"
+    >
+      <b className="font-weight-bold">{props.label}: </b>
```
Pavinati: `<b>` stands for bold, the class is not needed

examples/sensor-websocket/src/components/SensorData.jsx
```diff
@@ -0,0 +1,84 @@
+import React from "react";
+import { Alert, Col, Container, ListGroup, Row } from "react-bootstrap";
+
+function ListItem(props) {
+  return (
+    <ListGroup.Item
+      key={props.index}
+      className="temperature-list px-3 border-0 pb-0 py-2 font-weight-normal bg-transparent"
```
Pavinati: forcing a font weight here is not a good idea. It makes ALL children components inherit this style.
Use it only in the components that require it

examples/sensor-websocket/src/components/SensorData.jsx
```diff
@@ -0,0 +1,84 @@
...
+    {
+      label: "Last Update",
+      value: sensor.timestamp
+    }
+  ];
+  return (
+    <Row key={index} className="m-3 bg-light px-2 pt-3 pb-2">
+      <Col className="text-dark px-3 py-2 bg-transparent border-0 font-weight-bold w-100 text-left btn text-decoration-none">
+        <span className="bg-sensor-theme font-weight-light py-2 px-3 w-100 text-white d-inline-block">
+          <strong className="font-weight-bold">Sensor Name : &nbsp;</strong>
```
Pavinati: same as the previous comment. `<strong>` is not a the right tag for a label

Pavinati:
```
47 packages are looking for funding
  run `npm fund` for details

found 890 vulnerabilities (872 low, 18 moderate)
```
You should update the dependencies of this project. Most of them will probably fix themselves

examples/sensor-websocket/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "sensor-websocket",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "axios": "^0.19.2",
+    "bootstrap": "^4.4.1",
+    "moment": "^2.24.0",
```
Pavinati: is "moment" library still used?

examples/sensor-websocket/src/components/SensorData.jsx
```diff
@@ -0,0 +1,83 @@
...
+      value: sensor.value
+    },
+    {
+      label: "Last Update",
+      value: sensor.timestamp
+    }
+  ];
+  return (
+    <Row key={index} className="m-3 bg-light px-2 pt-3 pb-2">
+      <Col className="text-dark px-3 py-2 bg-transparent border-0 font-weight-bold w-100 text-left btn text-decoration-none">
```
Pavinati: font-weight-bold on a column?

examples/sensor-websocket/src/components/SensorData.jsx
```diff
@@ -0,0 +1,83 @@
...
+        <span className="bg-sensor-theme font-weight-light py-2 px-3 w-100 text-white d-inline-block">
+          <p className="font-weight-bold">Sensor Name : {sensor.name}</p>
+        </span>
+        <Col className="p-0">
```
Pavinati: here you have a Col inside an other Col, that shouldn't happen when the layout classes are used properly.

examples/sensor-websocket/src/components/SensorDisplay.jsx
```diff
@@ -0,0 +1,150 @@
...
+      submit: true
+    });
+  }
+
+  handleError(err) {
+    if (err.response.status === 403)
+      this.setState({ visible: true, loading: false });
+    if (err.response.status === 404) {
+      this.setState({ loading: false });
+      window.confirm("Device ID Invalid");
```
Pavinati: 404 is "not found". Invalid request is bound to error code 400

shubham11107: When we hit API with wrong Device ID it returns 404 status code

examples/sensor-websocket/src/components/SensorData.jsx
shubham11107: I need to test properly with data but right now no device streaming

shubham11107: Done

examples/sensor-websocket/src/components/SensorDisplay.jsx
shubham11107: let me know about this

Pavinati: not really. take for example `AAAAAAAAAAAAAAAAAAAAAA` is a valid device ID. If your realm contains a device with such a device ID, the API will return 200 (Ok), otherwise 404 (Not found).
On the other hand, `fakeid` is not a valid ID, so the API will always return 400 (bad request)

examples/sensor-channels/src/components/SensorDataHandler.jsx
```diff
@@ -0,0 +1,107 @@
...
+    if (response.event === "new_event") {
+      const { sensors } = this.state;
+      const time = response.payload.timestamp;
+      const event = response.payload.event;
+      if (event.path) {
+        sensors[event.path] = {
+          value: event.value,
+          // event.path => '/${device_name}/value'
+          // name = event.path.split("/")[1]
+          name: event.path.split("/")[1],
```
Pavinati, line 41: this expression is a bit cryptic.
Normally one would think it takes the second element (because indexes are 0 based) of the path, but if the path starts with `/` the returned array will contain an empty element, and the function will return the first group in the path.

Or add a comment explaining the meaning behind this one (it's meant to skip the first empty element returned considering the path starts with a `/`)

examples/sensor-channels/src/components/SensorDataHandler.jsx
```diff
@@ -0,0 +1,107 @@
...
+          timestamp: getLocaleFormat(time)
+        };
+      }
+      this.setState({ sensors });
+    }
+  };
+
+  onSocketError = () => {
+    const { alerts } = this.state;
+    alerts.unshift({ type: "info", msg: "An error occurred" });
```
Pavinati, line 51: It would be nice to see which error occurred. Also, you can use warning styling. 'Info` is not really suited for errors

shubham11107: for print error, I need to find a way to call this on socket error method, because socket always connected so not calling this method

examples/sensor-channels/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "name": "sensor-channels",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "axios": "^0.19.2",
+    "bootstrap": "^4.4.1",
+    "named-urls": "^2.0.0",
+    "phoenix": "^1.4.12",
+    "react": "^16.12.0",
+    "react-bootstrap": "^1.0.0-beta.16",
```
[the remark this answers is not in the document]
shubham11107: Done

examples/sensor-channels/src/components/SensorDataHandler.jsx
Pavinati: the docs say that onError function callback accepts 1 argument https://hexdocs.pm/phoenix/js/#socket
Try hardcoding a wrong URL and see what's in the error parameter

PR #255, opened 2020-02-21
Return an object when GETting on object aggregated interfaces

### LINKED ISSUE #249: AppEngine returns an array instead of an object when GETting object aggregate interface
When GETting on an object aggregated interface, instead of getting a proper object, the following situation occurs:
+ an empty array is returned when the interface has no data
+ an array is returned when the interface contains just one entry

matt-mazzucato:
Fix #249

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- [appengine_api] Always return an object when GETting on object aggregated interfaces.
+
 ### Removed
 - [appengine_api] Remove deprecated not versioned socket route.
 
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
@@ -842,7 +842,7 @@ defmodule Astarte.AppEngine.API.Device do
 
     cond do
       path == "/" and interface_values == {:error, :path_not_found} ->
-        {:ok, %InterfaceValues{data: []}}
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- [appengine_api] Return an object when GETting on object aggregated interfaces.
```
rbino: I'd say "_Always_ return..." etc, since it already did so sometimes

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -991,22 +991,46 @@ defmodule Astarte.AppEngine.API.Device do
...
-          device_id,
-          :object,
-          :datastream,
-          interface_row,
-          endpoint_id,
-          endpoint_row,
-          only_path,
-          opts
-        )
+        {:ok, %Astarte.AppEngine.API.Device.InterfaceValues{data: values, metadata: metadata}} =
```
rbino: Use a with, otherwise you'll crash if `retrieve_endpoint_values` returns an error

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -991,22 +991,46 @@ defmodule Astarte.AppEngine.API.Device do
...
+                values_map =
+                  %{simplified_path => values_list}
+                  |> MapTree.inflate_tree()
+
+                {:ok, %InterfaceValues{data: values_map, metadata: metadata}}
+            end
+
+          _ ->
+            {:ok, %InterfaceValues{data: values}}
+        end
```
rbino: When you add the `with` above, try to handle this inside the with, possibly adding small helper functions. Nested `case` are a bit ugly.

PR #272, opened 2020-03-10
Add cors plug to Pairing API

Pavinati:
Add plug for CORS request handling

Signed-off-by: Mattia <mattia.pavinati@gmail.com>


apps/astarte_pairing_api/lib/astarte_pairing_api_web/endpoint.ex
```diff
@@ -21,6 +21,7 @@ defmodule Astarte.Pairing.APIWeb.Endpoint do
 
   alias Astarte.Pairing.APIWeb.Metrics
 
+  plug CORSPlug
```
rbino: Put this just before the `Router` plug like in other services

PR #279, opened 2020-03-18
Force update of :re2 in all apps that have it as a dependency

rbino:
Fix a problem preventing re2_nif.so from being compiled and consequently
blocking all applications from starting up


drf:
Fixes the same issue with the new CI setup. LGTM

PR #336, opened 2020-04-13
Fix documentation for 0.11 (and beyond)

drf:
This PR integrates all needed changes/fixes for 0.11 documentation before it gets released. Individual PRs are referenced in individual commits.

### DIFF
--- a/doc/pages/architecture/030-interface.md
+++ b/doc/pages/architecture/030-interface.md
@@ -14,6 +14,7 @@ If you are already familiar with interface's basic concepts, you might want to j
 [Interface Schema](040-interface_schema.html).
 
 ## Versioning
+
 Interfaces are versioned, each interface having both a major version and a minor
 version number. The concept behind these two version numbers mimics [Semantic
 Versioning](http://semver.org/): arbitrary changes can happen exclusively between different major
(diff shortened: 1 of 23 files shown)


doc/pages/architecture/030-interface.md
```diff
@@ -104,11 +133,14 @@ This is how a parametrized mapping looks like:
...
 ```
+
 In this example, `/0/value`, `/1/value` or `/test/value` all map to a valid endpoint, while
 `/te/st/value` can't be resolved by any endpoint.
 
 ### Supported data types
+
 The following types are supported:
+
 * `double`: A double-precision floating-point format as specified by binary64, by the IEEE 754
```
bettio, line 144: * `double`: A double-precision floating-point number as specified by binary64, by the IEEE 754 (NaNs and other non numerical values are not supported).

doc/pages/user/030-manage_interfaces.md
```diff
@@ -72,52 +131,108 @@ _Sample Response_
...
+
+Do you want to continue? [y/n] y
+Interface com.my.Interface1 installed successfully
+Interface com.my.Interface2 installed successfully
+Interface com.my.Interface3 updated successfully to version 1.4
 ```
+
+After invocation, your Astarte Realm will be up to date with all Interfaces in your local directory.
+
+*Note: `astartectl realm-management interfaces sync` currently synchronizes Interfaces only from your local machine to the Realm, and not the other way round. In case the Realm has a more recent version of an interface compared to your local files, or it has some interfaces which are not referenced by your local files, no action will be taken.*
```
bettio, line 158: Let's add `<!-- TODO: blah blah, issue number here -->`. Using comments with TODO will allow us to update documentation as soon as we update astartectl.

drf: I don't think that command will ever be updated, at least in the short term, and that feature isn't planned. The note is ok as it is and it requires no TODOs.

Issue #339, opened 2020-04-13
Metrics expose one entry for each device


matt-mazzucato:
When exposing metrics with Prometheus, avoid generating a metrics' entry for each device.

rbino:
Closed by #341

PR #341, opened 2020-04-13
Fix exposed metrics

matt-mazzucato:
Limit the number of different metrics which are generated. In particular, don't split metrics based on request_path`. See Issue #339


apps/astarte_appengine_api/lib/astarte_appengine_api_web/metrics/pipeline_instrumenter.ex
```diff
@@ -19,7 +19,10 @@
 defmodule Astarte.AppEngine.APIWeb.Metrics.PipelineInstrumenter do
   use Prometheus.PlugPipelineInstrumenter
 
-  def label_value(:request_path, conn) do
-    conn.request_path
-  end
+  # TODO enable label value. Currently it was suppressed as to limit the number
+  # of the different metrics which are generated.
+  # see issue #339
+  #  def label_value(:request_path, conn) do
```
rbino, line 25: Nitpicking: align the code (leave only a single space after the `#`)

PR #346, opened 2020-04-20
data_updater_plant: add support for device heartbeat

### LINKED ISSUE #305: Fix bogus connected state
Unexpected behavior:
Sometimes appengine might report `"connected": true` also for devices that are not connected anymore.

This issue might be caused from devices being permanently disconnected while VMQ is down during a restart. 

Proposed solution:
connected "true" should be set using a TTL.
This change is required in order to avoid bogus connection status.
Existing astarte code should change how devices.connected column is handled: a device is connected when connected == true, therefore is not connected when the connected status is nil or false.
DataUpater process should refresh it every time a "ping" message is received from VMQ.

rbino:
See https://github.com/astarte-platform/astarte_vmq_plugin/pull/16

Fix #305

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
@@ -21,6 +21,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
   alias Astarte.Core.Device
   alias Astarte.Core.InterfaceDescriptor
   alias Astarte.Core.Mapping
+  alias Astarte.DataUpdaterPlant.Config
   alias CQEx.Query, as: DatabaseQuery
   alias CQEx.Result, as: DatabaseResult
   require Logger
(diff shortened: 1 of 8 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -475,6 +499,56 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
     DatabaseQuery.call!(db_client, device_update_query)
   end
 
+  defp refresh_device_connected!(db_client, device_id, ttl) do
+    refresh_connected_statement = """
+    UPDATE devices
+    USING TTL #{to_string(ttl)}
```
bettio, line 505: to_string is not needed, #{...} already uses a String.Chars protocol.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -458,9 +459,32 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
+    set_connection_info!(db_client, device_id, timestamp_ms, ip_address)
+
+    ttl = heartbeat_interval_seconds() * 8
+    refresh_device_connected!(db_client, device_id, ttl)
+  end
+
+  def maybe_refresh_device_connected!(db_client, device_id) do
+    with {:ok, remaining_ttl} <- get_connected_remaining_ttl(db_client, device_id) do
+      if remaining_ttl < heartbeat_interval_seconds() * 2 do
+        Logger.info("Refreshing connected status for device #{inspect(device_id)}")
```
bettio, line 471: I would use Logger.debug here

bettio: anyway beware that we don't need anymore to explicitly log device_id and realm, those information are attached to the processes.

bettio: we also add a tag to every kind of message, to make them machine parsable.

PR #347, opened 2020-04-20
Make ports configurable for each service

matt-mazzucato:
+ Allow to configure the ports for each service
+ All the services with an API allow the port configuration in a single point (i.e. releases.exs)
+ Adopt a coherent naming for port env (i.e. <SERVICE_NAME>_PORT)


apps/astarte_appengine_api/lib/astarte_appengine_api/config.ex
```diff
@@ -33,6 +33,12 @@ defmodule Astarte.AppEngine.API.Config do
     type: NonNegativeInteger,
     default: 10000
 
+  @envdoc "The port used from the Phoenix server."
```
rbino, line 36: Is this necessary? AFAIK this is configured from `releases.exs` (as all Phoenix APIs are)

PR #350, opened 2020-04-22
Wait schema agreement

bettio:


apps/astarte_housekeeping/lib/astarte_housekeeping/c_system.ex
```diff
@@ -0,0 +1,91 @@
...
+
+      any_other ->
+        any_other
+    end
+  end
+
+  def schema_versions(conn) do
+    with {:ok, local_version} <- query_local_schema_version(conn),
+         {:ok, peers_versions} <- query_peers_schema_versions(conn) do
+      {:ok, [local_version | peers_versions]}
```
rbino, line 60: Probably `uniq` goes here (otherwise you would never have a single version)

apps/astarte_housekeeping/lib/astarte_housekeeping/c_system.ex
```diff
@@ -0,0 +1,109 @@
...
+         {:ok, final} <- wait_schema_agreement(conn, timeout) do
+      unless expect_change and initial == final do
+        out
+      else
+        {:error, :no_schema_change}
+      end
+    end
+  end
+
+  def wait_schema_agreement(conn, timeout) when is_integer(timeout) and timeout >= 0 do
```
rbino, line 37: Do we also want to allow an `:infinity` timeout here?

bettio: We are not going to use it right now, so I would do that in a future public module.

Issue #351, opened 2020-04-22
Allow to configure a custom AMQP exchange for events.


bettio:
This is already possible, but there is no configuration API.

It should be added support to r.m. and r.m.a. to amqp actions having following options:
- exchange must be prefixed with "astarte_events_$REALM_"
- routing key: string (no { } accepted, it will be kept for future uses)
additional options:
- message ttl (in millis): mandatory, so we avoid filling up queues.
- message priority
- persistent: default to false

This change will allow any user (having access to realm management) to install triggers with an AMQP action. 

exchange names must be strings such as `astarte_events_$REALM_$NAME` for security reasons. Users will not be able to publish to any exchange. 

Additional future security restrictions:
1. config option for enabling (or disabling) AMQP action.
2. config option for allowing triggers with no message ttl / maximum allowed ttl
3. config option for allowing persistent messages

PR #363, opened 2020-04-27
Implement AMQP 0-9-1 trigger action and improve http action validation

bettio:
Publish by default to configured exchange, and allow configuring a custom one.
Fixes https://github.com/astarte-platform/astarte/issues/351


apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
```diff
@@ -458,7 +458,8 @@ defmodule Astarte.RealmManagement.Engine do
          simple_trigger_maps = build_simple_trigger_maps(serialized_tagged_simple_triggers),
          trigger = build_trigger(trigger_name, simple_trigger_maps, action),
          %Trigger{trigger_uuid: trigger_uuid} = trigger,
-         target = build_trigger_target_container("trigger_engine", trigger_uuid),
+         {exchange, routing_key, opts} = target_from_action(action),
+         target = build_trigger_target_container(exchange, routing_key, trigger_uuid, opts),
```
rbino, line 462: Idea: since we already have the `AMQPTriggerTarget` struct, you could build that in `target_from_action` (passing also `trigger_uuid` instead of returning a tuple, then you pass the `AMQPTriggerTarget in `build_trigger_target_container`

apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/amqp_action.ex
```diff
@@ -0,0 +1,94 @@
...
+defmodule Astarte.RealmManagement.API.Triggers.AMQPAction do
+  use Ecto.Schema
+  import Ecto.Changeset
+  alias Astarte.Core.Realm
+  alias Astarte.RealmManagement.API.Triggers.AMQPAction
+
+  @derive {Phoenix.Param, key: :realm_name}
+  @primary_key false
+  embedded_schema do
+    field :realm_name, :string
```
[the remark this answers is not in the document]
bettio: We need to validate exchange name, the amqp exchange must contain the realm name.

apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/action.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Astarte.RealmManagement.API.Triggers.Action do
+  alias Astarte.RealmManagement.API.Triggers.{AMQPAction, HttpAction}
+
+  @behaviour Ecto.Type
```
Adzz, line 4: This should be 
```elixir
  use Ecto.Type
```

then delete the `embed_as` and `equal` functions

apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/action.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  @impl true
+  def type, do: :map
+
+  @impl true
+  def cast(reward = %{"amqp_exchange" => _}) do
+    EctoMorph.cast_to_struct(reward, AMQPAction)
```
Adzz, line 11: I think instead of using ecto morph here, if you want validations you should call

```elixir
Astarte.RealmManagement.API.Triggers.AMQPAction.changeset(reward)
```

apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/http_action.ex
```diff
@@ -0,0 +1,186 @@
...
+
+  @max_headers_size 8192
+  @max_url_length 8192
+
+  @max_mustache_template_size 1024 * 1024
+
+  @doc false
+  def changeset(%HttpAction{} = action, %{"http_post_url" => _post_url} = attrs) do
+    action
+    |> cast(attrs, @all_attrs)
```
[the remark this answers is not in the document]
bettio: We are doing this as a sanity check to make sure that new and old action format don't get mixed up.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
```diff
@@ -420,21 +421,27 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandler do
...
+    AMQPEventsProducer.publish(exchange, routing_key, payload, opts)
+    |> wait_backoff_and_publish(1, exchange, routing_key, payload, opts)
   end
 
   defp dispatch_event(simple_event = %SimpleEvent{}, %AMQPTriggerTarget{
+         exchange: target_exchange,
          routing_key: routing_key,
-         static_headers: static_headers
+         static_headers: static_headers,
+         message_expiration: message_expiration,
```
rbino, line 437: message_expiration_ms

apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/trigger.ex
```diff
@@ -20,21 +20,73 @@ defmodule Astarte.RealmManagement.API.Triggers.Trigger do
...
-  def changeset(%Trigger{} = trigger, attrs) do
+  def changeset(%Trigger{} = trigger, attrs, opts) do
+    attrs = move_action(attrs)
+
     trigger
-    |> cast(attrs, [:name, :action])
-    |> validate_required([:name, :action])
+    |> cast(attrs, [:name])
+    |> validate_required([:name])
+    |> cast_embed(:amqp_action, required: false, with: {AMQPAction, :changeset, [opts]})
```
rbino, line 42: Is the extra `[]` around `opts` needed? Doesn't result in a list of lists since `opts` is already a list?

bettio: yes, we want that: we want to pass the opts list as a list.

apps/astarte_realm_management_api/test/astarte_realm_management_api/triggers/triggers_test.exs
```diff
@@ -30,7 +30,7 @@ defmodule Astarte.RealmManagement.API.TriggersTest do
       {:ok, trigger} =
         attrs
         |> Enum.into(@valid_attrs)
-        |> RealmManagement.API.Triggers.create_trigger()
+        |> RealmManagement.API.Triggers.create_trigger("test")
```
rbino, line 33:
```elixir
def create_trigger(realm_name, attrs \\ %{})
```

I think the arguments are backwards, you should break the pipeline and call `RealmManagement.API.Triggers.create_trigger("test", params)` (in reality the tests below are not even executed, but at least we don't introduce new wrong code)

PR #366, opened 2020-05-01
appengine_api: prevent crashes when updating metadata or aliases

### LINKED ISSUE #357: Do not crash when using "" as alias name or value
"" should be handled as invalid, instead of crashing. Also metadata should be checked for similar issues (empty string key should not be allowed, while empty string value is fine for metadata).

matt-mazzucato:
+ Do not crash when using an empty string for aliases (either key or
value). 400 is returned instead.
+ Do not crash when using an empty string for metadata keyi. 400 is
returned instead.
+ Metadata values can be set to empty string.

Fix #357

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
@@ -71,7 +71,7 @@ defmodule Astarte.AppEngine.API.Device do
          credentials_inhibited_change = Map.get(changeset.changes, :credentials_inhibited),
          :ok <- change_credentials_inhibited(client, device_id, credentials_inhibited_change),
          aliases_change = Map.get(changeset.changes, :aliases, %{}),
-         metadata_change <- Map.get(changeset.changes, :metadata, %{}),
+         metadata_change = Map.get(changeset.changes, :metadata, %{}),
          :ok <- update_aliases(client, device_id, aliases_change),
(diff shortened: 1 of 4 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -84,11 +84,27 @@ defmodule Astarte.AppEngine.API.Device do
...
+
+      {:error, :invalid_metadata} ->
+        {:error, :invalid_metadata}
+
+      {:error, :alias_already_in_use} ->
+        {:error, :alias_already_in_use}
+
+      error ->
+        error
     end
```
rbino, line 99: You're catching 3 special cases and returning them as-is, then you have a generic case that you also return as-is. Might as well drop the `else` completely since it will fall-through anyway

apps/astarte_appengine_api/lib/astarte_appengine_api_web/controllers/device_status_controller.ex
```diff
@@ -54,6 +54,15 @@ defmodule Astarte.AppEngine.APIWeb.DeviceStatusController do
...
         render(conn, "show.json", device_status: device_status)
+      else
+        {:error, :invalid_alias} ->
+          {:error, :invalid_alias}
+
+        {:error, :invalid_metadata} ->
+          {:error, :invalid_metadata}
+
+        error ->
+          error
```
rbino, line 65: Same here

PR #372, opened 2020-05-11
data_updater: check if path and interface are valid UTF-8 strings

### LINKED ISSUE #371: path and interface could contain invalid UTF-8 strings
It seems that invalid utf-8 strings are accepted as topic in VerneMQ, so data containing invalid strings could reach the `handle_data` in `data_updater/impl.ex` and be passed to the database.

We should validate such fields to ensure they are valid strings before manipulating them.

rbino:
Avoid crashes in the database due to passing around invalid strings.

Fix #371

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
@@ -327,7 +327,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
 
     new_state = execute_time_based_actions(state, timestamp, db_client)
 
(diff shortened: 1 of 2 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -532,6 +533,22 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
 
         update_stats(new_state, interface, nil, path, payload)
 
+      {:error, :invalid_interface} ->
+        Logger.warn("Received invalid interface: #{interface}.")
```
bettio, line 537: if invalid it should be inspected.

PR #423, opened 2020-06-25
Rename http_(custom)_headers to http_static_headers

bettio:
http_static_headers is consistent with AMQP amqp_static_headers and also
it underlines that they are just a subset of all http headers.
"http_custom_headers" has not been chosen since they might be also standard/well-know
headers, such as Authorization.
Furthermore an additional test has been added for http action Jason
encoder.


apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -180,7 +180,7 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
 
   defp execute_action(payload, headers, action) do
     with {:ok, method, url} <- fetch_method_and_url(action),
-         custom_headers_map = Map.get(action, "http_custom_headers", %{}),
+         custom_headers_map = Map.get(action, "http_static_headers", %{}),
```
rbino, line 183: `static_headers_map`

PR #424, opened 2020-06-26
Rename ...genericsensors.Values/_/value to ...genericsensors.Values/_…

bettio:
…/values

GET on AppEngine on /value will return a collection of items, therefore
it should be renamed to /values.
On the SDK side it would look as:
```
sendData("org.astarte-platform.genericsensors.Values", "/values", value);
```
Using astartectl on it would be something like:
```
astartectl rlm appengine devices get-samples device_id ...Values /id/values
```


rbino:
I can see this specific case making sense, but I don't know if it can be applied at large (_e.g._ a `com.something.AirSensor` interface should have `/temperatures`, `/humidities` and so on, which imho don't sound quite right).

PR #435, opened 2020-07-03
housekeeping: add realm deletion

rbino:
The feature can be enabled with an env variable and it's disabled by default


apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -35,6 +35,15 @@ defmodule Astarte.Housekeeping.Config do
     type: :integer,
     default: 4008
 
+  @envdoc """
+  By default Astarte Housekeeping doesn't support realm deletion. Set this variable to true to
+  enable this feature. WARNING: this feature can cause permanent data loss when deletin a realm.
```
bettio, line 40: deleting

PR #440, opened 2020-07-17
Update Phoenix and rework astarte metrics

matt-mazzucato:
+ Update Phoenix to version 1.5.
+ Rework astarte metrics to reduce the clutter while retrieving relevant information.


apps/astarte_appengine_api/lib/astarte_appengine_api/application.ex
```diff
@@ -42,22 +38,18 @@ defmodule Astarte.AppEngine.API.Application do
...
     children = [
-      supervisor(Astarte.RPC.AMQP.Client, []),
-      supervisor(Astarte.AppEngine.API.Rooms.MasterSupervisor, []),
-      supervisor(Astarte.AppEngine.API.Rooms.AMQPClient, []),
-      supervisor(Astarte.AppEngine.APIWeb.Endpoint, []),
+      Astarte.RPC.AMQP.Client,
+      Astarte.AppEngine.API.Rooms.MasterSupervisor,
+      Astarte.AppEngine.API.Rooms.AMQPClient,
+      Astarte.AppEngine.APIWeb.Endpoint,
+      {Phoenix.PubSub, name: Astarte.AppEngine.API.PubSub},
```
rbino, line 51: This should probably start before the `Endpoint` since the `Endpoint` calls into it

apps/astarte_appengine_api/lib/astarte_appengine_api/application.ex
```diff
@@ -42,22 +38,18 @@ defmodule Astarte.AppEngine.API.Application do
...
+      Astarte.AppEngine.APIWeb.Telemetry,
```
rbino, line 52: And this should probably be the first thing that starts since potentially everything else can emit metrics

apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/room.ex
```diff
@@ -200,8 +200,8 @@ defmodule Astarte.AppEngine.API.Rooms.Room do
       room_cleanup(state)
 
       :telemetry.execute(
-        [:astarte, :appengine, :channels, :room_closed],
-        %{},
+        [:astarte, :appengine, :channels, :room_opened],
```
rbino, line 203: I see the point of doing this (summing the `-1` with the same event name) but I'm not sure if emitting a `:room_opened` event when the room is closed is the correct behaviour.
Did you try to look at other libraries that do similar stuff and see how they do it?

apps/astarte_appengine_api/lib/astarte_appengine_api_web/controllers/metrics_controller.ex
```diff
@@ -14,10 +14,17 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
 
-defmodule Astarte.HousekeepingWeb.Metrics.PrometheusExporter do
-  @moduledoc false
+defmodule Astarte.AppEngine.APIWeb.MetricsController do
+  use Astarte.AppEngine.APIWeb, :controller
+
+  plug Plug.Telemetry, event_prefix: [:prometheus_metrics, :plug]
```
rbino, line 21: Do we need to instrument the metrics controller? It seems kind of redundant

apps/astarte_appengine_api/lib/astarte_appengine_api_web/telemetry/health_status.ex
```diff
@@ -14,29 +14,22 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
```
rbino, line 17: Extra comment line removed, restore it

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant_web/telemetry.ex
```diff
@@ -0,0 +1,107 @@
...
+        tags: [:realm]
+      ),
+      sum("astarte.data_updater_plant.data_updater.exchanged_bytes.bytes",
+        tags: [:realm]
+      ),
+      counter("astarte.data_updater_plant.triggers_handler.published_event.count",
+        tags: [:realm]
+      ),
+      counter("astarte.data_updater_plant.data_updater.detected_crash.count"),
+      last_value("astarte.data_updater_plant.service.health",
```
bettio, line 93: I don't understand this one

apps/astarte_housekeeping/lib/astarte_housekeeping_web/plug/metrics_plug.ex
```diff
@@ -16,8 +16,19 @@
...
+defmodule Astarte.HousekeepingWeb.MetricsPlug do
+  import Plug.Conn
+
+  def init(_args), do: nil
+
+  def call(%{request_path: "/metrics", method: "GET"} = conn, _opts) do
+    metrics = TelemetryMetricsPrometheus.Core.scrape()
+
+    conn
+    |> put_resp_content_type("text/plain")
```
bettio, line 28: Maybe `text/html; charset=utf-8` would be better, we should check it.

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/health/health.ex
```diff
@@ -29,9 +29,21 @@ defmodule Astarte.Housekeeping.API.Health do
   """
   def get_backend_health do
```
bettio, line 30: We are using try ... rescue in other places, should we use it everywhere?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant_web/telemetry.ex
matt-mazzucato: The idea is to monitor `health` and make it available to prometheus (i.e. the status can be plotted such that we can understand the service status across time)

apps/astarte_appengine_api/lib/astarte_appengine_api_web/plug/health_plug.ex
```diff
@@ -1,7 +1,7 @@
 #
 # This file is part of Astarte.
 #
-# Copyright 2019 Ispirata Srl
+# Copyright 2020 Ispirata Srl
```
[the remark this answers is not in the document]
matt-mazzucato: `health_controller.ex` was removed. `health_plug.ex` is a new file.

apps/astarte_housekeeping/lib/astarte_housekeeping_web/plug/metrics_plug.ex
matt-mazzucato: `utf-8` is already the default charset, while the content type `text/plain` is consistent with the previous implementation.

apps/astarte_appengine_api/lib/astarte_appengine_api_web/endpoint.ex
```diff
@@ -43,6 +38,9 @@ defmodule Astarte.AppEngine.APIWeb.Endpoint do
 
   plug Plug.RequestId
   plug PlugLoggerWithMeta
```
rbino, line 40: I just realized that if we move `plug PlugLoggerWithMeta` _below_ `HealthPlug` and `MetricPlug`, we would also solve the spammy logs caused by metrics and health requests (the conn gets halted in `HealthPlug` and `MetricsPlug`, so it won't pass in `PlugLoggerWithMeta`).

@bettio opinion? I'm personally in favor, spammy logs makes it very hard to debug problems that happened e.g. during the weekend.

bettio: I agree with this change

PR #452, opened 2020-07-30
Handle device disconnection

### LINKED ISSUE #398: Mark device as offline and send device_disconnected event when forcing a device disconnection
When Data Updater Plant forces the disconnection of a device
 https://github.com/astarte-platform/astarte/blob/1bfaad8f9108b3c107599e5176b015a2a8c72f3a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex#L1671
it currently doesn't mark it as disconnected in the database and it doesn't send a `DeviceDisconnected` event.

This should be added since the broker will not send an explicit `device_disconnected` message if the disconnection is forced from the API.

matt-mazzucato:
Mark device as offline and send device_disconnected event when forcing a device disconnection
Fix #398

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
@@ -161,40 +161,15 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
   def handle_disconnection(state, message_id, timestamp) do
     {:ok, db_client} = Database.connect(realm: state.realm)
 
-    new_state = execute_time_based_actions(state, timestamp, db_client)
-
-    timestamp_ms = div(timestamp, 10_000)
-
-    Queries.set_device_disconnected!(
-      db_client,
-      new_state.device_id,
-      timestamp_ms,
-      new_state.total_received_msgs,
-      new_state.total_received_bytes,
-      new_state.interface_exchanged_msgs,
-      new_state.interface_exchanged_bytes
-    )
-
-    trigger_targets = Map.get(new_state.device_triggers, :on_device_disconnection, [])
-    device_id_string = Device.encode_device_id(new_state.device_id)
-
-    TriggersHandler.device_disconnected(
-      trigger_targets,
-      new_state.realm,
-      device_id_string,
-      timestamp_ms
-    )
+    new_state =
+      state
+      |> execute_time_based_actions(timestamp, db_client)
+      |> set_device_disconnected(db_client, timestamp)
 
     MessageTracker.ack_delivery(new_state.message_tracker, message_id)
     Logger.info("Device disconnected.", tag: "device_disconnected")
 
-    :telemetry.execute(
-      [:astarte, :data_updater_plant, :data_updater, :device_disconnection],
-      %{},
-      %{realm: new_state.realm}
-    )
-
-    %{new_state | connected: false, last_seen_message: timestamp}
+    %{new_state | last_seen_message: timestamp}
   end
 
   defp execute_incoming_data_triggers(
@@ -572,7 +547,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
           tag: "write_on_server_owned_interface"
         )
 
-        ask_clean_session(new_state)
+        {:ok, new_state} = ask_clean_session(new_state, timestamp)
         MessageTracker.discard(new_state.message_tracker, message_id)
 
         :telemetry.execute(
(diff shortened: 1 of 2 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -161,42 +161,14 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
...
-
-    TriggersHandler.device_disconnected(
-      trigger_targets,
-      new_state.realm,
-      device_id_string,
-      timestamp_ms
-    )
+    new_state =
+      execute_time_based_actions(state, timestamp, db_client)
+      |> set_device_disconnected(db_client, timestamp)
```
rbino, line 166: Style nitpicking:
```elixir
state
|> execute_time_based_actions(timestamp, db_client)
|> set_device_disconnected(db_client, timestamp)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -576,15 +546,13 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
           tag: "write_on_server_owned_interface"
         )
 
-        ask_clean_session(new_state)
-        MessageTracker.discard(new_state.message_tracker, message_id)
+        {:ok, updated_state} = ask_clean_session(new_state, message_id, timestamp)
```
rbino, line 549: Just rebind `new_state`, imho having yet another name for a new version of the state does more harm than good (moreover, it introduces a lot of noise in the MR).

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -2007,15 +2011,18 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
 
     with :ok <- Queries.set_pending_empty_cache(db_client, device_id, true),
          :ok <- VMQPlugin.disconnect("#{realm}/#{encoded_device_id}", true) do
+      new_state = set_device_disconnected(state, db_client, timestamp)
+
+      MessageTracker.ack_delivery(new_state.message_tracker, message_id)
```
rbino, line 2016: I think you're "double acking" the message: you ack it here and then you discard it at the outer level (every `ask_clean_session` call is followed by a `MessageTracker.discard` call).

I think this `ack_delivery` is not needed here (and, by extension, you don't actually need to pass `message_id` to `ask_clean_session`), or am I missing something?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -2007,15 +2011,18 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
...
+      Logger.info("Device disconnected.", tag: "device_disconnected")
+
       :telemetry.execute(
         [:astarte, :data_updater_plant, :data_updater, :clean_session_request],
         %{},
-        %{
-          realm: state.realm
-        }
+        %{realm: state.realm}
```
rbino, line 2022: It's the same thing, but once you have `new_state` use that instead of `state`

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -2007,15 +2011,18 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
 
     with :ok <- Queries.set_pending_empty_cache(db_client, device_id, true),
          :ok <- VMQPlugin.disconnect("#{realm}/#{encoded_device_id}", true) do
+      new_state = set_device_disconnected(state, db_client, timestamp)
+
+      MessageTracker.ack_delivery(new_state.message_tracker, message_id)
+      Logger.info("Device disconnected.", tag: "device_disconnected")
```
rbino, line 2017: I would use a different message to distinguish between a server-initiated disconnection and a client-initiated one, something like
```elixir
Logger.info("Succesfully forced device disconnection.", tag: "forced_device_disconnection")
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -576,15 +547,13 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
           tag: "write_on_server_owned_interface"
         )
 
-        ask_clean_session(new_state)
+        {:ok, new_state} = ask_clean_session(new_state, timestamp)
         MessageTracker.discard(new_state.message_tracker, message_id)
```
rbino, line 551: Now that `ask_clean_session` can possibly fail, I'm not sure about the order of these two lines.

The current order gives us Consistency over Availability, while swapping them gives us Availability over Consistency. @bettio what's your opinion?

bettio: Let's start with Consistency over Availability.

PR #486, opened 2020-11-03
Include the sender's name within alert mails

matt-mazzucato:
Signed-off-by: Mattia Mazzucato <matt.mazzucato@gmail.com>

### DIFF
--- a/tools/astarte_e2e/lib/astarte_e2e/config/normalized_mail_address.ex
+++ b/tools/astarte_e2e/lib/astarte_e2e/config/normalized_mail_address.ex
@@ -1,0 +1,37 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2020 Ispirata Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


tools/astarte_e2e/lib/astarte_e2e/config/normalized_mail_address.ex
```diff
@@ -0,0 +1,40 @@
...
+
+    if String.match?(value, regexp) do
+      [name, mail] =
+        value
+        |> String.trim(">")
+        |> String.split("<")
+
+      {:ok, {name, mail}}
+    else
+      {:ok, value}
```
rbino, line 33: Since you're already matching with a regex, use this
https://hexdocs.pm/elixir/Regex.html#named_captures/3
so you can match and extract in a single pass
(I also fixed-up the regex a little bit)
```elixir
    regexp = ~r/^(?<name>.+) <(?<email>.+@.+)>$/

    case Regex.named_captures(regexp, value) do
      %{"email" => email, "name" => name} ->
        {:ok, {name, email}}

      _ ->
        {:ok, value}
    end
```

rbino (changes requested):

bettio (approved):

rbino (approved):

Issue #513, opened 2021-01-12
value_changed and value_change_applied triggers problem


rbino:
`value_changed` and `value_change_applied` triggers are not executed when `match_path` is `/*`

rbino:
Closed by #520

PR #520, opened 2021-01-21
RM: Disallow `/*` match path when using value_change* trigger

bettio:
See also issue #513.

### DIFF
--- a/apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
+++ b/apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
@@ -553,10 +553,18 @@ defmodule Astarte.RealmManagement.Engine do
     with {:ok, interface} <- Queries.fetch_interface(client, interface_name, interface_major) do
       case interface.aggregation do
         :individual ->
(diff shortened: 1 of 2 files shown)


apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
```diff
@@ -553,10 +553,18 @@ defmodule Astarte.RealmManagement.Engine do
...
-            :ok
+          cond do
+            interface.type != :properties and properties_trigger_type?(data_trigger_type) ->
+              {:error, :invalid_datastream_trigger}
+
+            match_path == "/*" and
+                (data_trigger_type == :VALUE_CHANGE or data_trigger_type == :VALUE_CHANGE_APPLIED) ->
+              # TODO: this is a workaround to a data updater plant limitation
+              # see also https://github.com/astarte-platform/astarte/issues/513
+              {:error, :invalid_trigger}
```
rbino, line 564: I'd return `{:error, :unsupported_trigger_type}` so it's clearer to the user what the specific problem is

rbino (changes requested):

rbino (approved):

PR #533, opened 2021-02-16
Add javascript examples documentation

davidebriani:
This PR updates JavaScript examples:
- dependencies are updated
- implementation code is fixed and updated
- a `README.md` file is added to each example as a mean to document supposed usage

Impacted examples include:

`javascript-sensors-viewer`

javascript-sensors-viewer-stream-test

`sensor-graph`

sensor-graph

`sensors-viewer`

sensors-viewer-stream-test


examples/README.md
```diff
@@ -0,0 +1,13 @@
+# Examples
+
+This is a collection of examples that show how you could approach some use cases in fetching and
```
Pavinati, line 3: What about `[...] that show how to interact with Astarte in common user cases.`

examples/javascript-sensors-viewer/README.md
```diff
@@ -0,0 +1,100 @@
...
+4. modify the `configuration` variable in `index.html` and specify: the `endpoint` of the Astarte
+   instance, the `realm` you chose, a valid JWT `token` that can be used to communicate with
+   AppEngine, the `device` ID your registered.
+
+## Docker walkthrough
+
+You can use Astarte's `stream-qt5-test` to emulate an Astarte device and generate a `datastream`:
+
+```sh
+docker run --net="host" -e "DEVICE_ID=<device_id>" -e "PAIRING_HOST=<pairing_host>" -e "REALM=<realm>" -e "AGENT_KEY=<agent_key>" -e "IGNORE_SSL_ERRORS=true" astarte/astarte-stream-qt5-test:0.11-snapshot
```
Pavinati, line 27: we should not suggest users to use snapshot images. Let's use a stable tag instead, like `0.11.4`

examples/javascript-sensors-viewer/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+where `device_id` is the device you registered, `pairing_host` is the Astarte endpoint for the
+Pairing APIs, `realm` is the name of your realm, `agent_key` is the valid JWT token of an Agent.
+
+If all went well, `stream-qt5-test` should now be streaming data to Astarte.
+
+Going back to the web page of this example, you can input the device ID to request an overview of
+the device sensors and their data.
+
+## Elixir walkthrough
```
Pavinati, line 38: While I'm ok with stream-qt5-test as it's a one-liner, i don't think this is the right place for explaining how to use the elixir sdk. I would just mention "... or use an Astarte SDK for the language you're most comfortable with"

davidebriani: That's a good point. I was not sure about this.
Showing a use case with one SDK could be beneficial since you have more flexibility and can use more interfaces to send both Values and advertise AvailableSensors, while the docker command can only send Values.
What's your take?

examples/javascript-sensors-viewer/README.md
davidebriani: Right, fixed.

examples/README.md
davidebriani: That's better, I agree. Fixed it.

examples/javascript-sensors-viewer/README.md
davidebriani: Since we can document the use case with `stream-qt5-test`, I removed the walkthrough with the Elixir SDK

Pavinati (approved):

examples/README.md
```diff
@@ -0,0 +1,12 @@
+# Examples
+
+This is a collection of examples that shows how to interact with Astarte in common use cases.
+
+## Javascript examples
+
+- [javascript-sensors-viewer](javascript-sensors-viewer): sample web page in vanilla Javascript to
```
bettio, line 7: Javascript or JavaScript? Which is the correct one?

bettio (changes requested):

examples/README.md
davidebriani: Thanks for the heads up, we will stick to JavaScript. I fixed all occurrences of it in the PR.

bettio (approved):

PR #542, opened 2021-02-24
Astarte_E2E: fix "unknown" failure

matt-mazzucato:
+ Update phoenix_gen_socket_client
+ Prevent failures from raising alerts with `unknown` id
+ Treat client disconnections as special cases which cause an immediate alert


tools/astarte_e2e/lib/astarte_e2e/service_notifier.ex
```diff
@@ -164,39 +181,41 @@ defmodule AstarteE2E.ServiceNotifier do
...
-        [{:reply, from, :nothing_to_do}]
-      else
-        [{:reply, from, :mail_sent}]
-      end
+    event_id = Hukai.generate("%a-%A")
 
-    updated_data = Map.put(data, :failures_before_alert, data.failures_before_alert - 1)
+    updated_data =
+      data
+      |> Map.put(:failures_before_alert, data.failures_before_alert - 1)
```
rbino, line 196: Is this necessary? It starts from 0 so now it's going to `-1`

rbino (changes requested):
A rebase is needed

bettio (approved):

rbino (approved):

PR #555, opened 2021-03-22
Rename device metadata to attributes

rbino:
Also add documentation to perform the manual steps needed on existing clusters
using a v1.0 development version. This was not handled using migrations since
the change happened between development versions.


apps/astarte_appengine_api/priv/static/astarte_appengine_api.yaml
```diff
@@ -292,7 +292,7 @@ paths:
       tags:
         - device
       summary: Update a device writeable property
-      description: Update any of the writeable device properties such as device aliases or device metadata.
+      description: Update any of the writeable device properties such as device aliases, device attributes or credentials inhibited.
```
Pavinati, line 295: is `credentials inhibited` no longer writeable?

rbino: The opposite, this documentation was not updated when `credentials_inhibited` was made writable so I fixed that

bettio (approved):

PR #551, opened 2021-03-12
appengine: encode binaryblob values with Base64 even in aggregates

rbino:
Fix #547

Signed-off-by: Riccardo Binetti <riccardo.binetti@ispirata.com>

Issue #550, opened 2021-03-11
Aggregate messages which contain a subset of the needed keys are accepted


rbino:
Data Updater Plant seems to not care if an object aggregation contains less keys than it's supposed to, and saves the available keys with their values and the missing keys with a null value. This means that right now all endpoints in a device owned object aggregation interface are considered "optional".

We should decide if we want to embrace this behaviour by documenting it or to change it by enforcing that an object aggregation should populate all of its endpoints. The middle ground could be to make this configurable with a new interface property (which would make sense only in object aggregate interfaces).

rbino:
Right now, Data Updater Plant checks that all keys that are present in the object have the correct type, but it doesn't check that all endpoints actually have a corresponding key in the object.

This means that if an interface has endpoints `/foo/a`, `/foo/b` and `/foo/c`, DUP accepts an object with payload `%{a: 1, b: 2}` even if `c` is missing (`c` remains null in the database).

This makes data from incomplete data points unfetchable right now, because [this MR](https://github.com/astarte-platform/astarte/pull/551), while correctly making it so that data from aggregate interfaces passes through to_json_friendly to support, e.g., binary to Base64 translation, introduced the side effect that now AppEngine is refusing to return null as value, which shouldn't be there in the first place.

PR #573, opened 2021-04-22
appengine: allow returning nil in AstarteValue.to_json_friendly

rbino:
The interaction between https://github.com/astarte-platform/astarte/issues/550
and https://github.com/astarte-platform/astarte/pull/551 was preventing
aggregate interfaces that got populated with incomplete objects to be queried.
This fixes that, allowing AppEngine to return null in some JSON fields (since
it's not its job to fix that kind of issue anyway).


bettio (changes requested):
Changelog?

PR #574, opened 2021-04-23
data_updater_plant: refute incomplete objects

### LINKED ISSUE #550: Aggregate messages which contain a subset of the needed keys are accepted

rbino:
Data Updater Plant was accepting objects containing a subset of the required
keys. This MR fixes that behaviour, leaving a way to accept the old (incorrect)
one for a transition period: passing
DATA_UPDATER_PLANT_ACCEPT_INCOMPLETE_OBJECTS=true as environment variable.

Fix #550


bettio (changes requested):
At the moment this would break any existing installation since devices having an older interface minor version have less keys than expected, however their data are still valid.
We need to compare keys against expected objects for given major and minor.
We already keeping track of the minor version which introduced any endpoint, however this sounds like a complex change.

bettio:
I suggest documenting this "issue" as a known issue and we must tell our users to pay attention to this that will have stricter checks in future versions.
I think that at the moment the easiest approach would be a device SDK side check.

Issue #569, opened 2021-04-16
Upgrade amqp library in all components


rbino:
A new major version was released which could possibly fix some reconnection failure corner cases, use that everywhere

PR #577, opened 2021-04-30
Upgrade AMQP library to version 2.1 

Annopaolo:
See [#569](https://github.com/astarte-platform/astarte/issues/569).


bettio (changes requested):
Same operation should be repeated on more repos.

Annopaolo:
> Same operation should be repeated on more repos.

Done!

bettio (approved):

PR #584, opened 2021-05-13
RM: Queries: fix all interfaces with same name deletion bug

bettio:
The query was limiting the delete operation on all interfaces with given
name. The DELETE query must be limited using (name, major) couple,
otherwise other interfaces are deleted by mistake.

This bug was not causing data loss, however it causes service
distruption and it requires a manual fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.11.5] - Unreleased
+- [realm_management] Avoid deleting all interfaces sharing the same by mistake, only the v0
+  interface can be deleted.
```
rbino, line 9: *sharing the same _name_

PR #595, opened 2021-06-10
Revert "DataUpdater: triggers are not yet supported on object aggregated

rbino:
This reverts commit 974c065bc937c683c32691ea90ba620984fd39f2.
Volatile triggers are now supported even in object aggregation since all
types are correctly JSON encoded even if they're contained in an object
aggregation (see
https://github.com/astarte-platform/astarte_core/commit/b22148daf84cf2355e9e7ba02e9ce39dae5db336)

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
@@ -1426,25 +1426,15 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
 
     if Map.has_key?(new_state.interface_ids_to_name, object_id) do
       interface_name = Map.get(new_state.interface_ids_to_name, object_id)
-
-      %InterfaceDescriptor{automaton: automaton, aggregation: aggregation} =
-        new_state.interfaces[interface_name]
+      %InterfaceDescriptor{automaton: automaton} = new_state.interfaces[interface_name]
 
       case trigger do
         {:data_trigger, %ProtobufDataTrigger{match_path: "/*"}} ->
-          if aggregation == :individual do
-            {:ok, load_trigger(new_state, trigger, target)}
-          else
-            {{:error, :unsupported_interface_aggregation}, state}
-          end
+          {:ok, load_trigger(new_state, trigger, target)}
 
         {:data_trigger, %ProtobufDataTrigger{match_path: match_path}} ->
           with {:ok, _endpoint_id} <- EndpointsAutomaton.resolve_path(match_path, automaton) do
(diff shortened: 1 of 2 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -1466,15 +1456,10 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
            interface_major: major,
            match_path: "/*"
          }} ->
-          with {:ok, db_client} <-
-                 Database.connect(realm: state.realm),
-               {:ok, %InterfaceDescriptor{aggregation: :individual}} <-
-                 InterfaceQueries.fetch_interface_descriptor(db_client, interface_name, major) do
+          with {:ok, db_client} <- Database.connect(realm: state.realm),
+               :ok <- InterfaceQueries.check_if_interface_exists(db_client, interface_name, major) do
```
bettio, line 1460: Let's check what `check_if_interface_exists` returns in case of missing interface, and make sure that it is properly managed by our "with else" block.

bettio (changes requested):
Let's add a CHANGRLOG entry and double check check_if_interface_exists return value.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
rbino: `check_if_interface_exists` returns one of these errors:
```
      [count: 0] ->
        {:error, :interface_not_found}

      %{acc: _, msg: error_message} ->
        Logger.warn("check_if_interface_exists: database error: #{error_message}")
        {:error, :database_error}

      {:error, reason} ->
        Logger.warn("check_if_interface_exists: failed, reason: #{inspect(reason)}.")
        {:error, :database_error}
```

The call stack is:
```
(incoming AMQP RPC)
-> RPC.Handler.call_rpc/1
-> DataUpdater.handle_install_volatile_trigger/8
-> Server.handle_call/4
-> Impl.handle_install_volatile_trigger/7
```

The result is passed out as-is up the stack until it reaches `call_rpc` where it's matched like this:
```elixir
    with :ok <-
           DataUpdater.handle_install_volatile_trigger(
             realm_name,
             device_id,
             object_id,
             object_type,
             parent_id,
             simple_trigger_id,
             simple_trigger,
             trigger_target
           ) do
      %GenericOkReply{}
      |> encode_reply()
      |> ok_wrap()
    else
      {:error, reason} ->
        generic_error(reason)
    end
```

bettio (approved):

PR #597, opened 2021-06-28
Data Updater Plant: use a reasonable backoff time when publishing to RabbitMQ.

Annopaolo:
Maximum (exponential) backoff time is now around 5 minutes.

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
@@ -25,7 +25,7 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandler do
   by the Trigger targets
   """
 
-  @max_backoff_exponent 10
+  @max_backoff_exponent 8
 
   use Astarte.Core.Triggers.SimpleEvents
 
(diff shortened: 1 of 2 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
```diff
@@ -469,4 +470,9 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandler do
 
     result
   end
+
+  defp compute_backoff_time(current_attempt) do
+    minimum_duration = (1 <<< current_attempt) * 1000
+    minimum_duration + round(minimum_duration * 0.25 * :rand.uniform())
+  end
```
rbino, line 477: You're not using `@max_backoff_exponent`

rbino (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
rbino: nvm, it's used in the caller

rbino (approved):

bettio (approved):

PR #612, opened 2021-09-14
DUP: add handle_data duration to metrics

matt-mazzucato:
Measure how long it takes to handle data.

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/server.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/server.ex
@@ -68,7 +68,16 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Server do
     timeout = Config.data_updater_deactivation_interval_ms!()
 
     if MessageTracker.can_process_message(state.message_tracker, message_id) do
+      start = System.monotonic_time()
+
       new_state = Impl.handle_data(state, interface, path, payload, message_id, timestamp)
+
+      :telemetry.execute(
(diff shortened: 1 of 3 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/server.ex
```diff
@@ -68,7 +68,16 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Server do
     timeout = Config.data_updater_deactivation_interval_ms!()
 
     if MessageTracker.can_process_message(state.message_tracker, message_id) do
+      start = System.monotonic_time()
+
       new_state = Impl.handle_data(state, interface, path, payload, message_id, timestamp)
+
+      :telemetry.execute(
+        [:astarte, :data_updater_plant, :data_updater, :handle_data],
+        %{duration: System.monotonic_time() - start},
```
bettio, line 77: Is System.monotonic_time() in milliseconds? I'm not really sure about that.

bettio (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/server.ex
matt-mazzucato: The metric is defined as
```
      last_value("astarte.data_updater_plant.data_updater.handle_data.duration",
        tags: [:realm],
        unit: {:native, :millisecond}
      ),
```
and the unit is explicitly set. More on this [here](https://hexdocs.pm/telemetry/readme.html).

bettio (approved):

PR #632, opened 2021-11-30
DUP: use multiple connections for RabbitMQ

Annopaolo:
Use [ExRabbitPool](https://github.com/esl/ex_rabbit_pool) to enable multiple (pooled) connections to RabbitMQ, instead of just one.
The total number of connections is obtained by dividing the total number of queues by the number of queues per single connection.


rbino (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_pipeline_supervisor.ex
```diff
@@ -34,6 +35,10 @@ defmodule Astarte.DataUpdaterPlant.DataPipelineSupervisor do
     children = [
       {Registry, [keys: :unique, name: Registry.MessageTracker]},
       {Registry, [keys: :unique, name: Registry.DataUpdater]},
+      # we use different connections for producers and consumers
+      {ExRabbitPool.PoolSupervisor,
```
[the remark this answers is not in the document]
Annopaolo: Actually, I fear having two `PoolSupervisor`s is a must as long as we can have different `rabbitmq_config`s for producers and consumers. Indeed, `PoolSupervisor.start_link/2` accepts [only one rabbit config](https://github.com/esl/ex_rabbit_pool/blob/89288bc967e7ac5d7d4662be91821942343af379/lib/pool_supervisor.ex#L12), and so the case in which the producer pool has a different set of options than the consumer pool can't be implemented

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_events_producer.ex
```diff
@@ -45,70 +42,55 @@ defmodule Astarte.DataUpdaterPlant.AMQPEventsProducer do
...
-    reply = Exchange.declare(chan, exchange, :direct, durable: true)
-
-    {:reply, reply, chan}
+    ExRabbitPool.with_channel(:producers_pool, fn {:ok, channel} ->
+      reply = Exchange.declare(channel, exchange, :direct, durable: true)
+      {:reply, reply, state}
+    end)
   end
 
   def handle_info(:try_to_connect, _state) do
```
rbino, line 68: This is actually unused now so it can be removed

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_events_producer.ex
```diff
@@ -45,70 +42,55 @@ defmodule Astarte.DataUpdaterPlant.AMQPEventsProducer do
...
     {:ok, new_state} = rabbitmq_connect()
     {:noreply, new_state}
   end
 
-  def handle_info({:DOWN, _, :process, _pid, reason}, _state) do
-    Logger.warn("RabbitMQ connection lost: #{inspect(reason)}. Trying to reconnect...",
-      tag: "events_producer_conn_lost"
-    )
+  defp rabbitmq_connect() do
```
rbino, line 73: Not sure I'll leave this with this name given it's actually already connected. Since now it's just called from init, and this is actually initialization, I'd probably just move this function body in `init`

rbino (changes requested):

apps/astarte_data_updater_plant/mix.exs
```diff
@@ -84,7 +84,8 @@ defmodule Astarte.DataUpdaterPlant.Mixfile do
       {:skogsra, "~> 2.2"},
       {:telemetry, "~> 0.4"},
       {:observer_cli, "~> 1.5"},
-      {:dialyzex, github: "Comcast/dialyzex", only: [:dev, :ci]}
+      {:dialyzex, github: "Comcast/dialyzex", only: [:dev, :ci]},
+      {:ex_rabbit_pool, github: "leductam/ex_rabbit_pool"}
```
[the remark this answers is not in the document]
Annopaolo: Actually it's because the esl repo does not even support `amqp` version `2.1`, while leductam's fork only change is supporting also `amqp ~> 2.0` and `~> 3.0`

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_events_producer.ex
Annopaolo: Is it? I see `declare_exchange` being called [here](https://github.com/astarte-platform/astarte/blob/ffad806921d493a4cf4eb7d3105acf423c28ac9b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex#L42)

rbino: I was talking about `:try_to_connect`

Annopaolo:
Superseded by #895 .

PR #635, opened 2021-12-17
Prepare 1.0.1 release

bettio:
Astarte 1.0.1 release chores.


apps/astarte_realm_management/mix.lock
```diff
@@ -1,10 +1,10 @@
 %{
```
rbino, line 1: @bettio I think you forgot to push `mix.exs` for Realm Management

rbino (approved):

PR #664, opened 2022-03-14
DUP: log the key when `unexpected_object_key` occurs

### LINKED ISSUE #663: Data Updater Plant: use more informative logging on unexpected key
When Astarte receives an unexpected object key, it logs the BSON document raw value as a binary, which is not helpful for debugging what is wrong in the object.
Log also the unexpected key (at least).

Annopaolo:
Also, fix a small typo in the log tag.
Closes #663.


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -909,6 +909,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
           {:halt, {:error, reason}}
 
         :error ->
+          Logger.warn("unexpected key #{inspect(key)} in object #{inspect(object)}.")
```
bettio, line 912: I think our habit is using uppercase for the first log line character. Also this Logger.warn has no tag, is it correct?

bettio (approved):

PR #666, opened 2022-03-17
Housekeeping API: add synchronous API for realm operations

Annopaolo:
Allow user to create/delete realms with synchronous calls using the `async_operation` query param. Default to the current behaviour, i.e. async calls.

### DIFF
--- a/apps/astarte_housekeeping_api/lib/astarte_housekeeping_api_web/controllers/realm_controller.ex
+++ b/apps/astarte_housekeeping_api/lib/astarte_housekeeping_api_web/controllers/realm_controller.ex
@@ -29,8 +29,16 @@ defmodule Astarte.Housekeeping.APIWeb.RealmController do
     render(conn, "index.json", realms: realms)
   end
 
-  def create(conn, %{"data" => realm_params}) do
-    with {:ok, %Realm{} = realm} <- Realms.create_realm(realm_params) do
+  def create(conn, %{"data" => realm_params} = params) do
+    async_operation =
+      if Map.get(params, "async_operation") == "false" do
+        false
+      else
+        true
+      end
+
+    with {:ok, %Realm{} = realm} <-
+           Realms.create_realm(realm_params, async_operation: async_operation) do
       conn
       |> put_status(:created)
       |> put_resp_header("location", realm_path(conn, :show, realm))
@@ -51,8 +59,15 @@ defmodule Astarte.Housekeeping.APIWeb.RealmController do
     end
   end
 
-  def delete(conn, %{"id" => id}) do
-    with :ok <- Realms.delete_realm(id) do
+  def delete(conn, %{"id" => id} = params) do
+    async_operation =
+      if Map.get(params, "async_operation") == "false" do
+        false
+      else
+        true
+      end
+
+    with :ok <- Realms.delete_realm(id, async_operation: async_operation) do
       send_resp(conn, :no_content, "")
     end
   end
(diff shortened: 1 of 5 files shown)


rbino (changes requested):
Let's add the new parameter to the Swagger doc

bettio:
Do we really want to merge this into master, instead of release-1.0?

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api_web/controllers/realm_controller.ex
```diff
@@ -29,8 +29,12 @@ defmodule Astarte.Housekeeping.APIWeb.RealmController do
     render(conn, "index.json", realms: realms)
   end
 
-  def create(conn, %{"data" => realm_params}) do
-    with {:ok, %Realm{} = realm} <- Realms.create_realm(realm_params) do
+  def create(conn, %{"data" => realm_params} = info) do
+    # booleans are atoms
+    async_operation = Map.get(info, "async_operation", "true") |> String.to_existing_atom()
```
rbino, line 34: Couple of observations:
- `info` -> `params` (that's their actual name)
- While legitimate, the call to `to_existing_atom` is a little brutal. Either just match on `"true"` and put `false` otherwise with an `if`, or return some kind of informative error (the average user doesn't have access to logs so seeing a 500 won't help them to understand they misspelled `treu`). I'd honestly go for the first one since we already do so in other places (see, e.g., the `details` query param in `AppEngine.APIWeb.DeviceStatusController`).

rbino (changes requested):

bettio (changes requested):

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api_web/controllers/realm_controller.ex
```diff
@@ -29,8 +29,16 @@ defmodule Astarte.Housekeeping.APIWeb.RealmController do
     render(conn, "index.json", realms: realms)
   end
 
-  def create(conn, %{"data" => realm_params}) do
-    with {:ok, %Realm{} = realm} <- Realms.create_realm(realm_params) do
+  def create(conn, %{"data" => realm_params} = params) do
+    async_operation =
+      if Map.get(params, "async_operation") == "false" do
```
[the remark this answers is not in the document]
Annopaolo: I'm not sure of what you mean:
```elixir
iex(1)> Map.get(%{"async_operation" => "0"}, "async_operation") == "false"
false
```
So the call will be async as it was before.

bettio (approved):

apps/astarte_housekeeping_api/priv/static/astarte_housekeeping_api.yaml
```diff
@@ -55,13 +55,22 @@ paths:
...
+        You can perform the call synchronously by setting the async_operation query
+        param to false.
       operationId: createRealm
       security:
         - JWT: []
+      parameters:
+        - name: async_operation
+          in: query
+          description: Whether the operation should be carried out asynchronously.
+          required: false
```
rbino, line 71: Let's add the default value (`true`) here so that it's documented here too

apps/astarte_housekeeping_api/priv/static/astarte_housekeeping_api.yaml
```diff
@@ -137,6 +151,12 @@ paths:
           required: true
           schema:
             type: string
+        - name: async_operation
+          in: query
+          description: Whether the operation should be carried out asynchronously.
+          required: false
```
rbino, line 157: As above

rbino (changes requested):

bettio (approved):

bettio (approved):

PR #669, opened 2022-03-17
Realm Management API: add synchronous API for interface operations

Annopaolo:
Allow user to create, delete or update interfaces in a synchronous fashion using the `async_operation` query param. Default to the current behaviour, i.e. async calls.


bettio:
Do we really want to merge this into master, instead of release-1.0?

bettio (changes requested):
Let's rebase and fix the changelog

bettio (approved):

PR #691, opened 2022-04-11
Appengine API: always consider properties unset succesful

### LINKED ISSUE #640: Unsets should always be considered successful
This is related to #438, which fixed this for properties setting but didn't fix it for properties unset. For the same reason, unsets should always return `200 OK` regardless of the number of matches on the broker because:
- If a session is present, the unset will be sent when the device reconnects
- If a session is not present (or the device asks for a clean one on reconnect), the information of the unset path will be delivered by the `/control/consumerProperties` message.

Incidentally, there's currently another bug because the unset gets saved in the database _before_ checking the return value of `ensure_unset`, but this was not put in a separate issue because after this issue is fixed `ensure_unset` will always succeed by construction, thus making it always valid to save the value to the DB before it is sent.

Reported by @szakhlypa

Annopaolo:
When unsetting properties, do not check if the device has a session on the broker and return a success.
Indeed, the device will receive the unset information anyway (either when it reconnects or in the
`/control/consumerProperties` message).
Closes #640 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   Fix [683](https://github.com/astarte-platform/astarte/issues/683).
 - [astarte_e2e] Allow setting custom subjects for alerting emails.
 - [astarte_e2e] Group in a single thread emails referencing the same failure_id.
+- [astarte_appengine_api] Make property unset succeed independently of whether there exist a device
+  session on the broker or not. Fix [#640](https://github.com/astarte-platform/astarte/issues/640).
 
 ## [1.0.2] - 2022-04-01
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.0.3] - Unreleased
+### Fixed
+- [astarte_appengine_api] Property unset succeeds independently of whether there exist a device
```
rbino, line 9: Property unset succeeds [...] -> Make property unset succeed [...]

Otherwise it's not clear whether you're describing the previous behaviour that was wrong or the new one that is right

rbino (changes requested):

rbino (approved):

bettio (approved):

PR #697, opened 2022-04-19
Replace dialyzex with dialyxir

### LINKED ISSUE #667: Replace dialyzex with dialyxir
In astarte we started using [dialyzex](https://github.com/Comcast/dialyzex) instead of the more popular [dialyxir](https://github.com/jeremyjh/dialyxir) since it supported sharing Erlang and Elixir PLTs across projects and it cut analysis times when using it in multiple projects (like the Astarte microservices).

Now dialyxir supports PLT sharing too, so we might as well use that and enjoy the friendlier error messages.

Annopaolo:
Use a more popular gradual typing tool, with better error messages. Continue caching PLTs in GH workflows.
Close #667.


apps/astarte_appengine_api/mix.exs
```diff
@@ -33,7 +33,9 @@ defmodule Astarte.AppEngine.API.Mixfile do
         "coveralls.post": :test,
         "coveralls.html": :test
       ],
-      dialyzer_cache_directory: dialyzer_cache_directory(Mix.env()),
+      dialyzer: [
+        plt_core_path: "./#{dialyzer_cache_directory(Mix.env())}"
```
rbino, line 37: Doesn't this work without prefixing it with `./`? I would assume paths are relative by default. If it doesn't work I'd fix the `dialyzer_cache_directory` function.

rbino (changes requested):

apps/astarte_appengine_api/mix.exs
Annopaolo: `plt_core_path` is absolute, I changed `dialyzer_cache_directory`

rbino (approved):

PR #706, opened 2022-05-10
AppEngine: do not crash on some malformed POST/PUT requests 

### LINKED ISSUE #264: AppEngine: crash on not server owned interfaces POST/PUT requests
Sending a POST (or PUT) request on a non server owned interface path crashes the API

Example
`POST https://my.astarte.instance.com/appengine/v1/test/devices/912p01EGRqe_ER1lLDn9lA/interfaces/org.astarte-platform.genericsensors.Values/my_sensor/value`

Expected result
405 Method Not Allowed

Actual result
500 - Internal server error

Annopaolo:
Instead of failing, return a 405 Method Not Allowed error if:
- A client tries to publish data on a device-owned interface;
- A client tries to publish data on an incomplete path of a server-owned interface.

Close #264 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [astarte_appengine_api] Consider `allow_bigintegers` and `allow_safe_bigintegers` params
   when querying the root of individual datastream / properties interfaces.
   Fix [#630](https://github.com/astarte-platform/astarte/issues/630).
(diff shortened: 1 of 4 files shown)


Pavinati (approved):

rbino (approved):

rbino (changes requested):
As mentioned in the [linked issue](https://github.com/astarte-platform/astarte/issues/264) the code to be returned should probably be 405 (`put_status(:method_not_allowed`)), since 403 is more centered around auth/authz. The code was crashing before so we're free to change the return code anyway.

Pavinati (approved):

bettio (approved):

PR #720, opened 2022-06-29
DUP: better logs on receving unexpected object key

Annopaolo:
Log the base64-encoded payload instead of the Elixir binary representation.

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
@@ -802,7 +802,12 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
         update_stats(new_state, interface, nil, path, payload)
 
       {:error, :unexpected_object_key} ->
-        Logger.warn("Object has unexpected key: #{inspect(payload)} sent to #{interface}#{path}.",
+        base64_payload = Base.encode64(payload)
+
+        Logger.warn(
+          "Received object with unexpected key, object base64 is: #{base64_payload} sent to #{
+            interface
+          }#{path}.",
           tag: "unexpected_object_key"
         )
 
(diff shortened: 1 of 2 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -802,7 +802,12 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
         update_stats(new_state, interface, nil, path, payload)
 
       {:error, :unexpected_object_key} ->
-        Logger.warn("Object has unexpected key: #{inspect(payload)} sent to #{interface}#{path}.",
+        base64_payload = Base.encode64(payload)
+
+        Logger.warn(
+          "Received object with unexpected key, object base64 is: #{inspect(base64_payload)} sent to #{
```
bettio, line 808: base64_payload is always a binary, so no need to inspect.

bettio (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -802,7 +802,12 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
...
 
       {:error, :unexpected_object_key} ->
-        Logger.warn("Object has unexpected key: #{inspect(payload)} sent to #{interface}#{path}.",
+        base64_payload = Base.encode64(payload)
+
+        Logger.warn(
+          "Received object with unexpected key, object base64 is: #{base64_payload} sent to #{
+            interface
+          }#{path}.",
           tag: "unexpected_object_key"
```
bettio, line 811: also, did we use the same tag for 2 different messages?

bettio (changes requested):

bettio (approved):

PR #733, opened 2022-10-05
Volatile triggers: do not install if target device does not exist

Annopaolo:
Volatile triggers are currently installed even if they refer to a device that does not exist; this causes 
1. DataUpdaterPlant to fail to initialise the data updater process for such device
2. the AppEngine RPC to DataUpdaterPlant to timeout.

Check that the relevant device exists before allowing volatile triggers.

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/room.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/room.ex
@@ -32,6 +32,9 @@ defmodule Astarte.AppEngine.API.Rooms.Room do
   alias Astarte.Core.Triggers.SimpleTriggersProtobuf.SimpleTriggerContainer
   alias Astarte.Core.Triggers.SimpleTriggersProtobuf.TaggedSimpleTrigger
   alias Astarte.Core.Triggers.SimpleTriggersProtobuf.TriggerTargetContainer
+  alias Astarte.Core.Device
+  alias Astarte.DataAccess.Database
+  alias Astarte.AppEngine.API.Rooms.Queries
 
   require Logger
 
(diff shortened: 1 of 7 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/room.ex
```diff
@@ -293,6 +296,47 @@ defmodule Astarte.AppEngine.API.Rooms.Room do
...
+  defp check_device_exists(client, device_id) do
+    with {:error, :database_error} <- Queries.check_device_exists(client, device_id) do
+      Logger.warn(
+        "Database error when checking existence of device #{inspect(device_id)}",
+        tag: "check_device_exists_db_error"
+      )
+
+      {:error, :database_error}
+    end
+  end
```
rbino, line 338: Imho this whole chunk is a little bit confusing, you have `verify_device_exists`, `device_exists?` and `check_device_exists` and it's not clear what's the difference between them without reading the implementation. You should just keep `verify_device_exists` because:
- `check_device_exists` is just catching the error to log, but you're actually already logging inside the function, so you could directly call `Queries.check_device_exists` where you're calling the local `check_device_exists`
- `device_exists?` has a `?` suffix but it doesn't actually return a boolean because it could return `{:error, reason}` or a boolean, which is a strange pattern (usually if you return `{:error, reason}` you return `:ok` or `{:ok, result}`). You should move `device_exists?` body inside `verify_device_exists` and then check result with an `if` (it's a boolean so no need for a `case`).

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater.ex
```diff
@@ -200,4 +206,45 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater do
 
     pid
   end
+
+  defp verify_device_exists(realm_name, device_id) do
```
rbino, line 210: See above

rbino (changes requested):

rbino (approved):

bettio (approved):

PR #749, opened 2023-01-12
TriggerEngine: allow custom prefetch_count on trigger policies

Annopaolo:
Using Trigger Delivery Policies, if AMQP consumer prefetch count > 1, ordered message delivery is not guaranteed because messages might not be dequeued (thus processed) sequentially. However, this allows for higher throughput. 
The `prefetch_count` field allows to specify the AMQP consumer prefetch count as a policy component.
The default policy will still use the `DATA_UPDATER_PLANT_AMQP_CONSUMER_PREFETCH_COUNT` env var, so as not to break compatibility with previous Astarte versions.


Annopaolo:
Based on https://github.com/astarte-platform/astarte_core/pull/83.

PR #750, opened 2023-01-16
RM API: provide useful feedback when trigger action is malformed

### LINKED ISSUE #748: Realm Management error feedback when creating triggers with invalid URLs
When installing a trigger with an invalid URL, Realm Management should return an error message specifying why the entity is not valid. Same as other fields in the Trigger.
eg:
```json
{"errors":{"name":["can't be blank"]}}
```

**Action:**
installing a trigger with an invalid `http_url`

**Expected result:**
```json
{"errors":{"http_url":["is not a valid URL"]}}
```

**Actual result:**
```json
{"errors":{}}
```

Annopaolo:
Display detailed feedback upon failed installation of a trigger with malformed action field.
Under the hood, the Trigger changeset now embeds an `Astarte.RealmManagement.API.Triggers.Action` changeset that wraps a valid AMQP/HTTP action. Data validation is still performed by the two related modules.
The Action modules handles JSON encoding/decoding, too.
Fixes #748.


apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/action.ex
```diff
@@ -0,0 +1,96 @@
...
+          %Action{
+            http_url: http_url,
+            http_method: http_method,
+            http_static_headers: http_static_headers,
+            template: template,
+            template_type: template_type
+          },
+          opts
+        )
+        when not is_nil(http_url) do
```
rbino, line 56: Extract [only what is needed](https://github.com/lucasvegi/Elixir-Code-Smells#complex-extractions-in-clauses) (i.e. `http_url`) in the function head, and extract the rest in the body

apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/action.ex
```diff
@@ -0,0 +1,96 @@
...
+            amqp_exchange: amqp_exchange,
+            amqp_routing_key: amqp_routing_key,
+            amqp_static_headers: amqp_headers,
+            amqp_message_expiration_ms: amqp_message_expiration_ms,
+            amqp_message_persistent: amqp_message_persistent,
+            amqp_message_priority: amqp_message_priority
+          },
+          opts
+        )
+        when not is_nil(amqp_exchange) do
```
rbino, line 78: Extract [only what is needed](https://github.com/lucasvegi/Elixir-Code-Smells#complex-extractions-in-clauses) (i.e. `amqp_exchange`) in the function head, and extract the rest in the body

rbino (changes requested):

rbino (approved):

PR #768, opened 2023-03-30
docker-compose: use a standard URL scheme

### LINKED ISSUE #761: Expose docker-compose services with a single standard endpoint URL
When running Astarte services via docker-compose, e.g. when following the *Astarte in 5 minutes* guide, the services are exposed via different dedicated endpoints.
This kind of setup could be improved to address the following points:
- The experience of using Astarte with docker-compose differs from the experience of usual deployments, e.g. with Kubernetes. This difference requires additional operational cost when migrating from a development environment to a production one.
- Integrating external services with Astarte could be an easier task if they could rely on a standard interface for Astarte services and if configuring a single endpoint would suffice to query all Astarte APIs.

Annopaolo:
Expose Astarte with Traefik. Use the following URL scheme:
- API: `http://api.astarte.localhost/<service name>`
- Broker: `mqtts://broker.astarte.localhost`
- Dashboard: `http://dashboard.astarte.localhost`

The old URLs (e.g. `http://localhost:4002`) will not work anymore.
Contextually, give an explicit name to the docker network used by services.
Close #761.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,9 +5,7 @@ services:
     build:
       context: apps/astarte_housekeeping
     env_file:
-     - ./compose.env
-    ports:
-      - "4008:4000"
+      - ./compose.env
     restart: on-failure
     depends_on:
       - "rabbitmq"
@@ -18,11 +16,9 @@ services:
     build:
       context: apps/astarte_housekeeping_api
     env_file:
-     - ./compose.env
+      - ./compose.env
     environment:
       HOUSEKEEPING_API_JWT_PUBLIC_KEY_PATH: "/keys/housekeeping_public.pem"
-    ports:
-      - "4001:4001"
     volumes:
       - type: bind
         source: ./compose/astarte-keys/housekeeping_public.pem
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -190,26 +252,31 @@ services:
...
-      # You can comment this out if you don't need Let's Encrypt
-      - "80:80"
-      # HTTP Listener for health check
-      - "8888:8888"
+    # ports:
+    #   # HTTP Listener for health check
+    #   - "8888:8888"
+    #   # 1883 is used internally and isn't exposed
+    #   # Used with HAProxy only
+    #   - "1885:1885"
```
Pavinati, line 260: I wouldn't rely on a docker compose comment for documentation. I would remove this from the file and rely on our doc to retrieve the same information

docker-compose.yml
```diff
@@ -138,20 +164,56 @@ services:
...
+      - "--entrypoints.web.address=:80"
+      # Expose Astarte broker
+      - "--entryPoints.vernemq.address=:8883"
+    ports:
+      # The HTTP port
+      - "80:80"
+      # The Web UI (enabled by --api.insecure=true)
+      - "8080:8080"
+      # The HTTPS port
+      - "443:443"
```
Pavinati, line 211: We're not using https. is this required anyway?

docker-compose.yml
Annopaolo: There's a bunch of documentation comments here and there in many docker-compose-related files (e.g. `compose.env`). Shall we remove all of them in another PR?

davidebriani (approved):

matt-mazzucato (commented):
Nice enhancement! In general it LGTM.
Anyway, some of the containers are missing the `restart` policy. Is it something that we want to add? The default value is `restart: no`.

Annopaolo:
> Nice enhancement! In general it LGTM. Anyway, some of the containers are missing the `restart` policy. Is it something that we want to add? The default value is `restart: no`.

@matt-mazzucato if the astarte-dashboard and astarte-grafana containers are crashing, it is not because of other Astarte services, so a restart policy different than `no` would have no effect. It is better to let those containers crash and analyze the reason why the crash happened.

matt-mazzucato:
> > Nice enhancement! In general it LGTM. Anyway, some of the containers are missing the `restart` policy. Is it something that we want to add? The default value is `restart: no`.
> 
> @matt-mazzucato if the astarte-dashboard and astarte-grafana containers are crashing, it is not because of other Astarte services, so a restart policy different than `no` would have no effect on the failure. It is better to let those containers crash and analyze the reason why the crash happened.

I see that a restart policy was added to traefik and scylla. LGTM

davidebriani (approved):

rbino (approved):

PR #774, opened 2023-04-07
RM API: Return a correctly-formatted Trigger Delivery Policy

### LINKED ISSUE #766: RM API: Types of `"on"` parameter inside `"error_handlers"` are handled incorrectly for Trigger Delivery Policy
It would be necessary to remove the extra map from the `"on"` parameter.

If you choose any of ("any_error", client_error" or "server_error") the response body will put that string in a new map where `"keyword"` is represented as the key.

Screenshot from 2023-03-24 15-47-34 Screenshot from 2023-03-24 15-47-53

If you choose any of ( [400 - 499] ) the response body will put the list in a new map where `"error_codes"` is represented as the key.

Screenshot from 2023-03-24 15-51-14 Screenshot from 2023-03-24 15-51-24

arahmarchak:
Display incorrect and excessive parameters solved by excluded ```"keyword"``` and ```"error_codes"``` from ```"error_handlers"``` field to get correctly-formatted Trigger Policy.

Modified ```trigger_policy_view.ex``` and added ```keys: :atoms!``` option in ```trigger_policy_controller.ex``` to use (safe) atoms as keys to simplify handler normalization in Trigger Policy View.

Fix #766.

### DIFF
--- a/apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/controllers/trigger_policy_controller.ex
+++ b/apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/controllers/trigger_policy_controller.ex
@@ -51,7 +51,9 @@ defmodule Astarte.RealmManagement.APIWeb.TriggerPolicyController do
 
   def show(conn, %{"realm_name" => realm_name, "id" => id}) do
     with {:ok, policy_source} <- Policies.get_trigger_policy_source(realm_name, id),
-         {:ok, decoded_json} <- Jason.decode(policy_source) do
+         # Use (safe) atoms as keys to simplify handler normalization in Trigger Policy View
+         # TODO: move this to a function in Astarte Core building a Policy from its source
+         {:ok, decoded_json} <- Jason.decode(policy_source, keys: :atoms!) do
       render(conn, "show.json", policy: decoded_json)
     end
   end
--- a/apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/views/trigger_policy_view.ex
+++ b/apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/views/trigger_policy_view.ex
@@ -19,6 +19,7 @@
 defmodule Astarte.RealmManagement.APIWeb.TriggerPolicyView do
   use Astarte.RealmManagement.APIWeb, :view
   alias Astarte.RealmManagement.APIWeb.TriggerPolicyView
+  alias Astarte.Core.Triggers.Policy.Handler
 
   def render("index.json", %{policies: policies}) do
     %{data: render_many(policies, TriggerPolicyView, "trigger_policy_name_only.json")}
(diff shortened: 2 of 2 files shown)


apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/controllers/trigger_policy_controller.ex
```diff
@@ -51,7 +51,7 @@ defmodule Astarte.RealmManagement.APIWeb.TriggerPolicyController do
 
   def show(conn, %{"realm_name" => realm_name, "id" => id}) do
     with {:ok, policy_source} <- Policies.get_trigger_policy_source(realm_name, id),
-         {:ok, decoded_json} <- Jason.decode(policy_source) do
+         {:ok, decoded_json} <- Jason.decode(policy_source, keys: :atoms!) do
```
Annopaolo, line 54: Put a comment on why `keys: :atoms!` was added

Annopaolo (changes requested):

Annopaolo (changes requested):

apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/views/trigger_policy_view.ex
```diff
@@ -33,10 +34,30 @@ defmodule Astarte.RealmManagement.APIWeb.TriggerPolicyView do
   end
 
   def render("policy.json", %{trigger_policy: policy}) do
-    policy
+    %{error_handlers: error_handlers} = policy
+
+    normalized_handlers =
+      error_handlers
+      |> Enum.map(&normalize_handler/1)
```
Annopaolo, line 41: When there's only a single function, it's better not to use the pipe operator

Annopaolo (changes requested):

Annopaolo (approved):

apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/controllers/trigger_policy_controller.ex
```diff
@@ -51,7 +51,8 @@ defmodule Astarte.RealmManagement.APIWeb.TriggerPolicyController do
 
   def show(conn, %{"realm_name" => realm_name, "id" => id}) do
     with {:ok, policy_source} <- Policies.get_trigger_policy_source(realm_name, id),
-         {:ok, decoded_json} <- Jason.decode(policy_source) do
+         # Use (safe) atoms as keys to simplify handler normalization in Trigger Policy View
```
rbino, line 54: This is ok as a temporary workaround, but in the long run we should add a `from_source` function to `Astarte.Core.Trigger.Policy`. So please add another line with:
```
# TODO: move this to a function in Astarte Core building a Policy from its source
```
so we don't forget to do this in a following iteration

rbino (changes requested):

rbino (approved):

PR #782, opened 2023-05-15
DUP: Use the `internal` event type for Astarte internal messages (e.g. device heartbeat).

Annopaolo:
Support the new `internal` message type (see https://github.com/astarte-platform/astarte_vmq_plugin/pull/71). For now, only the device heartbeat message is internal. Support the old heartbeat type, too, for the time being.

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
@@ -153,6 +153,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
     %{new_state | connected: true, last_seen_message: timestamp}
   end
 
+  # TODO make this private when all heartbeats will be moved to internal
   def handle_heartbeat(state, message_id, timestamp) do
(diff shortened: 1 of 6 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -166,6 +167,51 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
...
+    {:ok, db_client} = Database.connect(realm: state.realm)
+
+    new_state = execute_time_based_actions(state, timestamp, db_client)
+
+    Queries.maybe_refresh_device_connected!(db_client, new_state.device_id)
+
+    MessageTracker.ack_delivery(new_state.message_tracker, message_id)
+    Logger.info("Device heartbeat.", tag: "device_heartbeat")
+
+    %{new_state | connected: true, last_seen_message: timestamp}
```
rbino, line 180: Just call `handle_heartbeat(state, message_id, timestamp)`, and switch the `TODO` to `make this private when...`

rbino (changes requested):

bettio (approved):

PR #787, opened 2023-05-29
Init nix for development

noaccOS:
Init nix's development shell as an alternative to asdf. It can be used from both a standard nix-shell
or a nix develop with flakes.

MIX_HOME and HEX_HOME are set do directories in the project root to avoid conflicts with system packages.

~.envrc is provided for use with direnv, defaulting to the flake version of the shell.~

~To use environment variables in development, eg for testing, it is possible to create a `.env` file with said variables.
The syntax for said file can be found [here](https://github.com/bkeepers/dotenv) but in general it consists of lines of `VAR=value` pairs.~

[elixir-ls](https://github.com/elixir-lsp/elixir-ls) is provided (by default), with hooks tested working  for vscode and emacs (provided you don't hadn't previously installed elixir-ls through emacs' lsp-mode because that one takes priority)

To use the provided lsp server, you need to either
- start your editor from the `nix develop` environment
- have a `.envrc` with either `use flake` or `use nix` and a direnv plugin in your editor ([vscode](https://github.com/direnv/direnv-vscode), [emacs](https://github.com/wbolster/emacs-direnv))

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,5 +44,9 @@ deps
 .rebar3
 apps/*/_checkouts/
 
+# Direnv
+.direnv/
+.envrc
+
 examples/*/node_modules/
 examples/*/.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,10 @@ deps
 .rebar3
 apps/*/_checkouts/
 
+# Nix
+.nix-mix/
+.nix-hex/
+
 # Direnv
 .direnv/
 .envrc
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,98 @@
+{
+  "nodes": {
+    "elixir-utils": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1696521035,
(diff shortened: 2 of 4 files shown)


.nix/beam-overlay.nix
```diff
@@ -0,0 +1,11 @@
+final: prev:
+{
+  erlang = (prev.erlang_23.override {
+    version = "23.2.7";
+    sha256 = "sha256-uUWLUamo3N3whtkzsG8lG1eSjA1VP3Y5SObSYHFAI2s=";
+  });
+  elixir = ((prev.beam.packagesWith final.erlang).elixir_1_11.override {
+    version = "1.11.4";
```
bettio, line 8: 1.14.5

.nix/beam-overlay.nix
```diff
@@ -0,0 +1,11 @@
+final: prev:
+{
+  erlang = (prev.erlang_23.override {
+    version = "23.2.7";
```
bettio, line 4: 25.3.2

bettio (changes requested):

.nix/beam-overlay.nix
noaccOS: dependencies should now be up to date

Annopaolo (approved):

Annopaolo (approved):

PR #788, opened 2023-06-01
Optional notifications

### LINKED ISSUE #741: astarte_e2e: reduce noise when performing maintenance tasks on astarte
When maintenance tasks are being performed on an Astarte instance, temporary service interruptions are expected. In these scenarios, astarte_e2e should not raise alerts. Thus, astarte_e2e should provide the possibility of enabling/disabling alerts.

noaccOS:
Allows email notifications to be disabled.

Closes #741

### DIFF
--- a/tools/astarte_e2e/lib/astarte_e2e/config.ex
+++ b/tools/astarte_e2e/lib/astarte_e2e/config.ex
@@ -1,7 +1,7 @@
 #
 # This file is part of Astarte.
 #
-# Copyright 2020 Ispirata Srl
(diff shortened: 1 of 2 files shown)


noaccOS:
Will rebase and squash once ready

noaccOS:
I have used the config option `:mail_enabled` to allow the user to disable the notifications from the config, but when disabled this way it can't be manually re-enabled without changing the config and reloading manually, which might serve our use case but I'm not sure is the best approach.

Issue #772, opened 2023-04-07
Introduce a limit on the number of registered devices per realm


Annopaolo:
In order to allow for more advanced use cases of Astarte and to have fairness in resource consumption, a limit on the maximum number of registered devices per realm should be introduced. The value should be configurable.
When such limit is reached, user won't be able to register more devices, and therefore this issue is linked to #392.

PR #814, opened 2023-07-04
Introduce a limit on the number of registered devices per realm

Annopaolo:
See #772 for details.

### DIFF
--- a/apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
+++ b/apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
@@ -22,16 +22,29 @@ defmodule Astarte.Housekeeping.Engine do
   alias Astarte.Housekeeping.Config
   alias Astarte.Housekeeping.Queries
 
-  def create_realm(realm, public_key_pem, replication_factor, opts \\ []) do
+  def create_realm(
+        realm,
+        public_key_pem,
+        replication_factor,
+        device_registration_limit,
+        opts \\ []
+      ) do
     _ =
       Logger.info(
         "Creating new realm.",
         tag: "create_realm",
         realm: realm,
-        replication_factor: replication_factor
+        replication_factor: replication_factor,
+        device_registration_limit: device_registration_limit
       )
 
-    Queries.create_realm(realm, public_key_pem, replication_factor, opts)
+    Queries.create_realm(
+      realm,
+      public_key_pem,
+      replication_factor,
+      device_registration_limit,
+      opts
+    )
   end
 
   def get_health do
--- a/apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
+++ b/apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
@@ -99,12 +99,16 @@ defmodule Astarte.Housekeeping.Engine do
 
   @doc """
   Updates a given realm using the provided values map. Fails if realm does not exists or values are invalid.
-  At the moment, only updating the realm jwt_public_key_pem is supported.
+  At the moment, updates to jwt_public_key_pem and device_registration_limit are supported.
   Returns a tuple:
   - either {:ok, updated_realm} where updated_realm is a map describing the realm
   - or {:error, reason} where reason is an atom describing the error.
   """
-  @spec update_realm(String.t(), %{:jwt_public_key_pem => String.t(), optional(any()) => any()}) ::
+  @spec update_realm(String.t(), %{
+          :jwt_public_key_pem => String.t(),
+          :device_registration_limit => any(),
+          optional(any()) => any()
+        }) ::
           {:ok, map()}
           | {:error, :invalid_update_parameters}
           | {:error, :realm_not_found}
@@ -125,17 +129,14 @@ defmodule Astarte.Housekeeping.Engine do
 
   defp do_update_realm(realm_name, update_attrs) do
     %{
-      jwt_public_key_pem: new_jwt_public_key_pem
+      jwt_public_key_pem: new_jwt_public_key_pem,
+      device_registration_limit: new_limit
     } = update_attrs
 
     with realm when is_map(realm) <- Queries.get_realm(realm_name),
-         :ok <- update_jwt_public_key_pem(realm_name, new_jwt_public_key_pem) do
-      updated_realm =
-        if new_jwt_public_key_pem != nil do
-          Map.put(realm, :jwt_public_key_pem, new_jwt_public_key_pem)
-        else
-          realm
-        end
+         :ok <- update_jwt_public_key_pem(realm_name, new_jwt_public_key_pem),
+         :ok <- update_device_registration_limit(realm_name, new_limit) do
+      updated_realm = merge_realm_status(realm, update_attrs)
 
       _ = Logger.info("Successful update of realm #{realm_name}", tag: "realm_update_success")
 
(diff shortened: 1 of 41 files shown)


doc/pages/user/020-accessing_and_exploring_a_realm.md
```diff
@@ -1,3 +1,43 @@
...
+`PUT <astarte base API URL>/realmmanagement/v1/<realm name>/config/devices`
+
+The HTTP payload of the request must have the following format:
+
+```json
+{"data": {"limit": <value>}}
+```
+
+The value must be a positive integer. If more Devices are registered than
+the requested limit, the HTTP call will fail.
```
Pavinati, line 25:
```suggestion
The value must be a positive integer. Trying to register more Devices past
the realm limit will result in an error.
```

apps/astarte_realm_management_api/priv/static/astarte_realm_management_api.yaml
```diff
@@ -776,6 +853,21 @@ components:
...
+            type: object
+            required:
+              - data
+            properties:
+              data:
+                type: object
+                properties:
+                  limit:
+                    type: integer
+                    example: 100
```
davidebriani, line 869: We could add `minimum: 0` here too.

doc/pages/user/020-accessing_and_exploring_a_realm.md
```diff
@@ -1,3 +1,43 @@
...
+
+```json
+{"data": {"limit": <value>}}
+```
+
+The value can be either a positive number or `"infinity"`.
+
+### Removing the limit on the number of registered Devices in a Realm
+
+`DELETE <astarte base API URL>/realmmanagement/v1/<realm name>/config/devices`
```
Pavinati, line 41: I'm not a fan of using a `DELETE` here.
Here we're are unsetting a value, not removing the entity `device` from config.
I'd rather have a `PUT` or `PATCH` setting the limit to `null` which is consistent with other Astarte APIs

doc/pages/user/020-accessing_and_exploring_a_realm.md
```diff
@@ -1,3 +1,43 @@
...
+
+`PUT <astarte base API URL>/realmmanagement/v1/<realm name>/config/devices`
+
+The HTTP payload of the request must have the following format:
+
+```json
+{"data": {"limit": <value>}}
+```
+
+The value must be a positive integer. If more Devices are registered than
```
davidebriani, line 24: must be a ~~positive~~**non negative** integer

doc/pages/user/020-accessing_and_exploring_a_realm.md
```diff
@@ -1,3 +1,43 @@
...
+
+`GET <astarte base API URL>/realmmanagement/v1/<realm name>/config/devices`
+
+The HTTP payload of the response will have the following format:
+
+```json
+{"data": {"limit": <value>}}
+```
+
+The value can be either a positive number or `"infinity"`.
```
davidebriani, line 37: a ~~positive number~~**non negative integer**

davidebriani (changes requested):

doc/pages/user/020-accessing_and_exploring_a_realm.md
Annopaolo: The two are two sided of the double implication. Both are to be included

Pavinati: Oh now i see what you meant. I would use `value` instead of `requested limit` in the following sentence too.
What about:
```
If the value is less than the number of currently registered devices the HTTP call will fail.
```

Annopaolo:
This PR is based, based on #857

doc/pages/administrator/070-manage_realms.md
```diff
@@ -13,6 +13,50 @@ kubectl get secret -n astarte astarte-housekeeping-private-key -o=jsonpath={.dat
...
+```json
+{
+    "data": {
+        "device_registration_limit": <value>
+    }
+}
+```
+
+To set or update the limit, the value must be a non negative integer.
+If the value is less than the number of currently registered devices the HTTP call will fail.
```
rbino, line 37: IIRC this is not the actual behavior anymore

doc/pages/user/020-accessing_and_exploring_a_realm.md
```diff
@@ -1,3 +1,26 @@
...
-In Astarte, a Realm is a logical partition which holds a number of devices and an Authentication Realm 
+In Astarte, a Realm is a logical partition which holds a number of devices and an Authentication
+Realm.
+
+The [Astarte Dashboard](015-astarte_dahsboard) allows you to explore all resources of a given Realm,
+such as e.g. Interfaces, Triggers, Devices, Groups etc...
+
+## Device limit in a Realm
+While there is no limit on the number of Devices registered in a Realm, it is possible that an
+upper bound has been set in the Realm configuration using the [Housekeeping API]().
```
rbino, line 11: Missing link

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/rpc/housekeeping.ex
```diff
@@ -32,7 +32,9 @@ defmodule Astarte.Housekeeping.API.RPC.Housekeeping do
     GetRealmsList,
     GetRealmsListReply,
     Reply,
-    UpdateRealm
+    UpdateRealm,
+    RemoveLimit,
+    SetLimit
```
rbino, line 37: Since the list is in alphabetical let's keep that intact

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/realms/device_registration_limit_type.ex
```diff
@@ -0,0 +1,38 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Housekeeping.API.Realms.DeviceRegistrationLimitType do
+  use Ecto.Type
+  def type, do: :any
+
+  def cast(:remove), do: {:ok, :remove}
```
rbino, line 23: I'd use `:remove_limit` as atom here, as you do in other places. This makes it more readable because it's consistent.

apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
```diff
@@ -161,6 +162,48 @@ defmodule Astarte.Housekeeping.Engine do
...
 
+  defp update_device_registration_limit(realm_name, new_limit) do
+    _ = Logger.info("Updating device registration limit", tag: "update_device_registration_limit")
+
+    case new_limit do
+      nil -> :ok
+      :remove_limit -> delete_device_registration_limit(realm_name)
+      n when is_integer(n) -> set_device_registration_limit(realm_name, n)
+    end
+  end
```
rbino, line 173: You're logging even if you don't actually update the limit. Pattern match in the function head and map directly to the noop/delete/set implementations, so you can also have a more precise log ("Removing/Updating device registration limit")

apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
```diff
@@ -176,4 +219,20 @@ defmodule Astarte.Housekeeping.Engine do
 
     update_valid?
   end
+
+  defp merge_realm_status(old_realm, %{device_registration_limit: nil} = update_attrs) do
+    attrs = Map.drop(update_attrs, [:device_registration_limit])
```
rbino, line 224: Nit: if you have a single key to drop you can just use `Map.delete`

apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
```diff
@@ -176,4 +219,20 @@ defmodule Astarte.Housekeeping.Engine do
...
+  end
+
+  defp merge_realm_status(old_realm, %{device_registration_limit: :remove_limit} = update_attrs) do
+    attrs = Map.drop(update_attrs, [:device_registration_limit])
+    merge_realm_status(%{old_realm | device_registration_limit: nil}, attrs)
+  end
+
+  defp merge_realm_status(old_realm, update_attrs) do
+    Map.merge(old_realm, update_attrs, fn _k, old_value, new_value ->
+      if new_value == nil, do: old_value, else: new_value
```
rbino, line 235: More compact:
`new_value || old_value`

rbino (changes requested):

apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/views/realm_config_view.ex
```diff
@@ -26,6 +26,12 @@ defmodule Astarte.RealmManagement.APIWeb.RealmConfigView do
     }
   end
 
+  def render("show.json", %{device_registration_limit: limit}) do
+    %{
+      data: %{device_registration_limit: limit}
```
rbino, line 31: Shouldn't this just be `data: limit`? The auth config doesn't have an extra `%{data: %{auth_config: ...}}`

doc/pages/administrator/070-manage_realms.md
Annopaolo: Good catch

rbino (changes requested):

apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
Annopaolo: More compact, but at what cost? We are sacrificing clarity (the new_value *might* be the atom nil) for 2 lines of code that feel like `new_value` and `old_value` are booleans. Even if `||` is idiomatic Elixir, I would prefer the explicit comparison, in order to better convey the type information

bettio (approved):
I didn't go in the detail with the code, but the general approach looks good.

rbino (approved):

PR #834, opened 2023-08-09
refactor(pairing): port queries to xandra

### LINKED ISSUE #268: Port astarte_pairing to Xandra
In light of the gradual demise of CQEx in favor of Xandra, `astarte_pairing` should be adapted to use Xandra instead of CQEx.

This mainly translates to:
- Add Xandra as a dependency
- Start the Xandra connection
- Translate all queries from CQEx to Xandra, both in `lib` and in `test`.

To see an example of a service already using Xandra, have a look at `astarte_housekeeping` (and this PR https://github.com/astarte-platform/astarte/commit/8a5fbd52aecd where the replacement of CQEx with Xandra happened)

noaccOS:
use xandra in place of cqex for all queries to the cluster.

feat: expose custom query API
feat: expose cluster name as an env var internally to the config
fix: the tests now wait for the cluster to be connected
fix: remove :autodiscovery key from xandra_options as it is deprecated
chore: update xandra to v0.17.0

closes #268


apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -338,4 +367,109 @@ defmodule Astarte.Pairing.Queries do
...
+          {:error, :database_error}
+      end
+    end
+  end
+
+  defp use_realm(_conn, nil = _realm), do: :ok
+
+  defp use_realm(conn, realm) when is_binary(realm) do
+    with true <- Realm.valid_name?(realm),
+         {:ok, %Xandra.SetKeyspace{}} <- Xandra.execute(conn, "USE #{realm}") do
```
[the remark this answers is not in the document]
rbino: The reason we did it with string interpolation is that there was an issue with how Xandra handled `USE` queries in older versions (see https://github.com/lexhide/xandra/pull/180).
The PR took some time to be merged so we kept on using string interpolation, but now that it's upstreamed we should be fine using `USE` queries.

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -21,172 +21,190 @@ defmodule Astarte.Pairing.Queries do
...
+
+  @typedoc """
+  A string representing the current context, to include in error messages.
+  Not included if nil.
+  Default: nil
+  """
+  @type custom_context() :: String.t() | nil
+
+  @typedoc """
+  The type to cast the result in.
```
Annopaolo, line 36: The fact that the type is casted is an implementation detail that could be hidden

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -338,4 +332,111 @@ defmodule Astarte.Pairing.Queries do
...
+        Enum.at(page, 0)
+        |> case do
+          nil -> {:error, error}
+          first -> {:ok, first}
+        end
+    end
+  end
+
+  defp database_error_message(%Xandra.Error{message: message, reason: reason}, nil = _context) do
+    %{message: "Database error #{reason}: #{message}", tag: "db_error"}
```
Annopaolo, line 422: Since Xandra errors implement the Exception behaviour, you could use `Exception.message/1` rather than manually getting into the struct and retrieving message and reason

Annopaolo (approved):

PR #837, opened 2023-08-22
refactor(trigger_engine): port query to xandra

### LINKED ISSUE #459: Trigger Engine crash when any of the cassandra nodes are not available
In a setup where trigger engine is connected to a pool of cassandra nodes, whenever any of the cassandra nodes becomes unavailable, Trigger Engine crashes and restarts until the said node becomes available again

noaccOS:
yes there was only one query

closes #459


apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -248,27 +245,38 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
...
+         {:ok, result} <- Xandra.Cluster.execute(:xandra, prepared, params),
+         %{"value" => trigger_data} <-
+           Enum.at(result, 0, {:error, :invalid_query_result, result}),
          trigger <- Trigger.decode(trigger_data),
          {:ok, action} <- Jason.decode(trigger.action) do
       {:ok, action}
     else
-      {:error, :database_connection_error} ->
-        Logger.warn("Database connection error.")
+      {:error, %Xandra.ConnectionError{action: action, reason: reason}} ->
```
Annopaolo, line 262: The `{:error, %Xandra.Error{}}` case is not handled. It would be good to keep database and logic error (e.g. trigger not found) distinct

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -248,27 +245,38 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
...
         Logger.warn("Error while processing event: #{inspect(error)}")
         {:error, :trigger_not_found}
     end
   end
+
+  defp check_realm_name_ok(realm_name) do
+    if Astarte.Core.Realm.valid_name?(realm_name) do
+      :ok
+    else
+      {:error, :realm_not_found, realm_name}
```
Annopaolo, line 279: I'd say to use a more explicit `:invalid_realm_name` atom

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -248,27 +245,38 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
...
-      |> DatabaseQuery.put(:trigger_id, trigger_id)
+    statement =
+      "SELECT value FROM #{realm_name}.kv_store WHERE group='triggers' AND key=:trigger_id;"
+
+    params = %{"trigger_id" => trigger_id}
 
-    with {:ok, client} <- Database.connect(realm: realm_name),
-         {:ok, result} <- DatabaseQuery.call(client, query),
-         [value: trigger_data] <- DatabaseResult.head(result),
+    with :ok <- check_realm_name_ok(realm_name),
```
Annopaolo, line 253: The case of an invalid realm name could be handled explicitly rather than returning `{:error, :trigger_not_found}` which can be misleading

Annopaolo (changes requested):

apps/astarte_trigger_engine/lib/astarte_trigger_engine/application.ex
```diff
@@ -1,7 +1,7 @@
 #
 # This file is part of Astarte.
 #
-# Copyright 2017 Ispirata Srl
+# Copyright 2023 SECO Mind Srl
```
Annopaolo:
```suggestion
# Copyright 2017-2023 SECO Mind Srl
```

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -248,27 +245,52 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
...
     else
-      {:error, :database_connection_error} ->
-        Logger.warn("Database connection error.")
+      {:error, error = %Xandra.Error{}} ->
+        Logger.warning("Database error: #{Exception.message(error)}", tag: "database_error")
+
+        {:error, :database_error}
+
+      {:error, error = %Xandra.ConnectionError{}} ->
+        Logger.warning("Database connection error while #{Exception.message(error)}",
```
Annopaolo, line 265: I think this could be clearer and more coherent with the case just above
```suggestion
        Logger.warning("Database connection error: #{Exception.message(error)}",
```

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -248,27 +245,52 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
...
+          tag: "database_connection_error"
+        )
+
         {:error, :database_connection_error}
 
+      error ->
+        error
```
Annopaolo: This case is redundant, as the error is already handled in the `parse_trigger_data`/`check-realm_name` functions

apps/astarte_trigger_engine/lib/astarte_trigger_engine/config.ex
```diff
@@ -1,7 +1,7 @@
 #
 # This file is part of Astarte.
 #
-# Copyright 2017 Ispirata Srl
+# Copyright 2023 SECO Mind Srl
```
Annopaolo: See above for copyright

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -248,27 +245,53 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
...
 
       error ->
-        Logger.warn("Error while processing event: #{inspect(error)}")
+        error
+    end
+  end
+
+  defp parse_trigger_data(xandra_page) do
+    with %{"value" => trigger_data} <-
+           Enum.at(xandra_page, 0, {:error, :invalid_query_result, xandra_page}),
```
Annopaolo, line 278: I'm not sure we want to return the whole Xandra page in case of an error here

Annopaolo (changes requested):

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
noaccOS: We're not, we're returning `{:error, :trigger_not_found}` from the `else`
Or you mean we don't want to `inspect` it?

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
noaccOS: oversight on my part, done

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
Annopaolo: Right, my mistake

Issue #807, opened 2023-06-16
non-root user docker container


bettio:
Add `USER nobody` to the `Dockerfile`.

PR #842, opened 2023-09-07
Update docker images to `Debian bookworm` and add non-root user

### LINKED ISSUE #806: Update docker images to Debian "bookworm"
Use Debian "bookworm" as base image.
See also: https://github.com/hexpm/bob/pull/167

Hibe7:
- Update docker images to Debian "bookworm".
- Contextually, move from libssl1.1 to the default Debian libssl package
- Add `USER nobody` to the Dockerfiles.

Closes #806, #807

### DIFF
--- a/apps/astarte_appengine_api/Dockerfile
+++ b/apps/astarte_appengine_api/Dockerfile
@@ -32,6 +32,8 @@ FROM debian:bookworm-slim
 
 WORKDIR /app
 
+RUN chown -R nobody /app
+
 RUN apt-get -qq update
 
 # Set the locale
@@ -44,5 +46,8 @@ RUN apt-get -qq install openssl ca-certificates
 ARG BUILD_ENV=prod
 
 COPY --from=builder /app/_build/$BUILD_ENV/rel/astarte_appengine_api .
+
+# Change to non-root user
+USER nobody
 
 CMD ["./bin/astarte_appengine_api", "start"]
--- a/apps/astarte_appengine_api/Dockerfile
+++ b/apps/astarte_appengine_api/Dockerfile
@@ -1,4 +1,4 @@
-FROM hexpm/elixir:1.15.7-erlang-26.1-debian-bullseye-20230612-slim as builder
+FROM hexpm/elixir:1.15.7-erlang-26.1-debian-bookworm-20230612-slim as builder
 
 # install build dependencies
 # --allow-releaseinfo-change allows to pull from 'oldstable'
@@ -28,7 +28,7 @@ ADD . .
 RUN mix do compile, release
 
 # Note: it is important to keep Debian versions in sync, or incompatibilities between libcrypto will happen
-FROM debian:bullseye-slim
+FROM debian:bookworm-slim
 
 WORKDIR /app
 
@@ -38,7 +38,7 @@ RUN apt-get -qq update
 ENV LANG C.UTF-8
 
 # We need SSL
-RUN apt-get -qq install libssl1.1
+RUN apt-get -qq install openssl ca-certificates
 
 # We have to redefine this here since it goes out of scope for each build stage
 ARG BUILD_ENV=prod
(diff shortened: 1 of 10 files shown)


apps/astarte_appengine_api/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM hexpm/elixir:1.14.5-erlang-25.3.2-debian-bullseye-20230522-slim as builder
+FROM hexpm/elixir:1.15.5-erlang-24.3.4.9-debian-bookworm-20230612-slim as builder
```
Annopaolo, line 1: This is changing both Elixir and OTP version, other than changing the distro. I would expect this line to be something like
```
FROM hexpm/elixir:1.14.5-erlang-25.3.2-debian-bookworm-YYYYMMDD-slim as builder
```

apps/astarte_appengine_api/Dockerfile
```diff
@@ -28,21 +28,26 @@ ADD . .
...
+RUN chown -R nobody /app
+
 RUN apt-get -qq update
 
 # Set the locale
 ENV LANG C.UTF-8
 
 # We need SSL
-RUN apt-get -qq install libssl1.1
+RUN apt-get -qq install libssl3
```
Annopaolo, line 43: Is updating libssl needed?

apps/astarte_appengine_api/Dockerfile
```diff
@@ -28,21 +28,26 @@ ADD . .
...
 
 # We have to redefine this here since it goes out of scope for each build stage
 ARG BUILD_ENV=prod
 
 COPY --from=builder /app/_build/$BUILD_ENV/rel/astarte_appengine_api .
 
+# Change to non-root privilege
```
Annopaolo, line 50:
```suggestion
# Change to non-root user
```

Annopaolo (changes requested):
Thank you @Hibe7! There are some issues in Dockerfiles, I commented just one because they're all the same.

Also, could you please run Astarte in 5 minutes with the images built after your changes?

apps/astarte_appengine_api/Dockerfile
Hibe7: When i build images with libssl1.1 it breaks and i found that libssl1.1 package does not exist in Debian bookworm. It has been replaced by libssl3. And when I update it to libssl3 everything works and I run astarte in 5 minutes without any problems.

Annopaolo: Ok! I noticed that phoenix uses just `libssl` (without version), see https://github.com/phoenixframework/phoenix/blob/a66e7c33019a3d564c4f7f2e48b77efb63c54ad7/priv/templates/phx.gen.release/Dockerfile.eex#L71

Could you please try with that one, too?

Hibe7: I have change it and it work as well. Changed in all the docker files.

Annopaolo (changes requested):
Please, also add a CHANGELOG.md line specifying at least that the container user has changed.
Also, it would be great if commit titles were somewhat more meaningful, e.g. `Update docker images` ---> `Apps: update Debian to Bookworm in Dockerfile`

Annopaolo (changes requested):
Almost there! Please also refine the second commit (title and description)

apps/astarte_appengine_api/Dockerfile
```diff
@@ -28,21 +28,26 @@ ADD . .
 RUN mix do compile, release
 
 # Note: it is important to keep Debian versions in sync, or incompatibilities between libcrypto will happen
-FROM debian:bullseye-slim
+FROM debian:bookworm
```
Annopaolo, line 31: Let's keep the image small (and consistent with the builder one):
```suggestion
FROM debian:bookworm-slim
```

Annopaolo (changes requested):
LGTM! Just one small change: we should use the slim Debian image to keep the image small (as suggested by https://docs.docker.com/develop/dev-best-practices)

Annopaolo (approved):

PR #843, opened 2023-09-15
Add "Never Connected" and "Disconnected" fields

### LINKED ISSUE #157: Stats should also have a "Never Connected" field
Stats should show these counts:

* total devices
* connected devices
* disconnected devices
* never connected devices

Hibe7:
Add "Never Connected Devices" and "Disconnected Devices" stats fields.

Closes #157


Annopaolo (changes requested):
Good starting point, some more work needs to be done:
- Document these changes: edit the `apps/astarte_appengine_api/priv/static/astarte_appengine_api.yaml` to include the new fields
- Add tests for newly added fields
- Add a line to CHANGELOG.md briefly describing what was added, keeping in mind that it is used mostly to communicate changes to users.

PR #857, opened 2023-10-16
Housekeeping: add a PATCH-based API for realm update

Annopaolo:
Introduce a PATCH-based API for realm update. The old POST-based one is deprecated.
At the moment, only public key update is supported, but the new API allows for more flexibility. In the future (from ScyllaDB >= 5.3), it will allow to update the replication settings of a realm.
Based on astarte-platform/astarte_rpc#76, in the general context of  astarte-platform/astarte#772.


apps/astarte_housekeeping/test/astarte_housekeeping/engine_test.exs
```diff
@@ -0,0 +1,100 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2017-2018 Ispirata Srl
```
noaccOS, line 4:
```suggestion
# Copyright 2023 SECO Mind Srl
```

Annopaolo: Ty

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/realms/realm.ex
```diff
@@ -93,15 +114,32 @@ defmodule Astarte.Housekeeping.API.Realms.Realm do
...
+    replication_factor = get_field(changeset, :replication_factor)
+
+    if replication_class == @default_replication_class and replication_factor == nil do
+      put_change(changeset, :replication_factor, @default_replication_factor)
+    else
+      changeset
+    end
+  end
+
+  defp validate_pem_public_key_if_present(changeset, field) do
```
noaccOS, line 133: I don't think this is correct. The pem will always be present because it is required during the creation, and so will always be `!= nil` (unless manually set to `nil` in `params`).

What you probably want is to match `with {:changes, new_pem} <- fetch_field(changeset, field)`, so that it will only validate changes.

BUT there already is a function which only validates if we're editing the value, and that is `Ecto.Changeset.validate_change/3`.

So I suggest instead rewriting `validate_pem_public_key/2` as 

```elixir
defp validate_pem_public_key(:jwt_public_key_pem, pem) do
  try do
    [{:SubjectPublicKeyInfo, _, _}] = :public_key.pem_decode(pem)
    []
  rescue
    _ ->
      [jwt_public_key_pem: "is not a valid PEM public key"]
  end
end
```

and using it with `validate_change/2`. This would also work for `changeset` as it should be used as a creation changeset on an empty struct.

Also, even in `update_changeset` we want `:jwt_public_key_pem` required, because it will only invalidate the changeset if the value isn't present after the change.

```elixir
iex(15)> Ecto.Changeset.change(%HK{realm_name: "test"}, %{}) |> Ecto.Changeset.validate_required(:jwt_public_key_pem)
#Ecto.Changeset<
  action: nil,
  changes: %{},
  errors: [jwt_public_key_pem: {"can't be blank", [validation: :required]}],
  data: #Astarte.Housekeeping<>,
  valid?: false
>
iex(16)> Ecto.Changeset.change(%HK{realm_name: "test"}, %{jwt_public_key_pem: "pem"}) |> Ecto.Changeset.validate_required(:jwt_public_key_pem)
#Ecto.Changeset<
  action: nil,
  changes: %{jwt_public_key_pem: "pem"},
  errors: [],
  data: #Astarte.Housekeeping<>,
  valid?: true
>
iex(17)> Ecto.Changeset.change(%HK{realm_name: "test", jwt_public_key_pem: "pem"}, %{}) |> Ecto.Changeset.validate_required(:jwt_public_key_pem)
#Ecto.Changeset<action: nil, changes: %{}, errors: [],
 data: #Astarte.Housekeeping<>, valid?: true>
iex(18)> Ecto.Changeset.change(%HK{realm_name: "test", jwt_public_key_pem: "pem"}, %{jwt_public_key_pem: nil}) |> Ecto.Changeset.validate_required(:jwt_public_key_pem)
#Ecto.Changeset<
  action: nil,
  changes: %{},
  errors: [jwt_public_key_pem: {"can't be blank", [validation: :required]}],
  data: #Astarte.Housekeeping<>,
  valid?: false
>
```

so we should still keep `validate_required(@required_fields)`

apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
```diff
@@ -83,4 +83,83 @@ defmodule Astarte.Housekeeping.Engine do
...
+             update_attrs.datacenter_replication_factors == %{} do
+      _ =
+        Logger.warn("Trying to update replication values for realm",
+          tag: "invalid_replication_value_update"
+        )
+
+      false
+    else
+      true
+    end
```
noaccOS, line 163:
```suggestion
    realm_valid? = update_attrs.replication_factor == nil && update_attrs.replication_class == nil &&
             update_attrs.datacenter_replication_factors == %{}
             
    unless realm_valid? do
      Logger.warn("Trying to update replication values for realm",
        tag: "invalid_replication_value_update"
        )
    end

    realm_valid?
```

apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
```diff
@@ -83,4 +83,83 @@ defmodule Astarte.Housekeeping.Engine do
...
   end
+
+  @doc """
+  Updates a given realm using the provided values map. Fails if realm does not exists or values are invalid.
+  At the moment, only updating the realm jwt_public_key_pem is supported.
+  Returns a tuple:
+  - either {:ok, updated_realm}
+  - or {:error, reason} where reason is an atom describing the error.
+  """
+  @spec update_realm(binary(), %{:jwt_public_key_pem => binary(), optional(any()) => any()}) ::
```
noaccOS, line 94: `realm_name` can probably be `String.t()` instead of `binary()` as we want it to be a printable string.
Same for `jwt_public_key_pem`.

apps/astarte_housekeeping/lib/astarte_housekeeping/rpc/handler.ex
```diff
@@ -152,6 +153,47 @@ defmodule Astarte.Housekeeping.RPC.Handler do
     end
   end
 
+  defp call_rpc({:update_realm, %UpdateRealm{} = call}) do
+    %UpdateRealm{realm: realm_name} = call
```
noaccOS, line 157: I don't think it's worth it extracting the realm name this way as

1. we're matching a single value. `call.realm` works just as well
2. we're already matching on `%UpdateRealm{}` in the function head, so this doesn't give us new information
3. `realm_name` is only used in the `Engine.update_realm` call. it's probably safe to just call `call.realm` there, or even consider refactoring `update_realm` to extract the realm name itself, depending on the expected future usage of that function.

noaccOS (changes requested):

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/realms/realm.ex
```diff
@@ -122,4 +160,14 @@ defmodule Astarte.Housekeeping.API.Realms.Realm do
         |> add_error(field, "is not a valid PEM public key")
     end
   end
+
+  defp put_default_if_missing(changeset, field, default) do
+    value = get_field(changeset, field)
+
+    if value == nil do
```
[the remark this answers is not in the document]
Annopaolo: Note: to do this, we need to update Ecto to 3.10 and phoenix_ecto to 44.x. I think it's ok, though.

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/realms/realm.ex
Annopaolo: What you want to validate here is not a full realm configuration, but just a merge patch (see https://datatracker.ietf.org/doc/html/rfc7386) to the realm in order to update it. Therefore, the public key (as any other field) might not be present. As an example, this is a valid (albeit useless) patch: `{"data": {}}`.
The backend will then take care of updating the realm, merging the patch and the current status, and for that reason it will return an updated realm.

rbino (approved):

PR #868, opened 2023-11-02
AstarteExport: update to Elixir 1.14.x

osmanhadzic:
Update Elixir version from 1.11 to 1.14.5
Update dependencies astarte_core

Closes: #865


Annopaolo (changes requested):
Looks ok, thank you! Also, remember to update Erlang/Elixir in `Dockerfile`

tools/astarte_export/run_export.sh
Annopaolo, line 1: The env var values are hardcoded here, how could one change them if needed? I would expect this script to read them from the environment, rather than setting them.

tools/astarte_export/config/runtime.exs
Annopaolo, line 1: This config will override `prod.exs` and `dev.exs`. No need to duplicate code

tools/astarte_export/Dockerfile
```diff
@@ -37,7 +42,7 @@ RUN apt-get -qq install libssl1.1
 
 # We have to redefine this here since it goes out of scope for each build stage
 ARG BUILD_ENV=prod
-
+COPY ./run_export.sh /app/
 COPY --from=builder /app/_build/$BUILD_ENV/rel/astarte_export .
 
-CMD ["sh", "-c", "./bin/astarte_export export ${REALM} ${OUTPUT_XML_FILE}"]
+CMD /app/run_export.sh
```
Annopaolo, line 48: The preferred form for `CMD` is the so-called "exec" form, see https://docs.docker.com/engine/reference/builder/#cmd. Let's keep using best practices

tools/astarte_export/Dockerfile
```diff
@@ -1,11 +1,16 @@
-FROM elixir:1.11.4 as builder
+FROM hexpm/elixir:1.14.5-erlang-25.3.2-debian-bullseye-20230522-slim as builder
```
Annopaolo, line 1: Note that you are also updating Debian from stretch to bullseye

osmanhadzic: I do not understand what you are trying to convey here. Would it be possible for you to clarify this for me?

tools/astarte_export/Dockerfile
```diff
@@ -1,17 +1,27 @@
...
 
 # Pass --build-arg BUILD_ENV=dev to build a dev image
 ARG BUILD_ENV=prod
 
 ENV MIX_ENV=$BUILD_ENV
 
+ENV CASSANDRA_DB_HOST="127.0.0.1"
+ENV CASSANDRA_DB_PORT="9042"
+ENV REALM="test"
+ENV XML_FILE="../files/test.xml"
```
Annopaolo, line 23: These values should be settable by the `docker run` command; the values provided here can be sensible default but I would expect a user to be able to run the container by doing something like
```
docker run -e "CASSANDRA_DB_HOST=http://my.host.com" -e "CASSANDRA_DB_PORT=h9042" -e "REALM=myverynicerealm" -e "XML_FILE=a/b/c.xml" astarte/astarte_export
```

Annopaolo (changes requested):
Maybe also factor out the common congif found in `config/dev.exs` and `config/prod.exs` to `config/runtime.exs`, as there is no difference at the moment between the two

tools/astarte_export/Dockerfile
osmanhadzic: I have a problem with CASSANDRA_DB_HOST and CASSANDRA_DB_PORT when providing these variables in run application. This causes a break in function for connecting to the database. As a result of not reading the environment variables.

Annopaolo: You can take a look at the Edgehog backend config (https://github.com/edgehog-device-manager/edgehog/blob/main/backend/config/runtime.exs) and Dockerfile (https://github.com/edgehog-device-manager/edgehog/blob/main/backend/Dockerfile) as an example on handling env vars when using Docker

tools/astarte_export/config/config.exs
```diff
@@ -1,32 +1,7 @@
...
-#
-
-# It is also possible to import configuration files, relative to this
-# directory. For example, you can emulate configuration per environment
-# by uncommenting the line below and defining dev.exs, test.exs and such.
-# Configuration from the imported file will override the ones defined
-# here (which is why it is important to import them last).
-#
+config :cqerl,
+  cassandra_nodes: [{System.get_env("CASSANDRA_DB_HOST"), System.get_env("CASSANDRA_DB_PORT")}]
```
Annopaolo, line 4: Xandra, too, should receive the info on nodes and port (see https://hexdocs.pm/xandra/0.13.1/Xandra.html#start_link/1-options and https://github.com/astarte-platform/astarte_data_access/blob/45d4d20ae662751f47f4b8f2f9d5e302d5485ea9/lib/astarte_data_access/config.ex#L158-L168)

Annopaolo (changes requested):

tools/astarte_export/config/runtime.exs
```diff
@@ -0,0 +1,4 @@
+import Config
+
+config :cqerl,
+  cassandra_nodes: [{System.get_env("CASSANDRA_DB_HOST"), System.get_env("CASSANDRA_DB_PORT")}]
```
Annopaolo, line 4: Xandra, too, should receive the info on nodes and port (see https://hexdocs.pm/xandra/0.13.1/Xandra.html#start_link/1-options and https://github.com/astarte-platform/astarte_data_access/blob/45d4d20ae662751f47f4b8f2f9d5e302d5485ea9/lib/astarte_data_access/config.ex#L158-L168)

Annopaolo (changes requested):

Annopaolo (approved):

PR #871, opened 2023-11-08
AstarteImport: fix datetime and database connect error

osmanhadzic:
- Changing the Database.connec function parameters
- Fix data type reception_timestamp

### DIFF
--- a/tools/astarte_import/lib/astarte/import/populatedb.ex
+++ b/tools/astarte_import/lib/astarte/import/populatedb.ex
@@ -48,8 +48,8 @@ defmodule Astarte.Import.PopulateDB do
     nodes = Application.get_env(:cqerl, :cassandra_nodes)
     {host, port} = Enum.random(nodes)
     Logger.info("Connecting to #{host}:#{port} cassandra database.", realm: realm)
-
-    {:ok, conn} = Database.connect(realm)
+    opts = [cassandra_nodes: nodes, realm: realm]
+    {:ok, conn} = Database.connect(opts)
     {:ok, xandra_conn} = Xandra.start_link(nodes: ["#{host}:#{port}"])
 
     Logger.info("Connected to database.", realm: realm)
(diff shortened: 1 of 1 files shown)


tools/astarte_import/lib/astarte/import/populatedb.ex
```diff
@@ -359,7 +357,7 @@ defmodule Astarte.Import.PopulateDB do
 
       params =
         if explicit_timestamp do
-          reception_timestamp ++ params
+          [reception_timestamp] ++ params
```
Annopaolo, line 360:
```suggestion
          [reception_timestamp | params]
```

tools/astarte_import/lib/astarte/import/populatedb.ex
```diff
@@ -48,8 +48,8 @@ defmodule Astarte.Import.PopulateDB do
     nodes = Application.get_env(:cqerl, :cassandra_nodes)
     {host, port} = Enum.random(nodes)
     Logger.info("Connecting to #{host}:#{port} cassandra database.", realm: realm)
-
-    {:ok, conn} = Database.connect(realm)
+    opts = [cassandra_nodes: nodes, realm: realm]
+    {:ok, conn} = Database.connect(opts)
     {:ok, xandra_conn} = Xandra.start_link(nodes: ["#{host}:#{port}"])
```
Annopaolo, line 53: This might be ok, but I think that on another PR we should remove `cqerl` and just use Xandra in the whole application

Annopaolo (changes requested):

tools/astarte_import/lib/astarte/import/populatedb.ex
osmanhadzic: Ok, I will open new PR for remove call in application. Can we merge this code.

Annopaolo: Yes! But we should entirely remove cqerl, not just take it away from this file

Annopaolo (approved):

PR #883, opened 2023-12-04
BREAKING: Incoming introspection event introspection map

Annopaolo:
Now that astarte_core defines introspection as a map interface_name => %IntrospectionEntry{major: _, minor: _}
rather than a plaintext string in incoming introspection triggers, this should be handled in DUP and TE (see https://github.com/astarte-platform/astarte_core/pull/77).

AMQP trigger events are defined in astarte_core, while HTTP trigger events will return introspection as a JSON map interface_name -> {major, minor} instead of the plain introspection string sent by the device. 

It is possible to revert to the old behaviour by setting the `DATA_UPDATER_PLANT_GENERATE_LEGACY_INCOMING_INTROSPECTION_EVENTS` flag to `true` (default: `false`).
This change is thus BREAKING.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Update Erlang/OTP to 26.1.
 - Update container base image to Debian `Bookworm`.
 - Container user has changed from `root` to `nobody`.
+- BREAKING: incoming_introspection events display introspection as a map rather
+  than a plaintext string. Revert to the old behaviour by setting
+  `DATA_UPDATER_PLANT_GENERATE_LEGACY_INCOMING_INTROSPECTION_EVENTS` to `true`.
+  See https://github.com/astarte-platform/astarte_core/pull/77.
 
 ## [1.1.1] - 2023-11-15
 ### Fixed
--- a/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/triggers_handler_test.exs
+++ b/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/triggers_handler_test.exs
@@ -28,6 +28,7 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandlerTest do
     InterfaceAddedEvent,
     InterfaceMinorUpdatedEvent,
     InterfaceRemovedEvent,
+    InterfaceVersion,
     PathCreatedEvent,
     PathRemovedEvent,
     SimpleEvent,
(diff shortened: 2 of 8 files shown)


Annopaolo:
Based on https://github.com/astarte-platform/astarte_core/pull/77

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Update Erlang/OTP to 26.1.
 - Update container base image to Debian `Bookworm`.
 - Container user has changed from `root` to `nobody`.
+- incoming_introspection events display introspection as a map rather than
```
rbino, line 26: Nit: this needs a `BREAKING` notice since by default it's breaking (you need to take an action for it not being breaking)

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/triggers_handler_test.exs
```diff
@@ -334,6 +339,54 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandlerTest do
...
+      assert %IncomingIntrospectionEvent{introspection: @introspection} =
+               incoming_introspection_event
+
+      headers_map = amqp_headers_to_map(meta.headers)
+
+      assert Map.get(headers_map, "x_astarte_realm") == @realm
+      assert Map.get(headers_map, "x_astarte_device_id") == @device_id
+      assert Map.get(headers_map, "x_astarte_event_type") == "incoming_introspection_event"
+      assert Map.get(headers_map, static_header_key) == static_header_value
+      Config.put_generate_legacy_incoming_introspection_events(false)
```
rbino, line 387: Put this at the beginning in an `on_exit` callback:
```elixir
Config.put_generate_legacy_incoming_introspection_events(true)
on_exit(fn -> Config.put_generate_legacy_incoming_introspection_events(false))
```

Otherwise if the test terminates early it won't restore this, possibly leading to more tests failing (and making it hard to track the actual only cause of the failure)

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -185,6 +190,28 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
...
+      } = version
+
+      Map.put_new(acc, interface_name, %{
+        major: version_major,
+        minor: version_minor
+      })
+    end)
+  end
+
+  defp maybe_normalize_introspection_event(event), do: event
```
rbino, line 213: Why isn't this behind the feature flag env? If we maintain compatibility for AMQP triggers I though we were going to maintain it also for HTTP triggers

rbino (changes requested):

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
Annopaolo: Compatibility is maintained: as the IncomingIntrospectionEvent [protobuf](https://github.com/astarte-platform/astarte_core/blob/5335b0d6e9cd33c302a69c3ac905787515244750/lib/astarte_core/triggers/simple_events/incoming_introspection_event.proto#L27) has `introspection` set to `nil` only when `introspection_map` is not nil (see the [relevant part of this PR on DUP](https://github.com/astarte-platform/astarte/pull/883/files#diff-e4f382e29d9d340ecc3a4c68339fc94273df0e60f41e8e07c80c55e414f06aeaR175-R180)), an event with the old format will not match the first head and fall back safely into the second, being left as it is (just an introspection string)

rbino (approved):

Issue #721, opened 2022-06-30
Data Updater Plant: log base64-encoded payloads, instead of Elixir representation


Annopaolo:
A number of DUP logs display the Elixir binary representation of Astarte payloads, which may seem alien and difficult to understand. Moreover, it may happen that the displayed value is shortened, rendering it useless.
Logging the base64-encoded payload would make it easier to decode and handle it.

PR #884, opened 2023-12-07
Enhance DUP Logs Readability: Base64 Encode Astarte Payloads for Improved Decoding and Handling

osmanhadzic:
Enhance DUP Logs Readability: Base64 Encode Astarte Payloads for Improved Decoding and Handling

Closes: #721

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
@@ -172,7 +172,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
   end
 
   def handle_internal(state, path, payload, message_id, timestamp) do
-    Logger.warn("Unexpected internal message on #{path}, payload: #{inspect(payload)}",
+    Logger.warn(
+      "Unexpected internal message on #{path}, base64-encoded payload: #{inspect(Base.encode64(payload))}",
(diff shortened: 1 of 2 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -172,7 +172,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
   end
 
   def handle_internal(state, path, payload, message_id, timestamp) do
-    Logger.warn("Unexpected internal message on #{path}, payload: #{inspect(payload)}",
+    Logger.warn(
+      "Unexpected internal message on #{path}, payload: #{inspect(Base.encode64(payload))}",
```
Annopaolo, line 176:
```suggestion
      "Unexpected internal message on #{path}, base64-encoded payload: #{inspect(Base.encode64(payload))}",
```

Annopaolo (changes requested):
The PR is good, however it would be better to include in the log line that you're logging the base64 encoding of the payload, I left a comment as an example.

Also, the title of both the PR and the commit should be updated.

Annopaolo (changes requested):
Please, add an indication that displayed payloads are encoded in base64 to the log lines you modified and update PR and commit title (you're not decoding the payload)

Annopaolo (approved):

PR #887, opened 2023-12-12
Describe the device deletion API in the Realm Management's Swagger

davidebriani:
Add the HTTP DELETE action supported by Realm Management in the Swagger document describing the possibility to delete a device from a realm by specifying its device ID.

### DIFF
--- a/apps/astarte_realm_management_api/priv/static/astarte_realm_management_api.yaml
+++ b/apps/astarte_realm_management_api/priv/static/astarte_realm_management_api.yaml
@@ -464,6 +464,34 @@ paths:
           $ref: '#/components/responses/TriggerDeliveryPolicyCurrentlyUsedError'
         '500':
           description: Internal Server Error.
+  '/{realm_name}/devices/{device_id}':
+    parameters:
+      - $ref: '#/components/parameters/Realm'
+      - $ref: '#/components/parameters/DeviceID'
+    delete:
+      tags:
+        - device
+      summary: Delete device
+      description: >-
+        Deletes an existing device with a given `device_id`.
+        Device deletion happens asynchronously, and receiving a 204 response
+        doesn't guarantee immediate deletion of the device.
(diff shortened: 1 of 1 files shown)


apps/astarte_realm_management_api/priv/static/astarte_realm_management_api.yaml
```diff
@@ -464,6 +464,30 @@ paths:
           $ref: '#/components/responses/TriggerDeliveryPolicyCurrentlyUsedError'
         '500':
           description: Internal Server Error.
+  '/{realm_name}/devices/{device_id}':
```
szakhlypa, line 467: We can add parameters
```suggestion
  '/{realm_name}/devices/{device_id}':
    parameters:
      - $ref: '#/components/parameters/Realm'
      - $ref: '#/components/parameters/DeviceID'
```
And into `parameters` below
```
    DeviceID:
      name: device_id
      in: path
      description: Target realm
      required: true
      schema:
        type: string
        - name: device_id
          in: path
          description: Device ID
          required: true
          schema:
            type: string
```

apps/astarte_realm_management_api/priv/static/astarte_realm_management_api.yaml
```diff
@@ -464,6 +464,33 @@ paths:
...
+    parameters:
+      - $ref: '#/components/parameters/Realm'
+      - $ref: '#/components/parameters/DeviceID'
+    delete:
+      tags:
+        - device
+      summary: Delete device
+      description: >-
+        Deletes an existing device with a given `device_id`.
+        Device deletion is performed asynchronously.
```
Annopaolo, line 477: Maybe specify that, since deletion is asynchronous, a 204 response is not a guarantee that the device is deleted straightaway.

szakhlypa (approved):

Annopaolo (approved):

PR #889, opened 2023-12-18
DUP: do not crash when purge properties payload is not zlib deflated

Annopaolo:
Due to a quirk in OTP's zlib library, the `safe_inflate/1` function can actually throw an error if the payload is malformed.
> In all functions errors, {'EXIT',{Reason,Backtrace}}, can be thrown, where Reason describes the error.

The unhandled error made the `DataUpdater` process crash, resulting in the malformed purge properties message being requeued and so on and so forth.
Fix this behaviour by wrapping relevant `:zlib` calls in a try-catch block.

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
@@ -1,7 +1,7 @@
 #
 # This file is part of Astarte.
 #
-# Copyright 2017 Ispirata Srl
+# Copyright 2017 - 2023 SECO Mind Srl
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -1250,11 +1250,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
 
     timestamp_ms = div(timestamp, 10_000)
 
-    operation_result = prune_device_properties(new_state, "", timestamp_ms)
-
-    if operation_result != :ok do
-      Logger.debug("Result is #{inspect(operation_result)} further actions should be required.")
-    end
+    :ok = prune_device_properties(new_state, "", timestamp_ms)
 
     MessageTracker.ack_delivery(new_state.message_tracker, message_id)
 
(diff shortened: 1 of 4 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -1281,20 +1281,52 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
     decoded_payload = PayloadsDecoder.safe_inflate(zlib_payload)
 
     if decoded_payload != :error do
-      operation_result = prune_device_properties(new_state, decoded_payload, timestamp_ms)
+      payload_size = byte_size(payload)
```
rbino, line 1284: Why this has to be calculated outside?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -1281,20 +1281,52 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
...
+  defp handle_valid_purge_properties(
+         state,
+         decoded_payload,
+         payload_size,
+         message_id,
+         timestamp_ms
+       ) do
+    operation_result = prune_device_properties(state, decoded_payload, timestamp_ms)
+
+    # TODO at the moment operation_result is always :ok, we might want to revise this
```
rbino, line 1318: If right now this can't fail, let's be assertive and do:
```
:ok = prune_device_properties(state, decoded_payload, timestamp_ms)
```

Otherwise we could miss if we introduce a possible error value later (a debug log will go unnoticed for a long time in production and leave a false sense of security).

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/payloads_decoder.ex
```diff
@@ -65,8 +66,22 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.PayloadsDecoder do
   Safely decodes a zlib deflated binary and inflates it.
   This function avoids zip bomb vulnerabilities, and it decodes up to 10_485_760 bytes.
   """
-  @spec safe_inflate(binary) :: binary
+  @spec safe_inflate(binary) :: binary | :error
   def safe_inflate(zlib_payload) do
+    try do
```
rbino, line 71: Elixir supports `catch` at the `def` level (see [here](https://hexdocs.pm/elixir/1.16/Kernel.html#def/2-rescue-catch-after-else)), so you can skip the `do_` helper and just do

```elixir
def safe_inflate(zlib_payload) do
  ...
catch
  :error, error ->
    ...
end

File: apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs [hunk 4]
```diff
@@ -192,6 +192,11 @@ defmodule Astarte.DataUpdaterPlant.PayloadsDecoderTest do
     assert PayloadsDecoder.safe_inflate(compressed) == :error
   end
 
+  test "zlib inflate does not crash with a payload that is not zlib deflated" do
+    random_bytes = :crypto.strong_rand_bytes(System.unique_integer([:positive]))
```
rbino, line 196: There's a non-zero chance that this produces a valid a valid zlib payload, and also `System.unique_integer` might get very big if this test is executed last in the test-suite. Replace this with a fixed non-valid payload.

rbino (changes requested):

rbino (approved):

Issue #890, opened 2024-01-08
Astarte microservices should expose a version endpoint


Annopaolo:
Add a `/version` endpoint to each microservice API, so that it could be queried by third-party applications that depend on Astarte. Note that the endpoint would return the microservice version, not the API version.
The endpoint should be queried using the `GET` verb and it should be accessible without auth, as done with `/health`.

This would allow applications to act differently depending on the underlying Astarte services version. This might come in hand e.g. to know whether device deletion is allowed or not.

Annopaolo:
Closed by #893

PR #893, opened 2024-01-22
Astarte: update /version endpoint

osmanhadzic:
Provide third-party applications with information about the version of Astarte microservices they are interacting with, enabling them to adapt their behavior based on the underlying service version.

Standardization of routers for the Astarte API endpoints to have a uniform Phoenix pipeline structure.

Add a ```/version``` endpoint to each microservice API, allowing third-party applications to query and retrieve the microservice version using the GET verb. Authentication is not required for accessing this endpoint. 

Standardize code for Astarte APIs by refactoring the router for endpoints. Remove plugs from controllers and incorporate them into the router pipeline.

By improving code readability in the app, astarte will expose information about the microservice version to third-party applications without authorization.

(#890)

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
@@ -130,6 +130,10 @@ defmodule Astarte.AppEngine.APIWeb.Router do
       disable_validator: true
   end
 
+  scope "/version", Astarte.AppEngine.APIWeb do
+    get "/", VersionController, :show
+  end
+
   defp maybe_halt_swagger(conn, _opts) do
     if Application.get_env(:astarte_appengine_api, :swagger_ui, false) do
       conn
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
@@ -19,10 +19,28 @@ defmodule Astarte.AppEngine.APIWeb.Router do
   use Astarte.AppEngine.APIWeb, :router
   alias Astarte.AppEngine.APIWeb.Plug.LogRealm
 
-  pipeline :api do
+  pipeline :realm_api do
     plug :accepts, ["json"]
     plug LogRealm
     plug Astarte.AppEngine.APIWeb.Plug.AuthorizePath
+  end
+
+  pipeline :device_api do
+    plug Astarte.AppEngine.APIWeb.Plug.LogDeviceId
+  end
+
+  pipeline :device_group_api do
+    plug Astarte.AppEngine.APIWeb.Plug.LogGroupName
+  end
+
+  pipeline :device_alias_api do
+    plug Astarte.AppEngine.APIWeb.Plug.LogDeviceAlias
+  end
+
+  pipeline :interface_value_api do
+    plug Astarte.AppEngine.APIWeb.Plug.JoinPath
+    plug Astarte.AppEngine.APIWeb.Plug.LogInterface
+    plug Astarte.AppEngine.APIWeb.Plug.LogPath
   end
 
   pipeline :swagger do
@@ -30,7 +48,7 @@ defmodule Astarte.AppEngine.APIWeb.Router do
   end
 
   scope "/v1/:realm_name", Astarte.AppEngine.APIWeb do
-    pipe_through :api
+    pipe_through :realm_api
 
     get "/version", VersionController, :show
 
@@ -38,87 +56,117 @@ defmodule Astarte.AppEngine.APIWeb.Router do
       get "/devices", StatsController, :show_devices_stats
     end
 
-    resources "/devices", DeviceStatusController,
-      only: [:index, :show, :update],
-      param: "device_id"
+    scope "/devices" do
+      pipe_through :device_api
 
-    resources "/devices-by-alias", DeviceStatusByAliasController,
-      only: [:index, :show, :update],
-      param: "device_alias"
+      resources "/", DeviceStatusController,
+        only: [:index, :show, :update],
+        param: "device_id"
 
-    resources "/devices/:device_id/interfaces", InterfaceValuesController,
-      only: [:index, :show],
-      param: "interface"
+      scope "/:device_id/interfaces" do
+        pipe_through :interface_value_api
 
-    get "/devices/:device_id/interfaces/:interface/*path_tokens", InterfaceValuesController, :show
+        resources "/", InterfaceValuesController,
+          only: [:index, :show],
+          param: "interface"
 
-    put "/devices/:device_id/interfaces/:interface/*path_tokens",
-        InterfaceValuesController,
-        :update
+        get "/:interface/*path_tokens", InterfaceValuesController, :show
 
-    post "/devices/:device_id/interfaces/:interface/*path_tokens",
-         InterfaceValuesController,
-         :update
+        put "/:interface/*path_tokens",
+            InterfaceValuesController,
+            :update
 
-    delete "/devices/:device_id/interfaces/:interface/*path_tokens",
-           InterfaceValuesController,
-           :delete
+        post "/:interface/*path_tokens",
+             InterfaceValuesController,
+             :update
 
-    resources "/devices-by-alias/:device_alias/interfaces",
-              InterfaceValuesByDeviceAliasController,
-              only: [:index, :show],
-              param: "interface"
+        delete "/:interface/*path_tokens",
+               InterfaceValuesController,
+               :delete
+      end
+    end
 
-    get "/devices-by-alias/:device_alias/interfaces/:interface/*path_tokens",
-        InterfaceValuesByDeviceAliasController,
-        :show
+    scope "/devices-by-alias" do
+      pipe_through :device_alias_api
 
-    put "/devices-by-alias/:device_alias/interfaces/:interface/*path_tokens",
-        InterfaceValuesByDeviceAliasController,
-        :update
+      resources "/", DeviceStatusByAliasController,
+        only: [:index, :show, :update],
+        param: "device_alias"
 
-    post "/devices-by-alias/:device_alias/interfaces/:interface/*path_tokens",
-         InterfaceValuesByDeviceAliasController,
-         :update
+      scope "/:device_alias/interfaces" do
+        pipe_through :interface_value_api
 
-    delete "/devices-by-alias/:device_alias/interfaces/:interface/*path_tokens",
-           InterfaceValuesByDeviceAliasController,
-           :delete
+        resources "/",
+                  InterfaceValuesByDeviceAliasController,
+                  only: [:index, :show],
+                  param: "interface"
 
-    get "/groups", GroupsController, :index
-    post "/groups", GroupsController, :create
-    get "/groups/:group_name", GroupsController, :show
-    post "/groups/:group_name/devices", GroupsController, :add_device
-    delete "/groups/:group_name/devices/:device_id", GroupsController, :remove_device
+        get "/:interface/*path_tokens",
+            InterfaceValuesByDeviceAliasController,
+            :show
 
-    get "/groups/:group_name/devices", DeviceStatusByGroupController, :index
-    get "/groups/:group_name/devices/:device_id", DeviceStatusByGroupController, :show
+        put "/:interface/*path_tokens",
+            InterfaceValuesByDeviceAliasController,
+            :update
 
-    patch "/groups/:group_name/devices/:device_id", DeviceStatusByGroupController, :update
+        post "/:interface/*path_tokens",
+             InterfaceValuesByDeviceAliasController,
+             :update
 
-    get "/groups/:group_name/devices/:device_id/interfaces",
-        InterfaceValuesByGroupController,
-        :index
+        delete "/:interface/*path_tokens",
+               InterfaceValuesByDeviceAliasController,
+               :delete
+      end
+    end
 
-    get "/groups/:group_name/devices/:device_id/interfaces/:interface",
-        InterfaceValuesByGroupController,
-        :show
+    scope "/groups" do
+      pipe_through :device_group_api
 
-    get "/groups/:group_name/devices/:device_id/interfaces/:interface/*path_tokens",
-        InterfaceValuesByGroupController,
-        :show
+      get "/", GroupsController, :index
+      post "/", GroupsController, :create
 
-    put "/groups/:group_name/devices/:device_id/interfaces/:interface/*path_tokens",
-        InterfaceValuesByGroupController,
-        :update
+      scope "/:group_name" do
+        get "/", GroupsController, :show
+        get "/devices", DeviceStatusByGroupController, :index
+        post "/devices", GroupsController, :add_device
 
-    post "/groups/:group_name/devices/:device_id/interfaces/:interface/*path_tokens",
-         InterfaceValuesByGroupController,
-         :update
+        scope "/devices/:device_id" do
+          pipe_through :device_api
 
-    delete "/groups/:group_name/devices/:device_id/interfaces/:interface/*path_tokens",
-           InterfaceValuesByGroupController,
-           :delete
+          get "/", DeviceStatusByGroupController, :show
+          patch "/", DeviceStatusByGroupController, :update
+          delete "/", GroupsController, :remove_device
+
+          scope "/interfaces" do
+            pipe_through :interface_value_api
+
+            get "/",
+                InterfaceValuesByGroupController,
+                :index
+
+            get "/:interface",
+                InterfaceValuesByGroupController,
+                :show
+
+            get "/:interface/*path_tokens",
+                InterfaceValuesByGroupController,
+                :show
+
+            put "/:interface/*path_tokens",
+                InterfaceValuesByGroupController,
+                :update
+
+            post "/:interface/*path_tokens",
+                 InterfaceValuesByGroupController,
+                 :update
+
+            delete "/:interface/*path_tokens",
+                   InterfaceValuesByGroupController,
+                   :delete
+          end
+        end
+      end
+    end
   end
 
   scope "/swagger" do
--- a/apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
+++ b/apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
@@ -40,4 +40,8 @@ defmodule Astarte.Pairing.APIWeb.Router do
          DeviceController,
          :verify_credentials
   end
+
+  scope "/version", Astarte.Pairing.APIWeb do
+    get "/", VersionController, :show
+  end
 end
--- a/apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
+++ b/apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
@@ -19,26 +19,41 @@
 defmodule Astarte.Pairing.APIWeb.Router do
   use Astarte.Pairing.APIWeb, :router
 
-  pipeline :api do
+  pipeline :realm_api do
     plug :accepts, ["json"]
     plug Astarte.Pairing.APIWeb.Plug.LogRealm
   end
 
+  pipeline :agent_api do
+    plug Astarte.Pairing.APIWeb.Plug.AuthorizePath
+  end
+
+  pipeline :devices_api do
+    plug Astarte.Pairing.APIWeb.Plug.LogHwId
+  end
+
   scope "/v1/:realm_name", Astarte.Pairing.APIWeb do
-    pipe_through :api
+    pipe_through :realm_api
 
     get "/version", VersionController, :show
 
-    post "/agent/devices", AgentController, :create
-    delete "/agent/devices/:device_id", AgentController, :delete
+    scope "/agent" do
+      pipe_through :agent_api
 
-    get "/devices/:hw_id", DeviceController, :show_info
+      post "/devices", AgentController, :create
+      delete "/devices/:device_id", AgentController, :delete
+    end
 
-    post "/devices/:hw_id/protocols/:protocol/credentials", DeviceController, :create_credentials
+    scope "/devices" do
+      pipe_through :devices_api
 
-    post "/devices/:hw_id/protocols/:protocol/credentials/verify",
-         DeviceController,
-         :verify_credentials
+      get "/:hw_id", DeviceController, :show_info
+      post "/:hw_id/protocols/:protocol/credentials", DeviceController, :create_credentials
+
+      post "/:hw_id/protocols/:protocol/credentials/verify",
+           DeviceController,
+           :verify_credentials
+    end
   end
 
   scope "/version", Astarte.Pairing.APIWeb do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [astarte_housekeeping_api] Allow to read and set a realm's device registration
   limit using the realm fetch and update API, respectively.
 - [astarte_appengine_api] Show deletion status in device details.
+- [astarte_appengine_api] Add /version endpoint which returns the microservice
+  application version.
+- [astarte_realm_management_api] Add /version endpoint which returns
+  the microservice application version.
+- [astarte_pairing_api] Add /version endpoint which returns the microservice
+  application version.
+- [astarte_housekeeping_api] Add /version endpoint which returns
+  the microservice application version.
 - [astarte_housekeeping_api] Allow to customize the RPC call timeout with
   `HOUSEKEEPING_API_RPC_TIMEOUT` (default: 5 seconds).
 - Add API usage metrics.
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api_web/controllers/device_status_by_alias_controller.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api_web/controllers/device_status_by_alias_controller.ex
@@ -23,8 +23,6 @@ defmodule Astarte.AppEngine.APIWeb.DeviceStatusByAliasController do
   alias Astarte.AppEngine.API.Device.DeviceStatus
   alias Astarte.AppEngine.APIWeb.DeviceStatusView
 
-  plug Astarte.AppEngine.APIWeb.Plug.LogDeviceAlias
-
   action_fallback Astarte.AppEngine.APIWeb.FallbackController
 
   # TODO: should we allow to POST/create device aliases here by posting something like a DeviceAlias JSON object?
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api_web/controllers/device_status_by_group_controller.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api_web/controllers/device_status_by_group_controller.ex
@@ -24,9 +24,6 @@ defmodule Astarte.AppEngine.APIWeb.DeviceStatusByGroupController do
   alias Astarte.AppEngine.API.Groups
   alias Astarte.AppEngine.APIWeb.DeviceStatusView
 
-  plug Astarte.AppEngine.APIWeb.Plug.LogGroupName
-  plug Astarte.AppEngine.APIWeb.Plug.LogDeviceId
-
   action_fallback Astarte.AppEngine.APIWeb.FallbackController
 
   def index(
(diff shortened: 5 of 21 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api_web/plug/microservice_version_plug.ex
```diff
@@ -0,0 +1,41 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.AppEngine.APIWeb.MicroserviceVersionPlug do
+  @behaviour Plug
+  import Plug.Conn
+  alias Astarte.AppEngine.APIWeb.VersionView
```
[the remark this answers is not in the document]
osmanhadzic: No, I forgot delete alias on VerisonView module.

apps/astarte_appengine_api/lib/astarte_appengine_api_web/endpoint.ex
```diff
@@ -42,6 +42,7 @@ defmodule Astarte.AppEngine.APIWeb.Endpoint do
   plug Astarte.AppEngine.APIWeb.MetricsPlug
   plug PlugLoggerWithMeta
   plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint]
+  plug Astarte.AppEngine.APIWeb.MicroserviceVersionPlug
```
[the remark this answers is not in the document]
osmanhadzic: I am sorry but i don't see any typos?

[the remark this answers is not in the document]
Annopaolo: My bad!

Annopaolo (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -32,8 +32,6 @@ defmodule Astarte.AppEngine.APIWeb.Router do
   scope "/v1/:realm_name", Astarte.AppEngine.APIWeb do
     pipe_through :api
 
-    get "/version", VersionController, :show
-
```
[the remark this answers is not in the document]
osmanhadzic: Do I need to keep or remove /version in other APIs?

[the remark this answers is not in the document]
Annopaolo: What is already present in the current API should not be removed: do not delete `/version` in the `/v1/:realm_name` scope in the API of other services

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
```diff
@@ -22,6 +22,12 @@ defmodule Astarte.Pairing.APIWeb.Router do
   pipeline :api do
     plug :accepts, ["json"]
     plug Astarte.Pairing.APIWeb.Plug.LogRealm
+    plug Astarte.Pairing.APIWeb.Plug.AuthorizePath
+  end
+
+  pipeline :api_device do
+    plug :accepts, ["json"]
+    plug Astarte.Pairing.APIWeb.Plug.LogHwId
```
Annopaolo, line 30: The `Astarte.Pairing.APIWeb.Plug.LogRealm` plug is missing from this new pipeline, add it back

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -23,6 +23,12 @@ defmodule Astarte.AppEngine.APIWeb.Router do
     plug :accepts, ["json"]
     plug LogRealm
     plug Astarte.AppEngine.APIWeb.Plug.AuthorizePath
+    plug Astarte.AppEngine.APIWeb.Plug.LogDeviceAlias
+    plug Astarte.AppEngine.APIWeb.Plug.LogGroupName
+    plug Astarte.AppEngine.APIWeb.Plug.LogDeviceId
+    plug Astarte.AppEngine.APIWeb.Plug.JoinPath
+    plug Astarte.AppEngine.APIWeb.Plug.LogInterface
+    plug Astarte.AppEngine.APIWeb.Plug.LogPath
```
Annopaolo, line 31: Not all controllers using this pipeline need all the plugs. It would be better to factor out only common plugs. For example, the `InterfaceValuesController` had no `LogGroupName` plug, as there is no group involved in retrieving data.

Annopaolo (changes requested):
Some nitpicking on commits:
- Pay attention to typos in the title: `/verison` --> `/version`
- It would be better to keep consistency in commit titles.
If you prefer, you can have a different pipeline standardization commit for each app, but then it would be better to mention the app in the title rather than keeping it generic (as you already did for AppEngine and RealmManagement API). Otherwise, if you want, you can put all pipeline standardization commits together: in this case, it is ok to say that the commit scope includes all apps.

Annopaolo (changes requested):
More nitpicking:
- Commits `706f43a` and `28ad443` should refer to pairing API rather than be generic, as all others
- Commit `ff0e150` has a typo in the title
- The PR title is now outdated, as it is not just AppEngine that is changed

Finally, a new line should be added to `CHANGELOG.md` stating that the new `/version` endpoint was added. Refer to [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) for the format

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -19,32 +19,70 @@ defmodule Astarte.AppEngine.APIWeb.Router do
...
+  pipeline :interface_api do
+    plug Astarte.AppEngine.APIWeb.Plug.JoinPath
+    plug Astarte.AppEngine.APIWeb.Plug.LogInterface
+    plug Astarte.AppEngine.APIWeb.Plug.LogPath
+  end
+
+  pipeline :interface_values_api do
+    plug :device_api
+    plug :interface_api
+  end
```
davidebriani, line 52: I think the rationalization of the APIs looks promising but I will leave a few considerations.
I am not convinced about:
- defining pipelines of pipelines of plugs
- having several `scope`s at the same level just to use different pipelines/plugs on them

Indeed, the router now defines different scopes, all at the same level.
Each scope uses a pipeline of plugs.
Each pipeline of plugs can consist of other pipeline of plugs, which in turn can consist of other pipelines of plugs, and so on.

My doubt is that it might be clearer to instead have nested `scope`s. This way:
1. you don't have to redefine the route of each scope from scratch but rely on the fact that the structure of the routes is already effectively nested
2. in the outer scope you can insert the more generic plugs about the realm, but you don't have to insert them again for each nested scope.
3. by nesting scopes instead of pipelines you also have a visual feedback on the structure of the routes. At the same time, it's easier to spot which plugs are called for each route without having to follow through the definition of pipelines of pipelines of pipelines of plugs.

osmanhadzic: Every controller in Astarte has a specific scope of plugs. Some controllers have similar plugs but with one or two differences (e.g., controllers for interfaces). I try to avoid duplicate code by organizing plugs into groups and using pipelines, such as base pipelines, for other pipelines. (The base pipeline is auth_api).

Annopaolo: I think @davidebriani is suggesting that it would be clearer to have nested scopes, possibly adding plugs at each nesting level, rather than using pipelines as plugs inside other pipelines. I do not have a strong opinion on this, however the nested scopes approach could be more clear as it allows to segregate the different sections of the API without adding new names to keep in mind (i.e. pipeline names)

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -19,106 +19,173 @@ defmodule Astarte.AppEngine.APIWeb.Router do
...
     get "/version", VersionController, :show
 
     scope "/stats" do
       get "/devices", StatsController, :show_devices_stats
     end
 
-    resources "/devices", DeviceStatusController,
-      only: [:index, :show, :update],
-      param: "device_id"
+    scope "/devices/:device_id" do
```
Annopaolo, line 80: This could be nested inside the `"/devices"` scope

Annopaolo (commented):
@davidebriani's point was to use nested scopes, instead of nested pipelines. At each nesting level in scopes, it would be possible to add different plugs.
This could be clearer as it would allow to segregate the different sections of the API without adding new names to keep in mind (i.e. pipeline names).

I left a comment to show an example of a scope that could be nested (in appengine)

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -19,106 +19,172 @@ defmodule Astarte.AppEngine.APIWeb.Router do
...
+    end
+
+    scope "/groups" do
+      pipe_through :device_group_name_api
+
+      get "/:group_name/devices", DeviceStatusByGroupController, :index
+      get "/:group_name/devices/:device_id", DeviceStatusByGroupController, :show
+
+      patch "/:group_name/devices/:device_id", DeviceStatusByGroupController, :update
+    end
```
davidebriani, line 187: It feels strange to have the same scope declared multiple times.
I guess that you were trying to separate routes depending on the `pipe_through` pipeline.

However I would first start declaring the scopes, and nested scopes. Then I would add the appropriate plugs to each level of the nested scopes.
It is not necessary to use pipelines, we can also use plugs directly.

Also note that:
- both the pipelines `device_group_api` and `device_group_name_api` logs the group name. But we don't have a group name to log for GET /groups and POST /groups.
- only the pipeline `device_group_name_api` logs the device ID. But we also have a route with a device ID in the other scope, where the device ID is not logged.

I would suggest first declaring the structure of the routes thanks to nested `scope` blocks, and then adding the appropriate plugs where needed.
For example:

```elixir
scope "/groups" do
  get "/", GroupsController, :index
  post "/", GroupsController, :create

  scope "/:group_name" do
    plug Astarte.AppEngine.APIWeb.Plug.LogGroupName

    get "/", GroupsController, :show
    get "/devices", DeviceStatusByGroupController, :index
    post "/devices", GroupsController, :add_device

    scope "/devices/:device_id" do
      plug Astarte.AppEngine.APIWeb.Plug.LogDeviceId
      
      get "/", DeviceStatusByGroupController, :show
      patch "/", DeviceStatusByGroupController, :update
      delete "/", GroupsController, :remove_device
    end
  end
end
```

davidebriani (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
osmanhadzic: Thank you for the review and suggestions. But, I have a problem with implementing this suggestion. When run the command ` mix phx.router `  returns an error plug `(RuntimeError) cannot define plug at the router level, plug must be defined inside a pipeline`

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -19,106 +19,174 @@ defmodule Astarte.AppEngine.APIWeb.Router do
...
+               InterfaceValuesController,
+               :delete
+      end
+    end
 
-    get "/devices-by-alias/:device_alias/interfaces/:interface/*path_tokens",
-        InterfaceValuesByDeviceAliasController,
-        :show
+    scope "/groups/:group_name" do
+      pipe_through :interface_values_group_api
```
davidebriani, line 110: Be aware that you are still declaring the same scopes multiple times, such as `/groups/:group_name`.
It's probably better to define each scope only a single time and group all corresponding routes under that single scope.
We can do an additional check on the scopes to see if they are correctly organized with a nested structure, without repeating the same scope multiple times.

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
```diff
@@ -31,6 +38,10 @@ defmodule Astarte.Pairing.APIWeb.Router do
 
     post "/agent/devices", AgentController, :create
     delete "/agent/devices/:device_id", AgentController, :delete
+  end
+
+  scope "/v1/:realm_name", Astarte.Pairing.APIWeb do
+    pipe_through :api_device
```
davidebriani, line 44: For example, here too you are repeating the `/v1/:realm_name`.

However note that you did so to scope the routes that present a device ID, so that you can add `pipe_through :api_device`.
If you want to do that, then I would define a scope that matches the path until the device ID, not just the realm name.

For example, using nested scopes where each scope match the appropriate path needed to apply pipelines:

```elixir
scope "/v1/:realm_name", Astarte.Pairing.APIWeb do
  pipe_through :api
  post "/agent/devices", AgentController, :create

  scope "/devices/:hw_id", Astarte.Pairing.APIWeb do
    pipe_through :api_device
    get "/", DeviceController, :show_info
end
```

davidebriani (changes requested):

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
osmanhadzic: [DeviceController](https://github.com/astarte-platform/astarte/blob/595adbd9bb042809215ae3d94c134bcf5a8ce1a1/apps/astarte_pairing_api/lib/astarte_pairing_api_web/controllers/device_controller.ex#L34C27-L34C34)  does not require plug for authorization because created in separate scope.

Annopaolo: As the [Phoenix docs](https://hexdocs.pm/phoenix/Phoenix.Router.html#pipeline/2) say that
> Every time [pipe_through/1](https://hexdocs.pm/phoenix/Phoenix.Router.html#pipe_through/1) is called, the new pipelines are appended to the ones previously given.

A possible solution for handling nested scopes with slightly different pipelines in this case could be
```elixir
  pipeline :api do
    plug :accepts, ["json"]
    plug Astarte.Pairing.APIWeb.Plug.LogRealm
  end

  pipeline :agent_api do
    plug Astarte.Pairing.APIWeb.Plug.AuthorizePath
  end

  pipeline :device_api do
    plug Astarte.Pairing.APIWeb.Plug.LogHwId
  end

  scope "/v1/:realm_name", Astarte.Pairing.APIWeb do
    pipe_through :api

    get "/version", VersionController, :show

    scope "/agent" do
      pipe_through :agent_api

      post "/devices", AgentController, :create
      delete "/devices/:device_id", AgentController, :delete
    end

    scope "/devices" do
      pipe_through :device_api

      get "/:hw_id", DeviceController, :show_info
      post "/:hw_id/protocols/:protocol/credentials", DeviceController, :create_credentials
      post "/:hw_id/protocols/:protocol/credentials/verify",
           DeviceController,
           :verify_credentials
    end
  end
```

What do you think @osmanhadzic @davidebriani ?

davidebriani (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -19,106 +19,172 @@ defmodule Astarte.AppEngine.APIWeb.Router do
...
-              InterfaceValuesByDeviceAliasController,
-              only: [:index, :show],
-              param: "interface"
+        delete "/interfaces/:interface/*path_tokens",
+               InterfaceValuesController,
+               :delete
+      end
+    end
+
+    scope "/devices-by-alias/:device_alias" do
```
davidebriani, line 107: We can scope `/devices-by-alias/:device_alias` within the `/devices-by-alias` scope:

```elixir
pipeline :devices_by_alias_api do
  plug Astarte.AppEngine.APIWeb.Plug.JoinPath
  plug Astarte.AppEngine.APIWeb.Plug.LogInterface
  plug Astarte.AppEngine.APIWeb.Plug.LogPath
end

pipeline :device_alias_api do
  plug Astarte.AppEngine.APIWeb.Plug.LogDeviceAlias
end

scope "/devices-by-alias" do
  pipe_through :devices_by_alias_api
  
  ...
  
  scope "/:device_alias" do
    pipe_through :device_alias_api

    ...
  end
end
```

davidebriani (changes requested):

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
Annopaolo: It seems the suggestion was followed in AppEngine but not in Pairing, is there something that does not convince you @osmanhadzic?

Annopaolo (changes requested):

Annopaolo (changes requested):
Also, an AppEngine API route seems to be missing (see CI run):
```
(Phoenix.Router.NoRouteError) no route found for GET /v1/autotestrealm/devices-by-alias/device_a/interfaces/com.test.LCDMonitor/time/to (Astarte.AppEngine.APIWeb.Router)
```

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
```diff
@@ -25,13 +25,22 @@ defmodule Astarte.Pairing.APIWeb.Router do
...
+    plug Astarte.Pairing.APIWeb.Plug.LogHwId
+  end
+
   scope "/v1/:realm_name", Astarte.Pairing.APIWeb do
     pipe_through :api
 
     get "/version", VersionController, :show
 
     post "/agent/devices", AgentController, :create
     delete "/agent/devices/:device_id", AgentController, :delete
```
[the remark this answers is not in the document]
osmanhadzic: I created :api_device and :api pipeline to separate. DeviceController does not require using `Astarte.Pairing.APIWeb.Plug.AuthorizePath` . I create a new pipeline only for DeviceController to separate from other controllers. @Annopaolo
:api_device does not have anything with pipeline :api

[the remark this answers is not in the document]
Annopaolo: Both `"/agent/devices"` and `"/agent/devices/:device_id"` should have the `Astarte.Pairing.APIWeb.Plug.AuthorizePath` plug, that is currently missing (it has been removed by this PR, see https://github.com/astarte-platform/astarte/pull/893/files/cc196eee9bb629d3ef0863faba74d6718c8d1f41#diff-7bdaad5dbdd7c2ef3cc43ee2eaa46cc23afd1f6e6ae3ef955e274188665cc0c2L27-L28). Unless it is added back, those two endpoints are broken

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
```diff
@@ -20,17 +20,17 @@ defmodule Astarte.Pairing.APIWeb.Router do
   use Astarte.Pairing.APIWeb, :router
 
   pipeline :api do
-    plug :accepts, ["json"]
```
Annopaolo, line 23: The `:accepts, ["json"]` plug should be included in this pipeline too, as e.g. the `"/agent"` scope needs it

Annopaolo (changes requested):
Almost there!

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
osmanhadzic: The `accepts, [json]` is already added in `api_device` pipeline [accepts](https://github.com/osmanhadzic/astarte/blob/736eb8bdf77b8f92bbd183d9e49d5507c5408457/apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex#L28)

Annopaolo (changes requested):
Apart from two minor details, LGTM

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
Annopaolo: Ops, didn't notice, thank you

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
```diff
@@ -39,5 +39,22 @@ defmodule Astarte.Pairing.APIWeb.Router do
...
+
+      get "/", VersionController, :show
+    end
+
+    scope "/agent" do
+      pipe_through :api
+
+      post "/devices", AgentController, :create
+      delete "/devices/:device_id", AgentController, :delete
+    end
```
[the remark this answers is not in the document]
osmanhadzic: I made changes.

Annopaolo (approved):

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -19,106 +19,171 @@ defmodule Astarte.AppEngine.APIWeb.Router do
...
+  pipeline :interface_values_device_group_api do
+    plug Astarte.AppEngine.APIWeb.Plug.JoinPath
+    plug Astarte.AppEngine.APIWeb.Plug.LogInterface
+    plug Astarte.AppEngine.APIWeb.Plug.LogPath
+  end
+
+  pipeline :device_group_name_api do
+    plug Astarte.AppEngine.APIWeb.Plug.LogDeviceId
+    plug Astarte.AppEngine.APIWeb.Plug.LogGroupName
+  end
```
davidebriani, line 61: Since the plugs of the `device_group_name_api` pipelines are only applied by the pipelines `device_api` and `device_group_api`, we could use those two instead (structuring the API routes with appropriate scopes).
So then we can remove the definition of `device_group_name_api`.

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -19,106 +19,171 @@ defmodule Astarte.AppEngine.APIWeb.Router do
   use Astarte.AppEngine.APIWeb, :router
   alias Astarte.AppEngine.APIWeb.Plug.LogRealm
 
-  pipeline :api do
+  pipeline :auth_api do
```
davidebriani, line 22: Since we are naming the pipelines according to their scopes, and this pipeline is scoped to a specific realm, let's call the pipeline `:realm_api`
```suggestion
  pipeline :realm_api do
```

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -19,106 +19,171 @@ defmodule Astarte.AppEngine.APIWeb.Router do
...
+
+      resources "/", DeviceStatusController,
+        only: [:index, :show, :update],
+        param: "device_id"
 
-    resources "/devices-by-alias", DeviceStatusByAliasController,
-      only: [:index, :show, :update],
-      param: "device_alias"
+      scope "/:device_id" do
+        pipe_through :interface_value_api
```
davidebriani, line 84: This scope could be used for more than just interfaces, so it's odd to enforce a pipeline on interface values.

We could use a scope directly on interfaces instead:
```suggestion
      scope "/:device_id/interfaces" do
        pipe_through :interface_values_api
        
        resources "/", InterfaceValuesController,
          only: [:index, :show],
          param: "interface"
        
        ...
```

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -19,106 +19,171 @@ defmodule Astarte.AppEngine.APIWeb.Router do
...
-              param: "interface"
+      resources "/", DeviceStatusByAliasController,
+        only: [:index, :show, :update],
+        param: "device_alias"
 
-    get "/devices-by-alias/:device_alias/interfaces/:interface/*path_tokens",
-        InterfaceValuesByDeviceAliasController,
-        :show
+      scope "/:device_alias" do
+        pipe_through :interface_values_device_group_api
```
davidebriani, line 114: This scope could be used for more than just interfaces, so it's odd to enforce a pipeline on interface values.

We could use a scope directly on interfaces instead:
```suggestion
      scope "/:device_alias/interfaces" do
        pipe_through :interface_values_api
        
        resources "/", InterfaceValuesByDeviceAliasController,,
          only: [:index, :show],
          param: "interface"
        
        ...
```

apps/astarte_appengine_api/lib/astarte_appengine_api_web/router.ex
```diff
@@ -19,106 +19,171 @@ defmodule Astarte.AppEngine.APIWeb.Router do
...
+    plug Astarte.AppEngine.APIWeb.Plug.LogDeviceId
+  end
+
+  pipeline :device_group_api do
+    plug Astarte.AppEngine.APIWeb.Plug.LogGroupName
+  end
+
+  pipeline :device_alias_api do
+    plug Astarte.AppEngine.APIWeb.Plug.LogDeviceAlias
+  end
```
davidebriani, line 38: Ideally:
- the plug LogDeviceId should only be called for routes that have a `device_id` path param.
- the plug LogGroupName should only be called for routes that have a `group_name` path param.
- the plug LogDeviceAlias should only be called for routes that have a `device_alias` path param.

We could achieve that by using appropiate scopes for the routes, for example:

````elixir
scope "/v1/:realm_name", Astarte.AppEngine.APIWeb do
  pipe_through :realm_api
  
  get "/devices" ...
  
  scope "/devices/:device_id" do
    pipe_through :device_api
    
    get "" ...
    patch "" ...
    
    scope "/interfaces" do
      pipe_through :interface_values_api
      
      get "/" ...
      get "/:interface" ...
      get "/:interface/:path" ...
      put "/:interface/:path" ...
      post "/:interface/:path" ...
      delete "/:interface/:path" ...
    end
  end
  
  scope "/devices-by-alias/:device_alias" do
    pipe_through :device_alias_api
    
    get "" ...
    patch "" ...
    
     scope "/interfaces" do
      pipe_through :interface_values_api
      
      get "/" ...
      get "/:interface" ...
      get "/:interface/:path" ...
      put "/:interface/:path" ...
      post "/:interface/:path" ...
      delete "/:interface/:path" ...
    end
  end
  
  get "/groups" ...
  post "/groups" ...
  
  scope "/groups/:group_name" do
    pipe_through :group_name_api
    
    get "/" ...
    get "/devices" ...
    post "/devices" ...
    
    scope "/devices/:device_id" do
      pipe_through :device_api
    
      delete "/" ...
      
      scope "/interfaces" do
        pipe_through :interface_values_api
      
        get "/" ...
        get "/:interface" ...
        get "/:interface/:path" ...
        put "/:interface/:path" ...
        post "/:interface/:path" ...
        delete "/:interface/:path" ...
      end
    end
  end
end
```

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
```diff
@@ -22,6 +22,13 @@ defmodule Astarte.Pairing.APIWeb.Router do
   pipeline :api do
```
davidebriani, line 22: Since this is realm-scoped, let's call it `:realm_api` instead of `:api`.
```suggestion
  pipeline :realm_api do
```

apps/astarte_pairing_api/lib/astarte_pairing_api_web/router.ex
```diff
@@ -22,6 +22,13 @@ defmodule Astarte.Pairing.APIWeb.Router do
...
     plug :accepts, ["json"]
     plug Astarte.Pairing.APIWeb.Plug.LogRealm
+    plug Astarte.Pairing.APIWeb.Plug.AuthorizePath
+  end
+
+  pipeline :api_device do
+    plug :accepts, ["json"]
+    plug Astarte.Pairing.APIWeb.Plug.LogRealm
+    plug Astarte.Pairing.APIWeb.Plug.LogHwId
```
davidebriani, line 31: I would remove `plug :accepts, ["json"]` and `plug Astarte.Pairing.APIWeb.Plug.LogRealm`, since they are already applied by the `:realm_api` pipeline.

Let's use scopes to appropriately structure the routes and avoid applying the plugs multiple times.
The structure suggested by Arnaldo seems fine, was there some issues with it?

```elixir
  pipeline :realm_api do
    plug :accepts, ["json"]
    plug Astarte.Pairing.APIWeb.Plug.LogRealm
  end

  pipeline :agent_api do
    plug Astarte.Pairing.APIWeb.Plug.AuthorizePath
  end

  pipeline :devices_api do
    plug Astarte.Pairing.APIWeb.Plug.LogHwId
  end

  scope "/v1/:realm_name", Astarte.Pairing.APIWeb do
    pipe_through :realm_api

    get "/version", VersionController, :show

    scope "/agent" do
      pipe_through :agent_api

      post "/devices", AgentController, :create
      delete "/devices/:device_id", AgentController, :delete
    end

    scope "/devices" do
      pipe_through :devices_api

      get "/:hw_id", DeviceController, :show_info
      post "/:hw_id/protocols/:protocol/credentials", DeviceController, :create_credentials
      post "/:hw_id/protocols/:protocol/credentials/verify",
           DeviceController,
           :verify_credentials
    end
  end
```

davidebriani (changes requested):

Annopaolo (approved):

PR #895, opened 2024-02-05
DUP: stop custom handling of AMQP connections 

Annopaolo:
AMQPDataConsumer leaked channels when the process did not terminate correctly.
Replace the ad-hoc handling of connections and channels with the ExRabbitPool pooling library.
Contexually, add an env var to customize the number of connections to the broker:
- `DATA_UPDATER_PLANT_AMQP_CONSUMER_CONNECTION_NUMBER`  (default: 10)

As DUP is still using one channel per queue, the number of channels per connection is explicitly computed as ⎡`total_queue_number`/`DATA_UPDATER_PLANT_AMQP_CONSUMER_CONNECTION_NUMBER`⎤.


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_data_consumer.ex
```diff
@@ -164,57 +192,60 @@ defmodule Astarte.DataUpdaterPlant.AMQPDataConsumer do
...
-         {:ok, _consumer_tag} <- Basic.consume(chan, queue_name) do
-      {:ok, chan}
+      {:noreply, %State{state | channel: channel, monitor: ref}}
     else
       {:error, reason} ->
         Logger.warn("Error initializing AMQPDataConsumer: #{inspect(reason)}",
           tag: "data_consumer_init_err"
         )
 
-        {:stop, reason}
```
rbino, line 217: Probably the leak was happening here (if there was any kind of error the channel was opened and never closed)

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_data_consumer.ex
```diff
@@ -164,57 +192,60 @@ defmodule Astarte.DataUpdaterPlant.AMQPDataConsumer do
...
 
-  def handle_info({:async_init, queue_name}, _state) do
-    # This will block until Rabbit connection is up
-    rabbitmq_connection = ConnectionManager.get_connection()
+  defp schedule_connect() do
+    Process.send_after(self(), :connect, @reconnect_interval)
+  end
 
-    {:ok, chan} = initialize_chan(rabbitmq_connection, queue_name)
+  defp do_connect(state) do
```
rbino, line 205: I'd change the naming given you don't actually connect anymore here, maybe `init_consume` or something like that

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_events_producer.ex
```diff
@@ -44,73 +41,84 @@ defmodule Astarte.DataUpdaterPlant.AMQPEventsProducer do
     GenServer.call(__MODULE__, {:declare_exchange, exchange}, 60_000)
   end
 
-  # Server callbacks
+  # Gets a connection worker out of the connection pool, if there is one available
+  # takes a channel out of it channel pool, if there is one available
+  defp do_connect(state) do
```
rbino, line 46: See above regarding the naming

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_data_consumer.ex
```diff
@@ -94,64 +101,85 @@ defmodule Astarte.DataUpdaterPlant.AMQPDataConsumer do
 
   # Server callbacks
 
+  @impl true
   def init(args) do
     queue_name = Keyword.fetch!(args, :queue_name)
-    # Init asynchronously since we will wait for the AMQP connection
-    send(self(), {:async_init, queue_name})
-    {:ok, nil}
+    {:ok, %State{queue_name: queue_name}, {:continue, :connect}}
```
rbino, line 107: Does the async initialization provide any actual value or is this something we could simply remove while we're at it?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_events_producer.ex
```diff
@@ -44,73 +41,84 @@ defmodule Astarte.DataUpdaterPlant.AMQPEventsProducer do
...
+
+        schedule_connect()
+    end
   end
 
+  # Server callbacks
+
+  @impl true
+  def init(_opts) do
+    {:ok, [], {:continue, :connect}}
```
rbino, line 85: See above for async init

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_events_producer.ex
```diff
@@ -44,73 +41,84 @@ defmodule Astarte.DataUpdaterPlant.AMQPEventsProducer do
...
+        Process.monitor(channel_pid)
+
+        {:noreply, channel}
+
+      {:error, reason} ->
+        Logger.warn("RabbitMQ Connection error: #{inspect(reason)}",
+          tag: "events_producer_conn_err"
+        )
+
+        schedule_connect()
```
rbino, line 77: Here you're returning the `{:noreply, :not_connected}` returned by `schedule_connect()`, at line 54 you're ignoring it and returning another state instead.

`schedule_connect()` should used only for side effects (i.e. just sending the message) and the state must be handled outside of it (this also makes it clearer since you don't have to jump to the implementation of `schedule_connect` to see what it will return)

rbino (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_data_consumer.ex
Annopaolo: The AMQPDataConsumer may handle some `call`s while it is not consuming from a queue (e.g. [`start_messag_tracker`](https://github.com/astarte-platform/astarte/blob/7ff7557ceb4d511e6761e9d828ebb641f942f4ba/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_data_consumer.ex#L69-L73)), so the async init provides a way to do that even if it is still setting up the consuming  part

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_events_producer.ex
Annopaolo: In this case async init added nothing. It has been removed.

rbino (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_data_consumer.ex
```diff
@@ -164,57 +192,66 @@ defmodule Astarte.DataUpdaterPlant.AMQPDataConsumer do
...
+    conn = ExRabbitPool.get_connection_worker(:amqp_consumer_pool)
 
-    {:noreply, chan}
-  end
+    case ExRabbitPool.checkout_channel(conn) do
+      {:ok, channel} ->
+        _ =
+          Logger.debug(
+            "Successfully checked out channel for consumer on queue #{state.queue_name}"
+          )
```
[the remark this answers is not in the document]
Annopaolo: I would say this is not, as if the number of available channels is not enough, some queues will not be declared as `checkout_channel` will always fail

rbino (approved):

PR #899, opened 2024-02-08
Metrics: expose the number of bytes sent by Trigger Engine

### LINKED ISSUE #896: Metrics: expose the number of bytes sent by Trigger Engine
Currently, Trigger Engine metrics expose the number of HTTP triggers sent by the service, with realm name and status code as tags (`astarte_trigger_engine_http_action_executed_count`). Expose a new metric to display the cumulative number of bytes sent, keeping the same tag structure.

eddbbt:
Add telemetry in order to record the cumulative amount of bytes from triggers payload, resolve #896
Add new astarte.trigger_engine.http_action_executed.payload_bytes metric
Change how triggers are handled to provide accurate metric logging for number of retries and payload size, both are cumulative.
Payload size is recorded even in case of  http error (for example, when configured API returns 500)

### DIFF
--- a/apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
+++ b/apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
@@ -67,20 +67,36 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
          {:ok, action} <- retrieve_trigger_configuration(realm, trigger_id),
          {:ok, payload} <-
            event_to_payload(realm, device_id, event_type, event, action, timestamp_ms),
-         {:ok, headers} <- event_to_headers(realm, device_id, event_type, event, action),
-         :ok <- execute_action(payload, headers, action) do
-      :telemetry.execute(
-        [:astarte, :trigger_engine, :http_action_executed],
-        %{},
-        %{realm: realm, status: :ok}
-      )
-
-      :ok
+         {:ok, headers} <- event_to_headers(realm, device_id, event_type, event, action) do
+      send_simple_event(realm, payload, headers, action)
     else
+      error ->
+        Logger.warn("Error while processing event: #{inspect(error)}")
+
+        error
+    end
+  end
+
+  defp send_simple_event(realm, payload, headers, action) do
+    payload_size = :erlang.byte_size(payload)
+
+    case execute_action(payload, headers, action) do
+      :ok ->
+        :telemetry.execute(
+          [:astarte, :trigger_engine, :http_action_executed],
+          %{size: payload_size},
+          %{
+            realm: realm,
+            status: :ok
+          }
+        )
+
+        :ok
+
       {:error, {:http_error, status_code}} ->
         :telemetry.execute(
           [:astarte, :trigger_engine, :http_action_executed],
-          %{},
+          %{size: payload_size},
           %{realm: realm, status: status_code}
         )
 
@@ -89,16 +105,11 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
       {:error, :connection_error} ->
         :telemetry.execute(
           [:astarte, :trigger_engine, :http_action_executed],
-          %{},
+          %{size: payload_size},
           %{realm: realm, status: :connection_error}
         )
 
         {:error, :connection_error}
-
-      error ->
-        Logger.warn("Error while processing event: #{inspect(error)}")
-
-        error
     end
   end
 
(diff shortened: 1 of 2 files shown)


apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -62,41 +62,65 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
...
-      :telemetry.execute(
-        [:astarte, :trigger_engine, :http_action_executed],
-        %{},
-        %{realm: realm, status: :ok}
-      )
+         {:ok, action} <- retrieve_trigger_configuration(realm, trigger_id) do
+      with {:ok, payload} <-
+             event_to_payload(realm, device_id, event_type, event, action, timestamp_ms),
+           {:ok, headers} <- event_to_headers(realm, device_id, event_type, event, action) do
+        payload_size = payload |> :erlang.term_to_binary() |> :erlang.byte_size()
```
Annopaolo, line 78: It is wrong to include `:erlang.term_to_binary` in there:
- The payload is already a string (see the output of `event_to_payload/5)
- `term_to_binary` encodes a given term using the BEAM-specific encoding, that might differ from what one expects.


Indeed
```elixir
# An example of event_to_payload output
"{\"i_am\":\"a_map\"}" |> byte_size()
# > 16
"{\"i_am\":\"a_map\"}" |> :erlang.term_to_binary() |> byte_size()
# > 22
```

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -62,41 +62,65 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
...
+
+          {:error, :connection_error} ->
+            :telemetry.execute(
+              [:astarte, :trigger_engine, :http_action_executed],
+              %{size: payload_size},
+              %{realm: realm, status: :connection_error}
+            )
+
+            {:error, :connection_error}
+        end
```
Annopaolo, line 110: I'd suggest moving this in another function to decrease cognitive load when reading the code

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -62,41 +62,65 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
...
-           event_to_payload(realm, device_id, event_type, event, action, timestamp_ms),
-         {:ok, headers} <- event_to_headers(realm, device_id, event_type, event, action),
-         :ok <- execute_action(payload, headers, action) do
-      :telemetry.execute(
-        [:astarte, :trigger_engine, :http_action_executed],
-        %{},
-        %{realm: realm, status: :ok}
-      )
+         {:ok, action} <- retrieve_trigger_configuration(realm, trigger_id) do
+      with {:ok, payload} <-
```
Annopaolo, line 75: Actually, why is there a `with` inside a `with`? If the two error cases needs different handling, consider extracting the internal one to a new function.
This refactoring suggestion might come in hand: https://github.com/lucasvegi/Elixir-Code-Smells?tab=readme-ov-file#complex-else-clauses-in-with

apps/astarte_trigger_engine/lib/astarte_trigger_engine_web/telemetry.ex
```diff
@@ -60,6 +60,10 @@ defmodule Astarte.TriggerEngineWeb.Telemetry do
         tags: [:realm, :status],
         description: "TriggerEngine total http executed actions."
       ),
+      sum("astarte.trigger_engine.http_action_executed.size",
```
Annopaolo, line 63:
```suggestion
      sum("astarte.trigger_engine.http_action_executed.payload_bytes",
```

apps/astarte_trigger_engine/lib/astarte_trigger_engine_web/telemetry.ex
```diff
@@ -60,6 +60,10 @@ defmodule Astarte.TriggerEngineWeb.Telemetry do
...
+        tags: [:realm, :status],
+        description: "TriggerEngine total http executed actions size."
```
Annopaolo, line 65:
```suggestion
        description: "TriggerEngine total http executed actions payload size, in bytes."
```

Annopaolo (changes requested):

Annopaolo (changes requested):
Nitpicking

Annopaolo (approved):

PR #905, opened 2024-02-20
Metrics: expose API calls details in services

Annopaolo:
Add the total number of bytes exchanged via HTTP to metrics (both sent and received).
In general, each API service now exposes four new metrics:
- `<service_name>_requests_count`
- `<service_name>_requests_bytes`
- `<service_name>_responses_count`
- `<service_name>_responses_bytes`

Values are tagged by realm name, except for Housekeeping API.

In order not to slow down the handling of incoming calls, count and byte size are computed in a different process.


Annopaolo:
Superseded by #910.

PR #909, opened 2024-03-04
Edit Astarte keyspaces

eddbbt:
This pr enable multiple astarte instances to share a common db via keyspaces separation. With the recent deploy on google cloud marketplace this is likely to be a common scenario with shared and low tiers. 

At the actual state this project does not allow multiple instances to share the same db and thus is more prone to name conflicts for realms and organizations, due to the shared keyspaces and tables.

DB connector and queries have been edited to differentate beetween keyspaces via an env variable (called an astarte_instance_id), this way each deploy creates it's own astarte keyspace and for each realm, a keyspace wich name's prefixed by its own astarte_instance_id. 
Proper escaping and validation have been setup to assure there will be no conflicts and no incompatibilities with the selected scylla/cassandra db.

All tests for all astarte modules have been fixed and optimized for the new working logic, it was not necessary to change tests scope or goals.
Also the testing db has been corrected.

This pr is heavily dependant on:
- https://github.com/astarte-platform/astarte_core/pull/105 for keysapce transaltion and management function 
- https://github.com/astarte-platform/astarte_vmq_plugin/pull/85 for verneMQ module

There are dependencies pointing at the forked repos above, those needs to be changed before release

Further info can be found in notion

resolves https://github.com/astarte-platform/astarte/issues/924


apps/astarte_housekeeping/config/config.exs
```diff
@@ -2,4 +2,9 @@
 # and its dependencies with the aid of the Mix.Config module.
 import Config
 
+:persistent_term.put(
+  "customer_keyspace",
+  (System.get_env("keyspace") || "default") |> String.slice(0..39)
```
Annopaolo, line 7: Since we're using Skogsra to handle env vars (see https://github.com/astarte-platform/astarte/blob/master/apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex), let's use it for `"keyspace"` too.
Also, I think that calling this value `customer_keyspace` might be inaccurate, as every Astarte realm is mapped to a Cassandra keyspace. UUIDv5 uses `namespace` to refer to the fixed part to add to some data, also `instance_id` could be a good choice (as the issue arises only when multiple Astarte instances share the same database)

apps/astarte_housekeeping/config/test.exs
```diff
@@ -7,3 +7,8 @@ config :astarte_housekeeping, :enable_realm_deletion, true
 config :logger, :console,
   format: {PrettyLog.UserFriendlyFormatter, :format},
   metadata: [:realm, :function]
+
+:persistent_term.put(
+  "customer_keyspace",
+  "testkeyspace"
+)
```
Annopaolo, line 14: IIRC, if you use Skogsra you won't need this

Annopaolo (changes requested):

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -24,46 +24,92 @@ defmodule Astarte.Housekeeping.Queries do
 
   @default_replication_factor 1
 
+  def keyspace_from_realm(realm_name) do
+    customer_keyspace = Config.instance_id!()
+
+    case :persistent_term.get(customer_keyspace <> realm_name <> "encoded", nil) do
```
Annopaolo, line 30: But, but but! There is no need to store the `astarte_instance_id` as part  of the key in the persistent term storage, what you want is a map realm_name => keyspace_name.

So this function can become something like:
```elixir
def keyspace_from_name(realm_name) do
  case :persistent_term.get(realm_name, nil) do
    nil ->
      keyspace_name_base = Config.astarte_instance_id!() <> realm_name
      keyspace_name = some_clever_transformation_of(keyspace_name_base)
      :persistent_term.put(realm_name, keyspace_name)
    value -> value
  end
end
```
This will also apply to the `astarte` keyspace (even if it is a keyspace without realm)

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -270,6 +328,9 @@ defmodule Astarte.Housekeeping.Queries do
   end
 
   defp create_realm_keyspace(conn, realm_name, replication_map_str) do
+    # encoded
+    IO.inspect(realm_name)
```
Annopaolo, line 332: Leftover

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -1101,14 +1198,15 @@ defmodule Astarte.Housekeeping.Queries do
     end
   end
 
-  defp do_update_public_key(conn, realm_name, new_public_key) do
+  defp do_update_public_key(conn, keyspace_name, new_public_key) do
+    # encoded
     statement = """
     INSERT INTO :realm_name.kv_store (group, key, value)
```
Annopaolo, line 1204:
```suggestion
    INSERT INTO :keyspace_name.kv_store (group, key, value)
```

Annopaolo (changes requested):

osmanhadzic:
> WIP

Please rephrase both the commit message and the title.
In general, commits should describe:

Why a change is being made?
How does the change address the issue?
What effects does the change have?
Are there any limitations of the current code?
For the PR, you can take a look at [https://www.pullrequest.com/blog/writing-a-great-pull-request-description/](https://www.pullrequest.com/blog/writing-a-great-pull-request-description/)

eddbbt:
thank you @osmanhadzic, I appreciate your advice, It will be a priority to write a meaningful description and title after we figure out everything, but please consider this pr is currently a work in progress even at a design level

apps/astarte_appengine_api/lib/astarte_appengine_api/config.ex
```diff
@@ -128,6 +128,12 @@ defmodule Astarte.AppEngine.API.Config do
     type: :module,
     default: Astarte.RPC.AMQP.Client
 
+  @envdoc "The astarte instance id to generate keyspaces with"
+  app_env :astarte_instance_id, :astarte_appengine_api, :astarte_instance_id,
+    os_env: "ASTARTE_INSTANCE_ID",
+    default: "default",
```
rbino, line 134: The default `astarte_instance_id` should be an empty string, otherwise this would be a breaking change.

rbino (changes requested):

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -24,46 +24,92 @@ defmodule Astarte.Housekeeping.Queries do
...
+
   def create_realm(realm_name, public_key_pem, nil = _replication_factor, device_limit, opts) do
     create_realm(realm_name, public_key_pem, @default_replication_factor, device_limit, opts)
   end
 
   def create_realm(realm_name, public_key_pem, replication, device_limit, opts) do
+    keyspace_name = keyspace_from_realm(realm_name)
+
     with :ok <- validate_realm_name(realm_name),
+         :ok <- validate_realm_name(keyspace_name),
```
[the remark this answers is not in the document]
eddbbt: no it's not possible, fixed

rbino (changes requested):

shinnokdisengir (changes requested):

apps/astarte_housekeeping/lib/astarte_housekeeping/utils/astarte_instance_id_type.ex
```diff
@@ -0,0 +1,17 @@
+defmodule AstarteInstanceIdType do
+  use Skogsra.Type
+
+  @impl Skogsra.Type
+  def cast(value)
+
+  def cast(value) when is_binary(value) do
```
shinnokdisengir, line 7: I've seen doc of Skogsra a bit, but seems there are no best-pratice to raise inside cast (otherwise, it should have implemented `cast!`

I guess instance id is UTF8, so the best way to solve could be:

```elixir
def cast(value) when is_binary(value) and byte_size(value) <= 41 do
{:ok, value}
end
```

shinnokdisengir (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/auth/auth.ex
```diff
@@ -19,11 +19,15 @@
...
   alias Astarte.AppEngine.API.Queries
   alias Astarte.DataAccess.Database
+  alias Astarte.AppEngine.API.Config
+  alias Astarte.Core.CQLUtils
 
   require Logger
 
   def fetch_public_key(realm) do
-    with {:ok, client} <- Database.connect(realm: realm),
+    keyspace_name = CQLUtils.realm_name_to_keyspace_name(realm, Config.astarte_instance_id!())
```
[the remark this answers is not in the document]
eddbbt: this is a valid point, however, starting with the fact that Config.astarte_instance_id! must not crash (and so, return always a valid value) because is validated at the service start, if for any circumstances it crashes it's a unexpected behavior (thus the !) and it must crash the entire application (and leave a meaningful stacktrace).

Expecting a tuple is not valid because we don't expect this function to ever return error and bubbling up. 

On the other hand, it's not fetch_public_key or it's logic that crashes, but a function inside it , otherwise every function (or most of them) shall end with ! and/or catch an eventual exception to bubble up

apps/astarte_appengine_api/config/dev.exs
```diff
@@ -13,6 +13,8 @@ config :astarte_appengine_api, Astarte.AppEngine.APIWeb.Endpoint,
   check_origin: false,
   watchers: []
 
+config :astarte_appengine_api, :astarte_instance_id, ""
```
Annopaolo, line 16: This should not be needed as there's already Skogrsa handling the config and default (this holds for all other services, too)

apps/astarte_appengine_api/config/prod.exs
```diff
@@ -22,6 +22,8 @@ config :logger,
     [level_lower_than: :info]
   ]
 
+config :astarte_appengine_api, :astarte_instance_id, ""
```
Annopaolo, line 25: See above

apps/astarte_appengine_api/config/test.exs
```diff
@@ -27,6 +27,7 @@ config :logger, :console,
 config :astarte_rpc, :amqp_connection, host: System.get_env("RABBITMQ_HOST") || "rabbitmq"
 
 config :astarte_appengine_api, :rpc_client, MockRPCClient
+config :astarte_appengine_api, :astarte_instance_id, ""
```
Annopaolo, line 30: See above

apps/astarte_appengine_api/lib/astarte_appengine_api/utils/astarte_instance_id_type.ex
Annopaolo, line 1: I'm not sure we want this: `CQLUtils.realm_name_to_keyspace_name/2` has no limits on the instance id size, the worst that can happen is that the keyspace name is a base64 string, so enforcing the limit here seems too much
(this holds for all other services, too)

apps/astarte_appengine_api/mix.lock
Annopaolo, line 1: Some libraries here are not affected by the change, but have been updated nonetheless (e.g. `amqp_client`); I would expect only `astarte_core` and its dependencies to be updated

apps/astarte_appengine_api/test/astarte_appengine_api/device/map_tree_test.exs
```diff
@@ -19,6 +19,8 @@
 defmodule Astarte.AppEngine.API.MapTreeTest do
   use ExUnit.Case
   alias Astarte.AppEngine.API.Device.MapTree
+  alias Astarte.Core.CQLUtils
+  alias Astarte.AppEngine.API.Config
```
Annopaolo, line 23: Aliases not used

apps/astarte_appengine_api/test/astarte_appengine_api/stats/stats_test.exs
```diff
@@ -21,6 +21,8 @@ defmodule Astarte.AppEngine.API.StatsTest do
   alias Astarte.AppEngine.API.DatabaseTestHelper
   alias Astarte.AppEngine.API.Stats
   alias Astarte.AppEngine.API.Stats.DevicesStats
+  alias Astarte.Core.CQLUtils
+  alias Astarte.AppEngine.API.Config
```
Annopaolo, line 25: Unused aliases

apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/interface_values_by_device_alias_controller_test.exs
```diff
@@ -21,6 +21,8 @@ defmodule Astarte.AppEngine.APIWeb.InterfaceValuesByDeviceAliasControllerTest do
 
   alias Astarte.AppEngine.API.DatabaseTestHelper
   alias Astarte.AppEngine.API.JWTTestHelper
+  alias Astarte.Core.CQLUtils
+  alias Astarte.AppEngine.API.Config
```
Annopaolo, line 25: Unused aliases

Bottom text

apps/astarte_data_updater_plant/config/dev.exs
```diff
@@ -3,3 +3,5 @@ import Config
 config :logger, :console,
   format: {PrettyLog.LogfmtFormatter, :format},
   metadata: [:realm, :device_id, :ip_address, :module, :function, :tag]
+
+config :astarte_data_updater_plant, :astarte_instance_id, ""
```
Annopaolo, line 7: Same as AppEngine `dev.exs`

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -144,7 +151,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
       end)
 
     device_id_string = Device.encode_device_id(new_state.device_id)
-
+    # here prob
```
Annopaolo, line 154: Leftover

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -1696,8 +1735,12 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
            interface_major: major,
            match_path: "/*"
          }} ->
-          with :ok <-
-                 InterfaceQueries.check_if_interface_exists(state.realm, interface_name, major) do
+          keyspace_name =
+            CQLUtils.realm_name_to_keyspace_name(state.realm, Config.astarte_instance_id!())
+
+          with {:ok, _db_client} <- Database.connect(realm: keyspace_name),
```
Annopaolo, line 1741: Since `_db_client` is unused, this line can be removed

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -1711,8 +1754,12 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
            interface_major: major,
            match_path: match_path
          }} ->
-          with {:ok, %InterfaceDescriptor{automaton: automaton}} <-
-                 InterfaceQueries.fetch_interface_descriptor(state.realm, interface_name, major),
+          keyspace_name =
+            CQLUtils.realm_name_to_keyspace_name(state.realm, Config.astarte_instance_id!())
+
+          with {:ok, _db_client} <- Database.connect(realm: keyspace_name),
```
Annopaolo, line 1760: See above

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
```diff
@@ -493,6 +493,7 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandler do
     |> dispatch_event(target, policy)
   end
 
+  # make4 sure tihs does not need encoded realm
```
Annopaolo, line 496: Leftover?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/utils/astarte_instance_id_type.ex
```diff
@@ -0,0 +1,14 @@
+defmodule AstarteInstanceIdType do
```
Annopaolo, line 1: See same comment in AppEngine

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater_test.exs
```diff
@@ -1731,10 +1756,13 @@ defmodule Astarte.DataUpdaterPlant.DataUpdaterTest do
     # Check DUP start ack in deleted_devices table
     dup_start_ack_statement = """
     SELECT dup_start_ack
-    FROM #{realm}.deletion_in_progress
+    FROM #{CQLUtils.realm_name_to_keyspace_name(realm, Config.astarte_instance_id!())}.deletion_in_progress
     WHERE device_id = :device_id
     """
 
+    IO.inspect(encoded_device_id)
+    IO.inspect(realm)
```
Annopaolo, line 1764: Leftover

apps/astarte_data_updater_plant/test/support/amqp_test_helper.exs
```diff
@@ -15,6 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+Astarte.DataUpdaterPlant.AMQPTestHelper
```
Annopaolo, line 18: Leftover

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -28,6 +28,7 @@ defmodule Astarte.AppEngine.API.Device.Queries do
   alias Astarte.Core.Mapping
   alias Astarte.Core.Mapping.ValueType
   alias Astarte.Core.StorageType
+
```
Annopaolo, line 31: Leftover `\n`

apps/astarte_appengine_api/mix.exs
```diff
@@ -69,7 +69,7 @@ defmodule Astarte.AppEngine.API.Mixfile do
 
   defp astarte_required_modules(_) do
     [
-      {:astarte_core, github: "astarte-platform/astarte_core"},
+      {:astarte_core, github: "eddbbt/astarte_core", branch: "keyspace_feature", override: true},
```
Annopaolo, line 72: Now that the PR in `astarte_core` has been merged, you can point again to the main repo (here and in the other `mix.exs`s)

apps/astarte_data_updater_plant/config/test.exs
```diff
@@ -9,4 +9,5 @@ config :logger, :console,
   format: {PrettyLog.UserFriendlyFormatter, :format},
   metadata: [:realm, :device_id, :function]
 
+config :astarte_data_updater_plant, :astarte_instance_id, "test"
```
Annopaolo, line 12: Not strictly needed, but no problem imo to keep this if it's consistent (i.e. all other services have this `"test"` astarte instance id in test.exs)

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/health/queries.ex
```diff
@@ -18,11 +18,12 @@
 
 defmodule Astarte.DataUpdaterPlant.Health.Queries do
   require Logger
+  alias Astarte.DataUpdaterPlant.Config
 
   def get_astarte_health(consistency) do
     query = """
     SELECT COUNT(*)
-    FROM astarte.realms
+    FROM  #{Config.astarte_instance_id!()}astarte.realms
```
Annopaolo, line 26: This should be computed using ` CQLUtils.realm_name_to_keyspace_name("astarte", Config.astarte_instance_id!())` rather than just interpolating the instance id

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -1060,7 +1060,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
   def retrieve_realms! do
     statement = """
     SELECT *
-    FROM astarte.realms
+    FROM #{Config.astarte_instance_id!()}astarte.realms
```
Annopaolo, line 1063: This should be computed using ` CQLUtils.realm_name_to_keyspace_name("astarte", Config.astarte_instance_id!())` rather than just interpolating the instance id

apps/astarte_data_updater_plant/test/support/amqp_test_helper.exs
```diff
@@ -15,7 +15,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
```
Annopaolo, line 18: Leftover

apps/astarte_housekeeping/lib/astarte_housekeeping/migrator.ex
```diff
@@ -78,7 +78,7 @@ defmodule Astarte.Housekeeping.Migrator do
     query = """
     SELECT table_name
     FROM system_schema.tables
-    WHERE keyspace_name='astarte' AND table_name='kv_store'
+    WHERE keyspace_name='#{Config.astarte_instance_id!()}astarte' AND table_name='kv_store'
```
Annopaolo, line 81: This should be computed using ` CQLUtils.realm_name_to_keyspace_name("astarte", Config.astarte_instance_id!())` rather than just interpolating the instance id (also in the lines below)

apps/astarte_housekeeping/test/support/database_test_helper.ex
```diff
@@ -38,7 +39,7 @@ defmodule Astarte.Housekeeping.DatabaseTestHelper do
   end
 
   def drop_astarte_keyspace do
-    query = "DROP KEYSPACE astarte"
+    query = "DROP KEYSPACE #{Config.astarte_instance_id!()}astarte"
```
Annopaolo, line 42: This should be computed using ` CQLUtils.realm_name_to_keyspace_name("astarte", Config.astarte_instance_id!())` rather than just interpolating the instance id

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -836,10 +851,11 @@ defmodule Astarte.Housekeeping.Queries do
     # TODO: add support for creating the astarte keyspace with NetworkTopologyStrategy,
     # right now the replication factor is an integer so SimpleStrategy is always used
     astarte_keyspace_replication = Config.astarte_keyspace_replication_factor!()
+    instance_id = Config.astarte_instance_id!()
 
     with {:ok, replication_map_str} <- build_replication_map_str(astarte_keyspace_replication),
          query = """
-         CREATE KEYSPACE astarte
+         CREATE KEYSPACE #{instance_id}astarte
```
Annopaolo, line 858: The Astarte instance ID is arbitrary, so I suggest using ` CQLUtils.realm_name_to_keyspace_name("astarte", Config.astarte_instance_id!())` here too, just to make sure we do not go over the Scylla keyspace name length limit

Annopaolo (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/health/queries.ex
eddbbt: this is the main "*astarte" keyspace, it does not refer directly to a realm but rather to the entire instance (that can contains multiple realms), so is not correct to use
"realm_name_to_keyspace_name"

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
eddbbt: with 
` CQLUtils.realm_name_to_keyspace_name("astarte", Config.astarte_instance_id!())` 
it becames "astarteastarte.realms" breaking compatibility with old deployments

with  
`#{Config.astarte_instance_id!()}astarte.realms`
assuming an astarte_instance_id set as "" (old deployments) it becames "astarte.realms", preserving compatibility

Annopaolo: The interpolation `#{Config.astarte_instance_id!()}astarte.realms` leaks to the external world the implementation detail that an Astarte instance id must be shorter than 41 characters because of Scylla keyspace name length limit. This is not a business logic detail, and therefore it should be internally handled by Astarte.

As you also say, `#{CQLUtils.realm_name_to_keyspace_name("astarte", Config.astarte_instance_id!())}.realms` is backward-compatible if the instance id is set to `""`, as it will be for old Astartes.

I do not see the added value of keeping a different logic just for the `astarte` keyspace, since it both increases the complexity of the solution (by having two ways for keyspace names to be generated)  and adds an implicit constraint on the deployment (the instance id length)

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -44,13 +45,16 @@ defmodule Astarte.Housekeeping.Queries do
 
   def create_realm(realm_name, public_key_pem, replication, device_limit, max_retention, opts) do
     with :ok <- validate_realm_name(realm_name),
+         keyspace_name <-
+           CQLUtils.realm_name_to_keyspace_name(realm_name, Config.astarte_instance_id!()),
```
Annopaolo, line 49: Nit: in a `with` block, `<-` is used for matching, but here you're just binding the variabile (ok, matching is the way Elixir does binding, but...). Be more explicit using
```suggestion
         keyspace_name =
           CQLUtils.realm_name_to_keyspace_name(realm_name, Config.astarte_instance_id!()),
```

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -72,19 +77,24 @@ defmodule Astarte.Housekeeping.Queries do
...
-      do_delete_realm(realm_name)
+      do_delete_realm(realm_name, keyspace_name)
     end
   end
 
   def update_public_key(realm_name, new_public_key) do
-    with :ok <- validate_realm_name(realm_name) do
+    with :ok <- validate_realm_name(realm_name),
+         keyspace_name <-
+           CQLUtils.realm_name_to_keyspace_name(realm_name, Config.astarte_instance_id!()) do
```
Annopaolo, line 95: See above

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -951,14 +963,15 @@ defmodule Astarte.Housekeeping.Queries do
   end
 
   def is_realm_existing(realm_name) do
+    # TODO
```
Annopaolo, line 966: Leftover?

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -1021,9 +1035,10 @@ defmodule Astarte.Housekeeping.Queries do
   end
 
   def list_realms do
+    # unencoded
```
Annopaolo, line 1038: Leftover

apps/astarte_realm_management_api/test/astarte_realm_management_api/triggers/triggers_test.exs
```diff
@@ -18,6 +18,7 @@
 
 defmodule Astarte.RealmManagement.API.TriggersTest do
   use Astarte.RealmManagement.API.DataCase
+  alias Astarte.RealmManagement
```
Annopaolo, line 21: Leftover

Annopaolo (changes requested):

Annopaolo (approved):

apps/astarte_appengine_api/lib/astarte_appengine_api/auth/auth.ex
```diff
@@ -19,11 +19,15 @@
...
+
+    with {:ok, client} <- Database.connect(realm: keyspace_name),
```
rbino, line 30: You're going to hate me for this, _but_: I don't understand why you didn't just put the translation _inside_ the `Database.connect` implementation.

This would have solved 2 problems:
1. You wouldn't have had to touch the calling code
2. The current code has messed up naming: you're passing a `realm` option but you're actually passing a keyspace. Given that now those things are diverging, we should be extremely clear about where we're dealing with the `realm_name` and where we're dealing with the `keyspace`, otherwise we're causing unnecessary mental load when reading the code.

So my suggestion is just moving the translation inside `Database.connect` for all this sorts of call. Reverting shouldn't be too hard (the calling code basically remaing as-is and you just call `CQLUtils` in `astarte_data_access`).

Note that if the reason is "But `Config.astarte_instance_id()` comes from the application" then the second question is: why? That too should live in `astarte_data_access` and just `defdelegate`d in each `Config` application module (see, e.g., what we do for [`xandra_nodes`](https://github.com/astarte-platform/astarte/blob/4ebe50b4b17022ca60e4e83d9fc5f50526ed25b5/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/config.ex#L397))

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -1109,6 +1126,8 @@ defmodule Astarte.Housekeeping.Queries do
   end
 
   def set_datastream_maximum_storage_retention(realm_name, new_retention) do
+    realm_name = CQLUtils.realm_name_to_keyspace_name(realm_name, Config.astarte_instance_id!())
```
rbino, line 1129: For all that's not covered from the comment above on `Database.connect`, I'd like to see the correct naming for the same reason mentioned above.
If you do the substitution inline inside the query it's fine, but if you assign it to a variable then the variable (and all its uses below) should be named `keyspace`, not `realm_name`. This is wrong only in some instances, please check and fix where it is.

rbino (changes requested):

Annopaolo (approved):

apps/astarte_trigger_engine/lib/astarte_trigger_engine/events_consumer.ex
```diff
@@ -24,6 +24,7 @@ defmodule Astarte.TriggerEngine.EventsConsumer do
   alias Astarte.DataAccess.Database
   alias CQEx.Query, as: DatabaseQuery
   alias CQEx.Result, as: DatabaseResult
+
```
rbino, line 27: Fix

apps/astarte_realm_management/test/astarte_realm_management/engine_test.exs
```diff
@@ -903,7 +905,8 @@ defmodule Astarte.RealmManagement.EngineTest do
     assert Engine.list_interface_versions("autotestrealm", "com.ispirata.TestDatastream") ==
              {:ok, [[major_version: 0, minor_version: 10]]}
 
-    {:ok, client} = Database.connect(realm: "autotestrealm")
+    {:ok, client} =
+      Database.connect(realm: "autotestrealm")
```
rbino, line 909: Whitespace change (see all other above)

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -162,32 +172,31 @@ defmodule Astarte.Housekeeping.Queries do
...
+    with :ok <- delete_realm_keyspace(conn, keyspace_name),
          :ok <- remove_realm(conn, realm_name) do
       :ok
     else
       {:error, reason} ->
         _ =
           Logger.warning("Cannot delete realm: #{inspect(reason)}.",
             tag: "realm_deletion_failed",
-            realm: realm_name
+            realm: keyspace_name
```
rbino, line 199: Should be `realm_name` here

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -196,15 +205,17 @@ defmodule Astarte.Housekeeping.Queries do
...
          replication_map_str,
          device_limit,
          max_retention
        ) do
     Xandra.Cluster.run(:xandra, [timeout: 60_000], fn conn ->
       with :ok <- validate_realm_name(realm_name),
-           :ok <- create_realm_keyspace(conn, realm_name, replication_map_str),
-           {:ok, realm_conn} <- build_realm_conn(conn, realm_name),
+           :ok <- create_realm_keyspace(conn, keyspace_name, replication_map_str),
+           {:ok, realm_conn} <- build_realm_conn(conn, keyspace_name),
```
rbino, line 217: That's a `keyspace_conn`

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -196,15 +205,17 @@ defmodule Astarte.Housekeeping.Queries do
...
+           {:ok, realm_conn_plain} <- build_realm_conn(conn, realm_name),
```
rbino, line 218: And this just becomes `realm_conn`

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater_test.exs
```diff
@@ -1646,7 +1656,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdaterTest do
 
     DatabaseTestHelper.insert_device(device_id)
 
-    {:ok, db_client} = Database.connect(realm: realm)
+    {:ok, db_client} =
+      Database.connect(realm: realm)
```
rbino, line 1660: Leftovers (here and above)

apps/astarte_appengine_api/lib/astarte_appengine_api/stats/queries.ex
```diff
@@ -74,8 +75,11 @@ defmodule Astarte.AppEngine.API.Stats.Queries do
   # TODO: copypasted from Groups.Queries, this is going to be moved to Astarte.DataAccess
   # when we move everything to Xandra
   defp prepare_with_realm(conn, realm_name, query) do
+    keyspace_name =
+      CQLUtils.realm_name_to_keyspace_name(realm_name, Config.astarte_instance_id!())
+
     with {:valid, true} <- {:valid, Realm.valid_name?(realm_name)},
-         query_with_realm = String.replace(query, ":realm", realm_name),
+         query_with_realm = String.replace(query, ":realm", keyspace_name),
```
rbino, line 82: Naming: `query_with_keyspace` (and `:realm` should become `:keyspace`)

apps/astarte_appengine_api/lib/astarte_appengine_api/groups/queries.ex
```diff
@@ -523,8 +525,11 @@ defmodule Astarte.AppEngine.API.Groups.Queries do
   end
 
   defp prepare_with_realm(conn, realm_name, query) do
+    keyspace_name =
+      CQLUtils.realm_name_to_keyspace_name(realm_name, Config.astarte_instance_id!())
+
     with {:valid, true} <- {:valid, Realm.valid_name?(realm_name)},
-         query_with_realm = String.replace(query, ":realm", realm_name),
+         query_with_realm = String.replace(query, ":realm", keyspace_name),
```
rbino, line 532: Naming

rbino (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/room.ex
```diff
@@ -35,7 +35,7 @@ defmodule Astarte.AppEngine.API.Rooms.Room do
   alias Astarte.Core.Device
   alias Astarte.DataAccess.Database
   alias Astarte.AppEngine.API.Rooms.Queries
-
+  alias Astarte.AppEngine.API.Config
```
rbino, line 38: Leftover

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_test.exs
```diff
@@ -25,6 +25,8 @@ defmodule Astarte.AppEngine.API.DeviceTest do
   alias Astarte.AppEngine.API.Device.InterfaceValues
   alias Astarte.DataAccess.Database
   alias CQEx.Query, as: DatabaseQuery
+  alias Astarte.Core.CQLUtils
+  alias Astarte.AppEngine.API.Config
```
rbino, line 29: Leftovers

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_test.exs
```diff
@@ -964,6 +966,7 @@ defmodule Astarte.AppEngine.API.DeviceTest do
     device_id = "f0VMRgIBAQAAAAAAAAAAAA"
 
     {:ok, client} = Database.connect(realm: test)
+
```
rbino, line 969: Leftover

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater.ex
```diff
@@ -23,6 +23,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater do
   alias Astarte.DataUpdaterPlant.DataUpdater.Queries
   alias Astarte.DataAccess.Database
   alias Astarte.DataUpdaterPlant.MessageTracker
+
```
rbino, line 26: Leftover

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -2379,10 +2387,10 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
     end
   end
 
-  defp populate_triggers_for_object!(state, client, object_id, object_type) do
+  defp populate_triggers_for_object!(state, db_client, object_id, object_type) do
```
rbino, line 2390: Leftover variable name change

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -806,7 +806,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
     value
   end
 
-  def retrieve_endpoint_values(client, device_id, interface_descriptor, mapping) do
+  def retrieve_endpoint_values(db_client, device_id, interface_descriptor, mapping) do
```
rbino, line 809: Leftover variable name change

apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -196,28 +205,30 @@ defmodule Astarte.Housekeeping.Queries do
...
          public_key_pem,
          replication_map_str,
          device_limit,
          max_retention
        ) do
     Xandra.Cluster.run(:xandra, [timeout: 60_000], fn conn ->
       with :ok <- validate_realm_name(realm_name),
-           :ok <- create_realm_keyspace(conn, realm_name, replication_map_str),
+           :ok <- create_realm_keyspace(conn, keyspace_name, replication_map_str),
+           {:ok, keyspace_conn} <- build_realm_conn(conn, keyspace_name),
```
rbino, line 217: `realm_conn` here is just used to unpack the realm name and it's not very clear so I'd do this:
- rename `build_realm_conn` to `build_keyspace_conn`
- remove line `218` (not needed)
- pass the `realm_name` explicitly to `insert_realm` (i.e. `insert_realm(conn, realm_name, device_limit)`, it doesn't need the `keyspace_conn` since it acts on the `astarte` keyspace)

rbino (changes requested):

rbino (approved):

PR #912, opened 2024-03-05
Set and enforce a maximum storage TTL for datastreams 

Annopaolo:
While support for using a maximum period for datastreams data retention in a given realm is already present in DUP/AppEngine since 0.10, there was no support for setting its value.  Add it.
The value (in seconds) can be specified in Housekeeping API at realm creation, or changed with a realm update. When changed, it will apply oly to new data, i.e. data sent after the change.
Defaults to none, as before.

### DIFF
--- a/apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
+++ b/apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
@@ -27,6 +27,7 @@ defmodule Astarte.Housekeeping.Engine do
         public_key_pem,
         replication_factor,
         device_registration_limit,
+        max_retention,
         opts \\ []
       ) do
     _ =
@@ -35,7 +36,8 @@ defmodule Astarte.Housekeeping.Engine do
         tag: "create_realm",
         realm: realm,
         replication_factor: replication_factor,
-        device_registration_limit: device_registration_limit
+        device_registration_limit: device_registration_limit,
+        datastream_maximum_storage_retention: max_retention
       )
 
     Queries.create_realm(
@@ -43,6 +45,7 @@ defmodule Astarte.Housekeeping.Engine do
       public_key_pem,
       replication_factor,
       device_registration_limit,
+      max_retention,
       opts
     )
   end
@@ -99,14 +102,15 @@ defmodule Astarte.Housekeeping.Engine do
 
   @doc """
   Updates a given realm using the provided values map. Fails if realm does not exists or values are invalid.
-  At the moment, updates to jwt_public_key_pem and device_registration_limit are supported.
+  At the moment, updates to jwt_public_key_pem, device_registration_limit and datastream_maximum_storage_retention are supported.
   Returns a tuple:
   - either {:ok, updated_realm} where updated_realm is a map describing the realm
   - or {:error, reason} where reason is an atom describing the error.
   """
   @spec update_realm(String.t(), %{
           :jwt_public_key_pem => String.t(),
-          :device_registration_limit => any(),
+          :device_registration_limit => non_neg_integer() | :remove_limit | nil,
+          :datastream_maximum_storage_retention => non_neg_integer() | nil,
           optional(any()) => any()
         }) ::
           {:ok, map()}
@@ -130,12 +134,14 @@ defmodule Astarte.Housekeeping.Engine do
   defp do_update_realm(realm_name, update_attrs) do
     %{
       jwt_public_key_pem: new_jwt_public_key_pem,
-      device_registration_limit: new_limit
+      device_registration_limit: new_limit,
+      datastream_maximum_storage_retention: new_retention
     } = update_attrs
 
     with realm when is_map(realm) <- Queries.get_realm(realm_name),
          :ok <- update_jwt_public_key_pem(realm_name, new_jwt_public_key_pem),
-         :ok <- update_device_registration_limit(realm_name, new_limit) do
+         :ok <- update_device_registration_limit(realm_name, new_limit),
+         :ok <- update_datastream_maximum_storage_retention(realm_name, new_retention) do
       updated_realm = merge_realm_status(realm, update_attrs)
 
       _ = Logger.info("Successful update of realm #{realm_name}", tag: "realm_update_success")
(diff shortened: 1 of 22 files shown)


Annopaolo:
Based on https://github.com/astarte-platform/astarte_rpc/pull/81

eddbbt (approved):

apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
```diff
@@ -176,6 +182,31 @@ defmodule Astarte.Housekeeping.Engine do
...
+      )
+
+    set_datastream_maximum_storage_retention(realm_name, new_retention)
+  end
+
+  # ScyllaDB considers TTL=0 as unset, see
+  # https://opensource.docs.scylladb.com/stable/cql/time-to-live.html#notes
+  defp update_datastream_maximum_storage_retention(realm_name, 0) do
+    _ =
+      Logger.info("Removing datatream maximum storage retention",
```
rbino, line 199: typo: datatream

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api_web/controllers/realm_controller.ex
```diff
@@ -78,10 +78,12 @@ defmodule Astarte.Housekeeping.APIWeb.RealmController do
...
-    |> Map.replace_lazy("device_registration_limit", &normalize_device_registration_limit/1)
+    |> Map.replace_lazy(:device_registration_limit, &normalize_integer_or_nil/1)
+    |> Map.replace_lazy("device_registration_limit", &normalize_integer_or_nil/1)
+    |> Map.replace_lazy(:datastream_maximum_storage_retention, &normalize_integer_or_nil/1)
+    |> Map.replace_lazy("datastream_maximum_storage_retention", &normalize_integer_or_nil/1)
   end
 
-  defp normalize_device_registration_limit(value) when is_nil(value), do: :remove_limit
-  defp normalize_device_registration_limit(value), do: value
+  defp normalize_integer_or_nil(value) when is_nil(value), do: :none
```
rbino, line 87: I'd use `:remove` or `:unset` since this is the actual semantics (and renaming the `normalize` function accordingly). `:none` is unclear since it's too similar to `nil`.

apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
```diff
@@ -192,6 +223,22 @@ defmodule Astarte.Housekeeping.Engine do
...
+        :ok
+
+      {:error, reason} ->
+        _ =
+          Logger.warning(
+            "Cannot delete datastream maximum storage retention, error #{inspect(reason)}",
+            tag: "delete_datastream_maximum_storage_retention"
+          )
+
+        {:error, :delete_datastream_maximum_storage_retention_fail}
```
rbino, line 238: These are new errors so you have to handle them in Housekeeping APC in `extract_reply` in the `RPC` module and in `FallbackController`, otherwise an Internal Server Error is returned instead of a more informative error

apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
```diff
@@ -208,6 +255,23 @@ defmodule Astarte.Housekeeping.Engine do
...
+        :ok
+
+      {:error, reason} ->
+        _ =
+          Logger.warning(
+            "Cannot set datastream maximum storage retention, error #{inspect(reason)}",
+            tag: "set_datastream_maximum_storage_retention"
+          )
+
+        {:error, :set_datastream_maximum_storage_retention_fail}
```
rbino, line 271: Same here, see above

apps/astarte_housekeeping/lib/astarte_housekeeping/engine.ex
```diff
@@ -229,11 +293,24 @@ defmodule Astarte.Housekeeping.Engine do
...
   end
 
+  defp merge_realm_status(
+         old_realm,
+         %{datastream_maximum_storage_retention: 0} = update_attrs
+       ) do
+    attrs = Map.delete(update_attrs, :datastream_maximum_storage_retention)
+    merge_realm_status(%{old_realm | datastream_maximum_storage_retention: nil}, attrs)
+  end
+
```
rbino, line 313: Put the two clauses dealing with `datastream_maximum_storage_retention` close so related stuff are close together. Also a comment explaining the difference between the two cases wouldn't hurt since it's not clear.

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api_web/controllers/realm_controller.ex
Annopaolo: `:none` is needed because the cast in `NonNegativeIntegerOrNoneType`. Since both `device_registration_limit` and `datastream_maximum_storage_retention` use it (and the semantics slightly differs), I would stay with `:none`

rbino: Isn't the semantics "remove the limit if this is `nil`" for both of them? Or am I missing something? The previous iteration was already using `:remove` as a verb

Annopaolo: Datastream max storage retention uses both 0 and `nil` as valid values for removing the limit (as ScyllaDB uses TTL=0 to mean no TTL), while 0 is a valid value to set the device registration limit to.
A more explicit option would be to make 0 an invalid value for datastream max storage retention, leading to two Ecto custom types (one for { n | n > 0} U {:none} and one for { n | n >= 0} U {:none}), but I think that since this is actually an implementation detail due to the database we're using under the hood, it is better to hide the conversion from the user and perform it here https://github.com/astarte-platform/astarte/pull/912/files#diff-46c1424fa6f1d61351cd534f219215a0681236b66406f9be76e22bbc2b0a22acR287-R298

Annopaolo: Gone with `:unset` after IRL discussion

rbino (approved):

PR #913, opened 2024-03-08
RM: take maximum storage retention into account

Annopaolo:
Since we ~~will~~ allow to set/edit datastream max retention period in Housekeeping, add a way to retrieve its value in RM, too.

Moreover, as of today there are no checks  on whether an interface contains only mappings with `database_retention_ttl` lower than the maximum storage TTL allowed for the realm, making it possible to bypass the maximum allowed retention. Add this check on interface install and update. 
This change can be breaking, as some interfaces won't be installed anymore if the max allowed retention for the realm is set.

In the backend and RPC we use the Scylla [convention](https://opensource.docs.scylladb.com/stable/cql/time-to-live.html#notes) that TTL=0 means no TTL. However, the value given to the user if no maximum storage retention TTL is set will be the more understandable `null`.

### DIFF
--- a/apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
+++ b/apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
@@ -1073,4 +1073,27 @@ defmodule Astarte.RealmManagement.Engine do
         {:error, reason}
     end
   end
+
+  @doc """
+  Retrieves the maximum datastream storage retention of a realm.
+  Returns either `{:ok, limit}` or `{:error, reason}`.
+  The limit is a strictly positive integer (if set), 0 if unset.
+  """
+  @spec get_datastream_maximum_storage_retention(String.t()) ::
+          {:ok, non_neg_integer()} | {:error, atom()}
+  def get_datastream_maximum_storage_retention(realm_name) do
+    case Queries.get_datastream_maximum_storage_retention(realm_name) do
+      {:ok, value} ->
+        {:ok, value}
+
+      {:error, reason} ->
+        _ =
+          Logger.warning(
+            "Cannot get maximum datastream storage retention for realm #{realm_name}",
+            tag: "get_datastream_maximum_storage_retention_fail"
+          )
+
+        {:error, reason}
+    end
+  end
 end
--- a/apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
+++ b/apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
@@ -72,6 +72,7 @@ defmodule Astarte.RealmManagement.Engine do
          {:ok, json_obj} <- Jason.decode(interface_json),
          interface_changeset <- InterfaceDocument.changeset(%InterfaceDocument{}, json_obj),
          {:ok, interface_doc} <- Ecto.Changeset.apply_action(interface_changeset, :insert),
+         :ok <- verify_mappings_max_storage_retention(realm_name, interface_doc),
          interface_descriptor <- InterfaceDescriptor.from_interface(interface_doc),
          %InterfaceDescriptor{name: name, major_version: major} <- interface_descriptor,
          {:interface_avail, {:ok, false}} <-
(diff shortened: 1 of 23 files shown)


apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
```diff
@@ -937,6 +941,25 @@ defmodule Astarte.RealmManagement.Engine do
     end
   end
 
+  defp verify_mappings_max_storage_retention(realm_name, interface) do
+    with {:ok, max_retention} when is_integer(max_retention) <-
```
rbino, line 945: This will fallthrough with `{:ok, value}` if `max_retention` is not an integer. Is this what you want? I don't think so since the caller expects `:ok / {:error, reason}`

apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
```diff
@@ -937,6 +941,25 @@ defmodule Astarte.RealmManagement.Engine do
...
+           get_datastream_maximum_storage_retention(realm_name) do
+      if is_mappings_retention_valid?(interface.mappings, max_retention) do
```
rbino, line 947: Functions with `?` should not start with `is`

apps/astarte_realm_management/test/astarte_realm_management/engine_test.exs
```diff
@@ -833,6 +854,16 @@ defmodule Astarte.RealmManagement.EngineTest do
     assert Engine.delete_interface("autotestrealm", "com.ObjectAggregation", 0) == :ok
   end
 
+  test "fail to install interface when database_retention_ttl is higher than the maximum" do
```
rbino, line 857: Also add a success test (i.e. a realm with an explicit `datastream_maximum_storage_retention` + an interface with a lower TTL) to test all the combinations

rbino (changes requested):

apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
shinnokdisengir: ops. I've found an "example" written by @bettio 
`Queries.is_interface_major_available?`
I'm removing it

apps/astarte_realm_management/lib/astarte_realm_management/engine.ex
```diff
@@ -937,6 +941,25 @@ defmodule Astarte.RealmManagement.Engine do
...
   end
 
+  defp verify_mappings_max_storage_retention(realm_name, interface) do
+    with {:ok, max_retention} <- get_datastream_maximum_storage_retention(realm_name),
+         {:mappings_retention_valid, true} <-
+           {:mappings_retention_valid,
+            mappings_retention_valid?(interface.mappings, max_retention)} do
+      :ok
+    else
+      {:mappings_retention_valid, false} -> {:error, :maximum_database_retention_exceeded}
```
rbino, line 951: Just do
```elixir
defp verify_mappings_max_storage_retention(realm_name, interface) do
  with {:ok, max_retention} <- get_datastream_maximum_storage_retention(realm_name) do
    if mappings_retention_valid?(interface.mappings, max_retention) do
      :ok
    else
      {:error, :maximum_database_retention_exceeded}
    end
  end
end
```

Some reasons why:
- It's better having `with` without `else`
- Tuple-tagging the rhs in `with` blocks seemed a good idea at the time so you'll find code like this around the codebase, but it's actually not and it's less readable than the `if` version

The other possibility is changing the validation function to return `:ok/{:error, :maximum_database_retention_exceeded}` instead of a boolean (usually that kind of function is then called `validate_*`, so `validate_mappings_retention`), then you can just do

```elixir
defp verify_mappings_max_storage_retention(realm_name, interface) do
  with {:ok, max_retention} <- get_datastream_maximum_storage_retention(realm_name) do
    validate_mappings_retention(interface.mappings, max_retention)
  end
end
```

rbino (changes requested):

rbino (approved):

PR #919, opened 2024-03-29
Refactoring test suite using StreamData

shinnokdisengir:
TODO

refactoring test suite


apps/astarte_appengine_api/test/astarte_appengine_api/v2_group_test.exs
```diff
@@ -0,0 +1,129 @@
...
+          "group_name" => group_name,
+          "devices" => devices
+        }
+
+        response = post(auth_conn, groups_path(auth_conn, :create, keyspace), data: params)
+        assert [error] === json_response(response, 422)["errors"]["group_name"]
+      end
+    end
+
+    property "devices list empty", %{auth_conn: auth_conn, keyspace: keyspace} do
```
Annopaolo, line 48: See above

```suggestion
    property "fails when devices list empty", %{auth_conn: auth_conn, keyspace: keyspace} do
```

apps/astarte_appengine_api/test/astarte_appengine_api/v2_group_test.exs
```diff
@@ -0,0 +1,129 @@
...
+      assert "Group already exists" === json_response(response, 409)["errors"]["detail"]
+    end
+
+    property "creates groups with valid parameters", %{
+      auth_conn: auth_conn,
+      cluster: cluster,
+      keyspace: keyspace,
+      interfaces: interfaces,
+      devices: devices,
+      groups: groups
```
Annopaolo, line 103: Make explicit that these are old groups

```suggestion
      groups: old_groups
```

apps/astarte_appengine_api/test/astarte_appengine_api/v2_group_test.exs
```diff
@@ -0,0 +1,129 @@
...
+        response = post(auth_conn, groups_path(auth_conn, :create, keyspace), data: params)
+        assert params === json_response(response, 201)["data"]
+        response = get(auth_conn, groups_path(auth_conn, :show, keyspace, group_name))
+        assert group_name === json_response(response, 200)["data"]["group_name"]
+
+        for device <- device_ids do
+          {:ok, %DeviceStatus{groups: groups}} = Device.get_device_status!(keyspace, device)
+          assert group_name in groups
+          # TODO right way
+          # assert [group_name] === groups -- old_group_names
```
Annopaolo, line 124: If we're just testing that a group is being created, why checking that every device in the group has it in its status? This is probably a (slightly) different property that should be checked in another test

apps/astarte_appengine_api/test/astarte_appengine_api/v2_interface_test.exs
```diff
@@ -0,0 +1,50 @@
...
+  describe "interface generator" do
+    @tag timeout: :infinity
+    property "validate interface" do
+      check all interface <- InterfaceGenerator.interface() do
+        %Changeset{valid?: valid, errors: errors} = Interface.changeset(interface)
+
+        assert valid, "Invalid interface: " <> (errors |> Enum.join(", "))
+      end
+    end
+  end
```
Annopaolo, line 33: See the above comment for DeviceGenerator: what you're testing is that the  generates a valid interface, but it is exactly what the InterfaceGenerator is supposed to do

apps/astarte_appengine_api/test/astarte_appengine_api/v2_interface_test.exs
```diff
@@ -0,0 +1,50 @@
...
+
+  describe "interfaces db" do
```
Annopaolo, line 35: See above comment on DeviceGenerator too

apps/astarte_appengine_api/test/support_v2/cases/database.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Astarte.Test.Cases.Database do
+  use ExUnit.CaseTemplate
+
+  using do
+    quote do
+      # Do nothing for now
+    end
+  end
```
Annopaolo, line 8: If this is not used, it can be removed, see this example from ExUnit doc: https://hexdocs.pm/ex_unit/ExUnit.CaseTemplate.html#module-example

apps/astarte_appengine_api/test/support_v2/cases/device.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Astarte.Test.Cases.Device do
+  use ExUnit.CaseTemplate
+  use Astarte.Test.Cases.Interface
+
+  using do
+    quote do
+      # Do nothing for now
+    end
+  end
```
Annopaolo, line 9: Remove this (see above)

apps/astarte_appengine_api/test/support_v2/cases/group.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Astarte.Test.Cases.Group do
+  use ExUnit.CaseTemplate
+  use Astarte.Test.Cases.Device
+
+  using do
+    quote do
+      # Do nothing for now
+    end
+  end
```
Annopaolo, line 9: Remove this (see above)

apps/astarte_appengine_api/test/support_v2/cases/interface.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Astarte.Test.Cases.Interface do
+  use ExUnit.CaseTemplate
+  use Astarte.Test.Cases.Database
+
+  using do
+    quote do
+      # Do nothing for now
+    end
+  end
```
Annopaolo, line 9: Remove this (see above)

apps/astarte_appengine_api/test/support_v2/common.ex
Annopaolo, line 1: This file is empty and can be removed

apps/astarte_appengine_api/test/support_v2/generators/datetime.ex
```diff
@@ -0,0 +1,14 @@
...
+  @min_default 0
+  @max_default 2_556_143_999
+
+  def date_time(opts \\ []) do
+    [min: min, max: max] = Keyword.validate!(opts, min: @min_default, max: @max_default)
+    date_time(min, max)
+  end
+
+  defp date_time(min, max) when min < max, do: integer(min..max) |> map(&DateTime.from_unix!(&1))
+  defp date_time(min, max), do: raise("Datetime generator, received min: #{min} >= max: #{max}")
```
Annopaolo, line 13: Nit: when a function `raise`s, generally it is postfilxed with `!`

apps/astarte_appengine_api/test/support_v2/generators/device.ex
```diff
@@ -0,0 +1,116 @@
+defmodule Astarte.Test.Generators.Device do
+  alias Astarte.Core.Device
+  alias Astarte.Test.Generators.Common, as: CommonGenerator
+  alias Astarte.Test.Generators.DateTime, as: DateTimeGenerator
+  use ExUnitProperties
+
+  def id() do
+    # constant(Device.random_device_id())
```
Annopaolo, line 8: Leftover

apps/astarte_appengine_api/test/support_v2/generators/mapping.ex
```diff
@@ -0,0 +1,91 @@
+defmodule Astarte.Test.Generators.Mapping do
+  use ExUnitProperties
+
+  alias Astarte.Core.Mapping
+
+  defp endpoint(%{aggregation: aggregation, prefix: prefix}) do
+    # TODO: thinking about it
```
Annopaolo, line 7: Make this TODO a little more explicit so that the next person working on this knows what there's to think about

apps/astarte_appengine_api/test/support_v2/helpers/database.ex
```diff
@@ -0,0 +1,622 @@
...
+        Xandra.Batch.add(acc, prepared, %{
+          "name" => interface.name,
+          "interface_id" => interface.interface_id,
+          "major_version" => interface.major_version,
+          "minor_version" => interface.minor_version,
+          "type" => InterfaceType.to_int(interface.type),
+          "automaton_accepting_states" => :erlang.term_to_binary(:automaton_accepting_states),
+          "automaton_transitions" => :erlang.term_to_binary(:automaton_transitions),
+          "aggregation" => AggregationType.to_int(interface.aggregation),
+          # TODO Mapping to the other table
```
Annopaolo, line 391: Leftover?

Annopaolo (changes requested):
Also, some files are missing the copyright header

apps/astarte_appengine_api/test/astarte_appengine_api/v2_device_test.exs
```diff
@@ -0,0 +1,51 @@
...
+  @moduletag interface_count: 10
+  @moduletag device_count: 100
+
+  describe "device generator" do
+    property "validate device with pre-generated interfaces", %{interfaces: interfaces} do
+      check all device <- DeviceGenerator.device(interfaces: interfaces) do
+        :ok
+      end
+    end
+  end
```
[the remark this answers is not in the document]
shinnokdisengir: In a nutshell, generators are part of the tests, but only as a use, so they need unit tests themselves

apps/astarte_appengine_api/mix.exs
```diff
@@ -93,7 +93,8 @@ defmodule Astarte.AppEngine.API.Mixfile do
       # Required by :phoenix_swagger, otherwise it fails finding ex_json_schema.app
       {:ex_json_schema, "~> 0.7"},
       {:phoenix_swagger, "~> 0.8"},
-      {:xandra, "~> 0.13"},
+      {:xandra, "~> 0.18"},
+      {:exandra, "~> 0.10.2"},
```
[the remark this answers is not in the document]
shinnokdisengir: I am waiting for the implementation of exandra in `astarte_core`/`astarte_appengine_api`. Without the initially implemented instance of `exandra`, I cannot use it

apps/astarte_appengine_api/test/astarte_appengine_api/v2_device_test.exs
```diff
@@ -0,0 +1,51 @@
...
+
+  describe "devices db" do
+    test "validate insered devices", %{
```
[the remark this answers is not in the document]
shinnokdisengir: See above (testing generators). Unit testing for _cases_ and _fixtures_

apps/astarte_appengine_api/test/astarte_appengine_api/v2_device_test.exs
```diff
@@ -0,0 +1,51 @@
+defmodule Astarte.AppEngine.API.V2DeviceTest do
+  use ExUnit.Case, async: true
+  use ExUnitProperties
+  use Astarte.Test.Cases.Device
```
Annopaolo, line 4: This `use` hides the setup step(s) needed to populate the db, so it is unclear where e.g. devices selected at line 39 do come from.  It should be possible to clearly identify what is being created, what operation is being performed and what the expected output would be by just looking at a single test.
I find that explicitly implementing `setup_all` and `setup` in the test case helps doing so.

The same goes for the other tests on groups and interfaces.

apps/astarte_appengine_api/test/support_v2/helpers/database.ex
```diff
@@ -0,0 +1,622 @@
...
+  """
+
+  @select_devices """
+  SELECT * FROM :keyspace.devices WHERE device_id IN :device_ids
+  """
+
+  ###
+  ### Delete
+  @delete_interface """
+  DELETE FROM :keyspace.interfaces WHERE name = :name
```
Annopaolo, line 309: If tests from different cases run in different keyspaces and, inside a case, each test populates the tables it needs and truncates them on exit, you get database isolation for free, without the need to write more complex queries for deleting only some data.
A quick sketch of a possible way of handling it:
```elixir
defmodule MyTest do
  use ExUnit.Case
  
  setup_all do
     create_keyspace!("keyspace_#{System.unique_integer()}")
  end
  
  setup do
    # get keyspace from context 
    insert!(keyspace, devices_fixture())
    on_exit(fn -> run!("TRUNCATE TABLE devices") end)
  end

  test "my test 1" do
  # use the `devices` table
  end
end
```

apps/astarte_appengine_api/test/support_v2/setups/device.ex
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def setup(%{cluster: cluster, keyspace: keyspace, devices: devices}) do
+    on_exit(fn ->
+      DatabaseHelper.delete!(:device, cluster, keyspace, devices)
+    end)
+
+    DatabaseHelper.insert!(:device, cluster, keyspace, devices)
+    :ok
+  end
```
Annopaolo, line 17: These two can be moved in (or called from) a `setup` block , instead of being in a different module

apps/astarte_appengine_api/test/support_v2/cases/device.ex
```diff
@@ -0,0 +1,17 @@
...
+
+  alias Astarte.Test.Setups.Device, as: DeviceSetup
+
+  setup_all [
+    {DeviceSetup, :init},
+    {DeviceSetup, :setup}
+  ]
```
Annopaolo, line 16: Note that `setup_all` runs once before all tests in a case. In order to improve test isolation and clarity, it would be better to generate device data for each test (i.e. with `setup`).
The same goes also for groups and interfaces

Annopaolo (changes requested):
What's being done here is very good. In my opinion, the overall structure of tests can be improved for enhancing isolation and clarity. My suggestion is:
    - A `setup_all` part in which a realm - whose name is randomly generated - is created, for each test file (i.e. each ExUnit.Case). This allows for non-interference of tests when running in parallel.
    - A `setup` part, for each test, in which tables relevant for the test are populated and truncated on exit. This allows for non-interference of tests running in the same case and using the database.
        - Note: cleanup is necessary only when tests change database state. For example, a test “fail to create group with invalid name” does not need a cleanup, as no group will be added to the db; a “succeeds to create group with valid input” will need cleanup.

In general, by looking at a single test it should be possible to clearly identify what is being created, what operation is being performed and what the expected output would be.

apps/astarte_appengine_api/test/support_v2/helpers/database.ex
shinnokdisengir: I have already discussed this with @rbino. The point is to write as little code as possible in each test file. Cases are made to achieve this

apps/astarte_appengine_api/test/support_v2/helpers/database.ex
shinnokdisengir: currently, `select!(:interface)` doesn't join with mapping table. It's need to be done

PR #921, opened 2024-04-03
Add support for double values in payload cast

eddbbt:
Add cast to double in case of update of a server owned property, before sending data to device

Resolves first and second part of https://github.com/astarte-platform/astartectl/issues/239

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
@@ -747,6 +747,22 @@ defmodule Astarte.AppEngine.API.Device do
     end
   end
 
+  # conversion adding 0.0 is required because anyvalue comes from a json value,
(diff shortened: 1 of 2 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -755,6 +755,20 @@ defmodule Astarte.AppEngine.API.Device do
     end
   end
 
+  defp cast_value(:double, anyvalue) do
+    {:ok, anyvalue + 0.00}
```
Annopaolo, line 759:
```suggestion
    {:ok, anyvalue + 0.0}
```
Also, add a comment to remember why we're doing this (JSON does not distinguish between ints and floats etc etc)

Annopaolo (changes requested):
I think this should be rebased on `release-1.0` and forward-ported to master, as the error is present also there.
Remember to add an entry to the changelog

Annopaolo (approved):

Issue #933, opened 2024-05-02
The device registration limit prevents re-registration of already existing devices


davidebriani:
The device registration limit set for an Astarte realm is meant to prevent device registration when the limit is reached.

The main purpose is to restrict the ability of realm users to create new devices that would allocate more resources.

However, when the limit is reached Astarte also currently forbids re-registering existing devices.
This operation could arguably be allowed since it is useful for generating a new credential secret for the device when the previous one is lost or compromised.

Annopaolo:
Closed by #940

PR #940, opened 2024-05-08
Pairing: allow re-registering devices even if limit is reached

Annopaolo:
The device registration limit set for an Astarte realm is meant to prevent registration of new devices when the limit is reached. However, it blocked re-registration of already existing devices.
Allow to re-register existing devices without checking if registration limit has been reached.
Closes  [#933](https://github.com/astarte-platform/astarte/issues/933).

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/queries.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/queries.ex
@@ -103,7 +103,7 @@ defmodule Astarte.Pairing.Queries do
   end
 
   def unregister_device(client, device_id) do
-    with :ok <- check_already_registered_device(client, device_id),
+    with :ok <- verify_already_registered_device(client, device_id),
          :ok <- do_unregister_device(client, device_id) do
       :ok
     else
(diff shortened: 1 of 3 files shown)


eddbbt (approved):

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -117,6 +117,34 @@ defmodule Astarte.Pairing.Queries do
     end
   end
 
+  def verify_already_registered_device(realm_name, device_id) do
```
shinnokdisengir, line 120: In my opinion it would be more correct, since it returns `{:ok, boolean}`, to change the function to `verify_already_registered_device?(...)`. 
Also, the function is a bit too specific... might it be worth changing the pattern/logic to simply use `select_device_by_id(realm_name, device_id)` and manage externally whether it is present/registered in the db or not?

shinnokdisengir (changes requested):

apps/astarte_pairing/lib/astarte_pairing/queries.ex
Annopaolo: On function name: we usually follow the `check_*` convention for `{:ok, bool}` results (see e.g. `check_device_exists/2` in RealmManagement), good catch!


On function logic: code that is already in place uses the pattern of selecting only relevant information (see e.g. `select_device_for_credentials_request/2` and `select_device_for_info/2`), I have followed the same pattern. Your suggestion would be interesting when refactoring Pairing!
Also, note that a function that performs the same query does already exist (`check_already_registered_device/2`), however it uses CQEx and therefore it needs a CQEx client rather than a simple realm name  to run. I decided to keep `engine.ex` code simple and duplicate the functionality in `queries.ex` - not only for that, but also because we will soon™ remove CQEx entirely.

davidebriani (approved):

PR #956, opened 2024-06-06
Setup build for dev and prod

shinnokdisengir:
To set up the build for local development, we need several project modifications related to container settings. Specifically:

- Different Dockerfiles are needed (Dockerfile and Dockerfile.dev)
- Different docker-compose.yaml files are needed (docker-compose.yaml and docker-compose-dev.yaml)

These modifications are necessary for an optimal development environment, including hot code reloading.

- Added dev mode to astarte_appengine_api


Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+WORKDIR /src
+
+FROM base as deps
+# Cache elixir deps
+ADD mix.exs mix.lock ./
+RUN mix do deps.get, deps.compile
+
+FROM deps as builder
+# Add all the rest
+ADD . .
```
Annopaolo, line 29: If everything is being added, could this stage be skipped?

Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+
+# ------------------------
```
Annopaolo, line 31: Remove (or put the separator between every stage, for consistency)

Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+
+# Symlink to the service, to make a single entry point 
+# for all the apps
+RUN APP_NAME=$(ls | head -n 1) && \
+    ln -s ${APP_NAME}/bin/${APP_NAME} ${ASTARTE_SERVICE} 
+
+USER nobody
+CMD ./pre-cmd.sh && ./${ASTARTE_SERVICE} start
+
+
```
Annopaolo, line 81: Leftover newlines

apps/astarte_appengine_api/.dockerignore
```diff
@@ -11,6 +11,8 @@ erl_crash.dump
 *.beam
 config/*.secret.exs
 .elixir_ls/
+README.md
+.formatter.exs
```
Annopaolo, line 15: Leftover?

apps/astarte_appengine_api/.dockerignore
```diff
@@ -23,3 +25,5 @@ config/*.secret.exs
 .rebar
 rel/example_project
 ebin/*.beam
+
+# Project spec
```
Annopaolo, line 29: Remove this

docker-compose.dev.yml
```diff
@@ -0,0 +1,88 @@
...
+  astarte-housekeeping-api:
+    build:
+      target: builder
+    command: >
+      mix phx.server
+    develop:
+      watch:
+        - path: apps/astarte_housekeeping_api/lib
+          action: sync
+          target: /src/lib
```
Annopaolo, line 22: I'm not sure we want to trigger a reload only on `lib` changes. I would include also `mix.exs` and `mix.lock` so that a project setting or a library is modified, the change is watched.
The same goes for all other services

docker-compose.dev.yml
```diff
@@ -0,0 +1,88 @@
...
+          action: rebuild
+  astarte-trigger-engine:
+    build:
+      target: builder
+    command: >
+      mix run --no-halt
+    develop:
+      watch:
+        - path: apps/astarte_trigger_engine/lib
+          action: rebuild
```
Annopaolo, line 88: Watch out for trailing newlines!

rabbitmqadmin
Annopaolo, line 1: Leftover file

standard-interfaces/org.astarte-platform.genericevents.DeviceEvents.json
Annopaolo, line 1: This standard interface should not be removed

standard-interfaces/org.astarte-platform.rust.e2etest.ServerProperty.json
Annopaolo, line 1: This interface should not be added

docker-compose.yml
```diff
@@ -1,8 +1,8 @@
-version: '3.8'
```
Annopaolo, line 1: Is this needed?

Annopaolo (changes requested):
The general framework looks good to me. I see however some (minor) issues:
- There is no documentation on how this change impacts our build, as it is no more as easy as `docker build . -t $APP_NAME`. Commit messages should include a small comment on this
- There is no documentation on how to use the new `docker-compose.dev.yaml`. Commit messages should include a small comment on this
- The PR description seems to be outdated
- Imo, changes to Dockerfiles and to the `docker-compose.yaml` have a different scope and could be splitted in two different commits

Dockerfile
shinnokdisengir: The separator divides _prod_ between _dev_. No problem to add it to each layer

Dockerfile
shinnokdisengir: The layer explicitly separates the compilation of dependencies and the build of the code, to prevent the `rebuild` from taking a long time. You could also manage without the layer, but it seems clearer this way - by porting from the previous version I just lost this information -

apps/astarte_appengine_api/.dockerignore
shinnokdisengir: Mumble, isn't it better to create a separate section for each project?

apps/astarte_appengine_api/.dockerignore
shinnokdisengir: se above

docker-compose.yml
shinnokdisengir: Yes, become obsolete -> Warning

Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+RUN mix local.hex --force && \
+    mix local.rebar --force && \
+    mix hex.info
+
+WORKDIR /src
+
+FROM base as deps
+# Cache elixir deps
+ADD mix.exs mix.lock ./
+RUN mix do deps.get, deps.compile
```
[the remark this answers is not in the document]
shinnokdisengir: there is no way to hot-reload if you change MIX_ENV, so you must `docker compose down` and `up/watch` again

apps/astarte_housekeeping/pre-cmd.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+./astarte-service eval Elixir.Astarte.Housekeeping.ReleaseTasks.init_database || exit 1
+./astarte-service eval Elixir.Astarte.Housekeeping.ReleaseTasks.migrate || exit 1
```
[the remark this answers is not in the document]
shinnokdisengir: I don't understand...Is this an alternative?

docker-compose.dev.yml
```diff
@@ -0,0 +1,88 @@
+services:
+  astarte-housekeeping:
+    build:
+      target: builder
+    command: >
+      mix run -e Elixir.Astarte.Housekeeping.ReleaseTasks.init_database || exit 1
+      mix run -e Elixir.Astarte.Housekeeping.ReleaseTasks.migrate || exit 1
+      mix run --no-halt
```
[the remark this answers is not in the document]
shinnokdisengir: Could it be the docker/docker-compose version?

apps/astarte_housekeeping/pre-cmd.sh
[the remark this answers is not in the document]
noaccOS: `set -e` exits (with error) if one command fails. see https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html for reference

Dockerfile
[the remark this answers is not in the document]
noaccOS: line 40 of this file will recompile the dependencies

Dockerfile
noaccOS: Maybe a comment could be more significant than a separator in this case

standard-interfaces/org.astarte-platform.genericevents.DeviceEvents.json
shinnokdisengir: ops, sorry

apps/astarte_housekeeping/pre-cmd.sh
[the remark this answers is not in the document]
shinnokdisengir: done!

Dockerfile
shinnokdisengir: I've just added both, one doesn't seem to be enough

Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+
+COPY --from=release --chown=nobody:nobody /rel/* .
+
+# Symlink to the service, to make a single entry point 
+# for all the apps
+RUN APP_NAME=$(ls | head -n 1) && \
+    ln -s ${APP_NAME}/bin/${APP_NAME} ${ASTARTE_SERVICE} 
+
+USER nobody
+CMD ./pre-cmd.sh && ./${ASTARTE_SERVICE} start
```
Annopaolo, line 80: Nit: something is still wrong with the final newline

Dockerfile
Annopaolo: Ah, now I get it!
Then it's ok for me. I would suggest to make clearer however which sections of this dockerfile are common, which ones are only for `dev` (if any) and which ones are only for `prod`. The `#----` separator alone looks still a bit obscure

apps/astarte_appengine_api/.dockerignore
```diff
@@ -22,4 +22,4 @@ config/*.secret.exs
 # rebar 2.x
 .rebar
 rel/example_project
-ebin/*.beam
+ebin/*.beam
```
Annopaolo, line 25: Nit: something is still wrong with the final newline

docker-compose.yml
```diff
@@ -286,4 +294,4 @@ volumes:
 
 networks:
   default:
-    name: astarte
+    name: astarte
```
Annopaolo, line 297: Nit: something is still wrong with the final newline

Annopaolo (changes requested):

Dockerfile
shinnokdisengir: After discussing it with @noaccOS, I've just added a comment as well

davidebriani (approved):

davidebriani (changes requested):

apps/astarte_housekeeping/entrypoint.sh
davidebriani, line 1: The flags for this file are being changed to remove the execution permission: 755 -> 644.
Is there a specific reason why? I assumed we can leave it executable.

davidebriani (changes requested):

docker-compose.dev.yml
```diff
@@ -0,0 +1,89 @@
+services:
+  astarte-housekeeping:
+    build:
+      target: builder
+    command: >
+      set -e
+      mix run -e Elixir.Astarte.Housekeeping.ReleaseTasks.init_database
+      mix run -e Elixir.Astarte.Housekeeping.ReleaseTasks.migrate
+      mix run --no-halt
```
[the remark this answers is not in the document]
davidebriani: My bad, I misread the lines and I though they were *exactly* equal to the content of the production `entrypoint.sh`.
Then it's fine, disregard my comment.

davidebriani (approved):

Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+RUN mix local.hex --force && \
+    mix local.rebar --force && \
+    mix hex.info
+
+WORKDIR /src
+
+FROM base as deps
+# Cache elixir deps
+ADD mix.exs mix.lock ./
+RUN mix do deps.get, deps.compile
```
bettio, line 25: I'm not sure if it is a good practice to run the `deps.compile` step without setting an environment.

This means that we are budiling test deps while aiming production, but that might even be fine. However, does the dependencies code have different code paths according to the environment?

For the records here is what we were doing:
```dockerfile
# Pass --build-arg BUILD_ENV=dev to build a dev image
ARG BUILD_ENV=prod

ENV MIX_ENV=$BUILD_ENV
```

Here is the default Phoenix generated dockerfile:
```dockerfile
# set build ENV
ENV MIX_ENV="prod"

# install mix dependencies
COPY mix.exs mix.lock ./
RUN mix deps.get --only $MIX_ENV
```

I'm not saying that what we do is wrong, but I'm not 100% sure it is right. We might ask for feedbacks on the elixir forum (https://elixirforum.com/).

shinnokdisengir: Fixed, I've added the default `prod' and `docker-compose.dev.yml' to set `MIX_ENV=dev'.

docker-compose.dev.yml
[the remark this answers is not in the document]
shinnokdisengir: @noaccOS I found myself looking back at the docker-compose/yaml multiline talk and the situation is still different: the form of the classic multiline is what you were saying, but it has to become an element of the command array: 
```
command: - | set -e mix eval Elixir.Astarte.Housekeeping.ReleaseTasks.init_database mix eval Elixir.Astarte.Housekeeping.ReleaseTasks.migrate mix run --no-halt 
``` 
you can also use >, but you have to put everything in &&, since everything is folded in this case. My bad

rbino (approved):

Annopaolo (approved):

davidebriani (approved):

PR #957, opened 2024-06-10
feat(device_fleet_simulator): cache credentials secrets

noaccOS:
add the ability to avoid re-registering the devices each test, by storing the device credentials secret in a csv file.

this is optional and defaults to the re-registering the devices each time.


Annopaolo (changes requested):
LGTM but please add a couple lines to `README.md` to document the new env vars

Annopaolo (approved):

PR #961, opened 2024-06-21
Trigger Engine: fwport connection failure fix from release-1.1

Annopaolo:


Annopaolo:
It seems that the codecov action broke one again, yay

davidebriani (approved):

PR #969, opened 2024-07-03
AppEngine: handle type casting for `longinteger` inputs

### LINKED ISSUE #967: `longinteger` values cannot receive a string input in server-owned interfaces
`longinteger` values are exposed as strings by default by AppEngine API since they are 64 bit integers and JSON is only able to represent ~53 bit integers at most since all numbers are double.

This means that we should also accept integer values as input (since they are the only way to send `longinteger` values bigger than 53 bits with a JSON), but we currently don't.

### Sketch of the fix

Add another clause to `cast_value` [here](https://github.com/astarte-platform/astarte/blob/fb5657d06ebfc806d944cce354e261c18796960d/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex#L764) which more or less does this:

```elixir
  defp cast_value(:longinteger, string_value) when is_binary(string_value) do
    # Try to parse the value as integer, return `{:ok, int_value}` or `{:error, :unexpected_value_type}`
  end

  defp cast_value(:longintegerarray, string_values) do
    # Similar but for an array of values. Map over them and return an `{:ok, list_of_ints}` or an error tuple.
  end
```

Hibe7:
Implemented `cast_value/2` functions to support string input parsing for `longinteger` values and arrays in server-owned interfaces. This implementation enables parsing of string inputs into integers where necessary.

Fixes #967

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
@@ -766,6 +766,34 @@ defmodule Astarte.AppEngine.API.Device do
     end
   end
 
+  defp cast_value(:longinteger, string_value) when is_binary(string_value) do
+    case Integer.parse(string_value) do
+      {int_value, ""} ->
+        {:ok, int_value}
+
+      _ ->
+        {:error, :unexpected_value_type, expected: :longinteger}
+    end
+  end
+
+  defp cast_value(:longinteger, int_value) when is_integer(int_value) do
+    {:ok, int_value}
+  end
+
+  defp cast_value(:longinteger, _value) do
+    {:error, :unexpected_value_type, expected: :longinteger}
+  end
+
+  defp cast_value(:longintegerarray, values) do
+    case map_while_ok(values, &cast_value(:longinteger, &1)) do
+      {:ok, mapped_values} ->
+        {:ok, mapped_values}
+
+      _ ->
+        {:error, :unexpected_value_type, expected: :longintegerarray}
+    end
+  end
+
   defp cast_value(_anytype, anyvalue) do
     {:ok, anyvalue}
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [astarte_trigger_engine] Always treat event TTL for trigger policies in seconds,
   not milliseconds.
 - [astarte_trigger_engine] ack messages even with unreachable target (see https://github.com/astarte-platform/astarte/issues/936)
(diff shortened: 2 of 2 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -775,6 +775,29 @@ defmodule Astarte.AppEngine.API.Device do
     end
   end
 
+  defp cast_value(:longinteger, string_value) when is_binary(string_value) do
+    with {int_value, _} <- Integer.parse(string_value) do
+      {:ok, int_value}
+    else
+      :error ->
+        {:error, :unexpected_value_type, expected: :longinteger}
+    end
```
rbino, line 784: Since you have a single clause and you have to explicitly handle the `else` clause anyway, use a `case`:
```elixir
case Integer.parse(string_value) do
  # ...
end
```

Also note that you should match on `{int_value, ""}`, otherwise a value like `"12345foobar"` would pass the check (since `Integer.parse` returns `{12345, "foobar"}`)

rbino (changes requested):

Pavinati (approved):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -766,6 +766,30 @@ defmodule Astarte.AppEngine.API.Device do
...
+    case Integer.parse(string_value) do
+      {int_value, ""} ->
+        {:ok, int_value}
+
+      _ ->
+        {:error, :unexpected_value_type, expected: :longinteger}
+    end
+  end
+
+  defp cast_value(:longinteger, _string_value) do
```
rbino, line 779: You're missing a clause above this one:
```elixir
defp cast_value(:longinteger, int_value) when is_integer(int_value) do
  {:ok, int_value}
end
```
Otherwise we would stop accepting integer values, which are valid


Also, small nit: in the latest clause (the one I'm commenting on) the value is actually _not_ a string, so just call the unused variable `_value`.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -766,6 +766,30 @@ defmodule Astarte.AppEngine.API.Device do
...
+    {:error, :unexpected_value_type, expected: :longinteger}
+  end
+
+  defp cast_value(:longintegerarray, string_values) do
```
rbino, line 783: Nit: you don't know yet if values are string or something else, so just call the variable `values`

rbino (changes requested):

rbino (approved):

rbino (approved):

PR #978, opened 2024-07-17
`astarte_dev_tool`: setup and features

shinnokdisengir:
`astarte_dev_tool` first commit

The branch is based on https://github.com/astarte-platform/astarte/pull/956, which is necessary to start
`docker compose` in development mode

Addition of `astarte_dev_tool` functionalities: the functionalities are stateless
and based on the `mix task` mechanism

- [X] `mix astarte_dev_tool.system.up`: start `docker compose` in dev mode
- [x] `mix astarte_dev_tool.system.down`: stop `docker compose` in dev mode
- [X] `mix astarte_dev_tool.system.watch`: watch mode to enable hot-code-reloading


tools/astarte_dev_tool/lib/commands/system/watch.ex
```diff
@@ -0,0 +1,47 @@
...
+    with {_result, 0} <-
+           System.cmd(@command, @command_watch_args, @options ++ [cd: path]) do
+      :ok
+    else
+      {:error, reason} -> {:error, "Cannot run system watching: #{reason}"}
+      {result, exit_code} -> {:error, "Cannot exec system.watch: #{result}, #{exit_code}"}
+    end
+  end
+
+  defp kill_zombie_process(path) do
```
eddbbt, line 37: what if, instead of checking for orphan processes, we set up an internal process listener like https://gitlab.com/Nicd/mbu/-/blob/master/lib/apprunner.exs ?  This one checks for STDIN but can be adapted easily, but it's the idea behind it. A process that kill itself under certain conditions sounds a little bit cleaner than wait for the next run to kill eventual orphans

This want to be a suggestion, not a blocker for the PR, we can discuss this later if you want

tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/system/down.ex
```diff
@@ -0,0 +1,75 @@
...
+      your config files will not be used. If this is not provided, no level
+      will be set, so that if you set it yourself before calling this task
+      then this won't interfere. Can be any of the `t:Logger.level/0` levels
+  """
+
+  @impl true
+  def run(args) do
+    {opts, _} = OptionParser.parse!(args, strict: @switches, aliases: @aliases)
+
+    unless Keyword.has_key?(opts, :path), do: Mix.raise("The --path argument is required")
```
eddbbt, line 59: we can setup a default path also? when we clone the repo, everything has a already a fixed path

tools/astarte_dev_tool/README.md
```diff
@@ -0,0 +1,62 @@
...
+      Astarte dev mode up & ready
+  end note
+
+  Realm --> Interface
+  Realm --> Device
+  Realm --> Break
+  Realm --> [*]
+
+  Break:::interrupt --> [*]
+```
```
davidebriani, line 44: This diagram is not related to the work included in this PR.
Let's remove it.

tools/astarte_dev_tool/lib/application.ex
```diff
@@ -0,0 +1,32 @@
...
+
+  @impl true
+  def start(_type, args) do
+    {_opts, _, _} =
+      OptionParser.parse(args, switches: [verbose: :boolean], aliases: [v: :verbose])
+
+    # TODO start application
+    {:ok, self()}
+  end
+end
```
davidebriani, line 32: We don't need an application.
We can remove:
- `application.ex`
- The reference to the application in `mix.ex`
- The `config` folder.

Actually you can achieve just what is needed by running `mix new astarte_dev_tool` and using the output as the first commit of the PR.

davidebriani (changes requested):

tools/astarte_dev_tool/lib/commands/system/watch.ex
shinnokdisengir: I'm evaluating to implement it for the next PR (the improvement one)

tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/system/down.ex
```diff
@@ -0,0 +1,75 @@
...
+
+    if log_level = opts[:log_level],
+      do: Logger.configure(level: String.to_existing_atom(log_level))
+
+    with path <- opts[:path],
+         {:ok, abs_path} <- Path.normalize_path(path),
+         _ = Mix.shell().info("Stopping astarte system..."),
```
Annopaolo, line 66: Here you use `Mix.shell()` to print to console, but below (line 61-62) you are configuring Logger. It would be better to unify everything to use only one of the two. Or is there a specific reason to keep them separate?

tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/system/up.ex
```diff
@@ -0,0 +1,72 @@
...
+    {opts, _} = OptionParser.parse!(args, strict: @switches, aliases: @aliases)
+
+    unless Keyword.has_key?(opts, :path), do: Mix.raise("The --path argument is required")
+
+    if log_level = opts[:log_level],
+      do: Logger.configure(level: String.to_existing_atom(log_level))
+
+    with path <- opts[:path],
+         {:ok, abs_path} <- Path.normalize_path(path),
+         _ = Mix.shell().info("Starting system in dev mode..."),
```
Annopaolo, line 63: See the comment to `tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/system/down.ex`

tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/system/down.ex
shinnokdisengir: ok. In the prev setup there was a different repo, so the default path was not possible

shinnokdisengir: @eddbbt I'm thinking about this point..and I'm not sure default path is a good approach. Perhaps it would be better to implement inside the stateful application, keeping Tasks more agnostic as possible

tools/astarte_dev_tool/lib/application.ex
shinnokdisengir: @davidebriani the config folder has `logger` configuration..are u sure?

Annopaolo (changes requested):
I suggest using a different commit structure:
- First, a "new" commit, containing the command used to generate the folder, and the generated files
- If needed, a "cleanup" commit, very briefly stating what's been removed and why
- A commit to add the different functionalities, explaining with some more detail what's the goal of these

tools/astarte_dev_tool/lib/commands/system/up.ex
```diff
@@ -0,0 +1,32 @@
...
+  use AstarteDevTool.Constants.System
+
+  def exec(path) do
+    with {_result, 0} <-
+           System.cmd(@command, @command_up_args, @options ++ [cd: path]) do
+      :ok
+    else
+      {:process, _} -> {:error, "The system is already running"}
+      {result, exit_code} -> {:error, "Cannot exec system.up: #{result}, #{exit_code}"}
+    end
```
davidebriani, line 30: The `with` statement is useful when you want to handle a sequence of operations that constitutes the "happy path".
When you have one operation but need to handle different return values, a `case` block is more handy and readable.
Similar argument for the other commands/files.

```suggestion
    case System.cmd(@command, @command_up_args, @options ++ [cd: path]) do
      {_result, 0} -> :ok
      {:process, _} -> {:error, "The system is already running"}
      {result, exit_code} -> {:error, "Cannot exec system.up: #{result}, #{exit_code}"}
    end
```

tools/astarte_dev_tool/lib/utilities/path.ex
```diff
@@ -0,0 +1,11 @@
+defmodule AstarteDevTool.Utilities.Path do
```
davidebriani, line 1: Let's check that all files have the initial license notice.

tools/astarte_dev_tool/lib/utilities/process.ex
```diff
@@ -0,0 +1,53 @@
...
+
+  defp find_matching_pid([pid | rest], path) do
+    with {:ok, cwd_path} <- get_cwd(pid) do
+      if cwd_path == path do
+        {:ok, pid}
+      else
+        find_matching_pid(rest, path)
+      end
+    else
+      {:error, reason} -> {:error, reason}
```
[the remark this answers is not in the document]
shinnokdisengir: ops!

tools/astarte_dev_tool/lib/constants/system.ex
```diff
@@ -0,0 +1,54 @@
...
+        "compose",
+        "-f",
+        "docker-compose.yml",
+        "-f",
+        "docker-compose.dev.yml",
+        "up",
+        "--build",
+        "--watch",
+        "-d"
+      ]
```
davidebriani, line 33: The [`~w` sigil](https://hexdocs.pm/elixir/sigils.html#word-lists) might be handy here.

```suggestion
      @command_up_args ~w(compose -f docker-compose.yml -f docker-compose.dev.yml up --build --watch -d)
```

davidebriani (changes requested):

tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/system/down.ex
Annopaolo: Don't know if it helps, but for the moment we can assume that the mix tasks are being run inside the `astarte` directory (either from a subdirectory or not)

eddbbt (changes requested):

davidebriani (changes requested):

tools/astarte_dev_tool/lib/application.ex
davidebriani: `mix new astarte_dev_tool` does not generate a `config` folder, so I assume it's not needed.

tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/system/down.ex
shinnokdisengir: `Mix.shell()` is only used when launched with `mix task`, while in the final project it will not be used

tools/astarte_dev_tool/lib/utilities/process.ex
```diff
@@ -0,0 +1,53 @@
...
+    File.read_link("/proc/#{pid}/cwd")
+  end
+
+  defp get_cwd_darwin(pid) do
+    with {output, 0} <- System.cmd("lsof", ["-p", pid]),
+         [_, cwd_path] <- Regex.run(~r/\s+cwd\s+(.*?)\s+/, output) do
+      {:ok, cwd_path}
+    else
+      {_, _} -> {:error, "Error running lsof on macOS."}
+      _ -> {:error, "Unable to determine the working directory on macOS."}
```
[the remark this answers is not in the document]
shinnokdisengir: I consider the operating system check a plus, not a minus, provided that it falls within the permitted OS. Personally speaking, if I would one day prefer to work on a mac, I would like to have the code working

tools/astarte_dev_tool/lib/commands/system/watch.ex
```diff
@@ -0,0 +1,47 @@
...
+  @moduledoc false
+  use AstarteDevTool.Constants.System
+  require Logger
+  alias AstarteDevTool.Utilities.Process, as: AstarteProcess
+
+  def exec(path) do
+    kill_zombie_process(path)
+
+    with {_result, 0} <-
+           System.cmd(@command, @command_watch_args, @options ++ [cd: path]) do
```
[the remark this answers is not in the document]
shinnokdisengir: `@command_watch_args` will later be used by the stateful application, I prefer to anticipate and not chase. As for the rest of the comment, I did as you suggested

tools/astarte_dev_tool/lib/utilities/process.ex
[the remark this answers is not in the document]
eddbbt: operating system check IS a plus, it is needed in order to have different ways to check the same thing.
But, OS errors due to a missing tool is simply not the right error, we're stopping due to a missing lsof or an error inside it, it's not a macOS problem, just that

tools/astarte_dev_tool/lib/astarte_dev_tool.ex
Annopaolo, line 1: This file is unused, it can be removed

tools/astarte_dev_tool/lib/constants/system.ex
```diff
@@ -0,0 +1,31 @@
...
+  @command_up_args ~w(compose -f docker-compose.yml -f docker-compose.dev.yml up --build --watch -d)
+  @command_down_args ~w(compose -f docker-compose.yml -f docker-compose.dev.yml down)
+  @command_watch_args ~w(compose -f docker-compose.yml -f docker-compose.dev.yml watch --no-up)
+  @options [stderr_to_stdout: true, into: IO.stream(:stdio, :line)]
+
+  def command, do: @command
+  def command_up_args, do: @command_up_args
+  def command_down_args, do: @command_down_args
+  def command_watch_args, do: @command_watch_args
+  def options, do: @options
```
Annopaolo, line 30: Since `Constants.options() ++ [cd: path]` is in every call to `Constants.options()`, could this be parametrized? Or maybe add some keyword options, like
```elixir
def options(opts \\ []) do
  if Keyword.get(opts, :cd)  do
  #etc etc
  end
end

File: tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/system/down.ex [hunk 14]
```diff
@@ -0,0 +1,75 @@
...
+      do: Logger.configure(level: String.to_existing_atom(log_level))
+
+    with path <- opts[:path],
+         {:ok, abs_path} <- Path.directory_path_from(path),
+         _ = Mix.shell().info("Stopping astarte system..."),
+         :ok <- Down.exec(abs_path, opts[:volumes]) do
+      Mix.shell().info("Astarte's system stopped successfully.")
+      :ok
+    else
+      {:error, output} ->
```
Annopaolo, line 71: Nit: usually the second part of the error tuple is called `reason`:
```suggestion
      {:error, reason} ->
```

tools/astarte_dev_tool/lib/utilities/process.ex
```diff
@@ -0,0 +1,40 @@
+defmodule AstarteDevTool.Utilities.Process do
+  def check_process(command, args, path) when is_list(args) do
+    check_process(command <> " " <> Enum.join(args, " "), path)
```
Annopaolo, line 3: The call is somewhat hard to parse. I suggest
```suggestion
    command = "#{command} #{Enum.join(args, "")}"
    check_process(command, path)
```

Annopaolo (changes requested):
We're almost there!

tools/astarte_dev_tool/lib/constants/system.ex
shinnokdisengir: I'd rather not: in this case `options` is always used with functionality that requires the `path` but it won't necessarily be so. It's related to _any_ bash command

tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/system/down.ex
```diff
@@ -0,0 +1,75 @@
...
+
+  ## Command line options
+    * `-p` `--path` - (required) working Astarte project directory
+
+    * `--log-level` - the level to set for `Logger`. This task
+      does not start your application, so whatever level you have configured in
+      your config files will not be used. If this is not provided, no level
+      will be set, so that if you set it yourself before calling this task
+      then this won't interfere. Can be any of the `t:Logger.level/0` levels
+  """
```
davidebriani, line 53: We can document the `-v / --volumes` option too.

tools/astarte_dev_tool/README.md
```diff
@@ -0,0 +1,60 @@
...
+def deps do
+  [
+    {:astarte_dev_tool, "~> 0.1.0"}
+  ]
+end
+```
+
+## Usage
+
+Actually, the functionalities of `astarte_dev_tool` are stateless and based on the `mix task` mechanism.
```
eddbbt, line 18:
```suggestion
Currently, the functionalities of `astarte_dev_tool` are stateless and based on the `mix task` mechanism.
```

davidebriani (changes requested):

tools/astarte_dev_tool/lib/astarte_dev_tool.ex
eddbbt, line 1: leftover?

tools/astarte_dev_tool/lib/constants/system.ex
Annopaolo: Using keyword options would allow to write both these calls:
- `Constants.options(cd: path)`
- `Constant.options()`
And also allow for extensibility of the options list

tools/astarte_dev_tool/lib/utilities/process.ex
```diff
@@ -0,0 +1,41 @@
+defmodule AstarteDevTool.Utilities.Process do
```
Annopaolo, line 1: Missing copyright header

tools/astarte_dev_tool/mix.exs
```diff
@@ -0,0 +1,25 @@
+defmodule AstarteDevTool.MixProject do
```
Annopaolo, line 1: Missing copyright header

tools/astarte_dev_tool/test/astarte_dev_tool_test.exs
```diff
@@ -0,0 +1,8 @@
+defmodule AstarteDevToolTest do
```
Annopaolo, line 1: Missing copyright header (this file could also be removed, your call)

Annopaolo (changes requested):

tools/astarte_dev_tool/lib/constants/system.ex
eddbbt: Sorry, but I don't get it, if we parameterize  eventual extra options to the Constants.options call, we get something like
```
System.cmd(Constants.command(), args, Constants.options( [cd: path])) 
```

instead of 

```
System.cmd(Constants.command(), args, Constants.options() ++ [cd: path])
```

Imho it will end up cleaner, and every optional params can still be used

shinnokdisengir: I really don't like to assume that constants are functions, so I would have preferred a metaprogramming approach (https://medium.com/@chakravartyaashish/creating-constant-modules-in-elixir-497590a5283f). Tied to that, I want to keep the constants as agnostic as possible

shinnokdisengir: I've also read this https://dev.to/olddutchcap/two-kinds-of-elixir-constants-5gad

davidebriani (approved):

tools/astarte_dev_tool/lib/constants/system.ex
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule AstarteDevTool.Constants.System do
+  @command "docker"
```
rbino, line 20: This was recently added to the Elixir docs, tl;dr: if you don't have some very specific needs you should directly make your functions return the constant instead of assigning that to a module attribute and then returning the module attribute
https://hexdocs.pm/elixir/module-attributes.html#as-compile-time-constants

tools/astarte_dev_tool/lib/constants/system.ex
rbino: I'd agree with @shinnokdisengir here, the options passed in `@options` (which I would call `base_opts` since it better conveys the idead) are universal and I'd expect to use them in every command, while the fact that you have to `cd` is very specific to this set of commands. Passing `cd` to the function with a single data point seems like premature refactoring, so I'd stay with `Constants.base_opts() ++ [cd: path]` for the time being.

tools/astarte_dev_tool/lib/constants/system.ex
shinnokdisengir: I've just discussed with @eddbbt about it

tools/astarte_dev_tool/lib/constants/system.ex
Annopaolo: @rbino I'm not sure of that as keyword lists would allow to add a number of other options in the future, too. Let's keep the option (pun intented) in mind when developing next and move on, however.

Annopaolo (approved):

PR #992, opened 2024-09-11
`astarte-dev-tool` Added generating private/public keys feature

shinnokdisengir:

This PR introduces the feature for generating public and private keys using the `mix astarte_dev_tool.auth.keys` task. This task allows the creation of both private and public keys, as well as the option to derive a public key from a provided private key using `arg`.

The feature enhances key management within the development tooling, enabling easy and streamlined keypair generation for authentication and security purposes.

#### Examples
```bash
 $ mix astarte_dev_tool.auth.key
 $ mix astarte_dev_tool.auth.key "-----BEGIN EC PRIVATE KEY-----Base64-----END EC PRIVATE KEY-----"
 $ mix astarte_dev_tool.auth.key "$(mix astarte_dev_tool.auth.key)"
```

Nothing

The feature includes two main functionalities:
- Generating a new keypair (both private and public keys).
- Deriving a public key from a given private key using the `--from-key` flag.

### DIFF
--- a/tools/astarte_dev_tool/lib/commands/auth/keys.ex
+++ b/tools/astarte_dev_tool/lib/commands/auth/keys.ex
@@ -1,0 +1,26 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule AstarteDevTool.Commands.Auth.Keys do
+  @moduledoc false
+  alias AstarteDevTool.Utilities.Auth
+
+  def exec(), do: Auth.new_ec_private_key()
+  def exec(nil), do: exec()
+  def exec(priv), do: Auth.public_key_from(priv)
+end
--- a/tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/auth/keys.ex
+++ b/tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/auth/keys.ex
@@ -1,0 +1,70 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Mix.Tasks.AstarteDevTool.Auth.Keys do
+  use Mix.Task
+  alias AstarteDevTool.Commands.Auth.Keys
+
+  @shortdoc "Astarte keys generation"
+
+  @aliases []
+
+  @switches [
+    log_level: :string
+  ]
+
+  @moduledoc """
+  Astarte auth key generation.
+  If no `arg` is provided, a private key is released
+  If a private key is provided via `arg`, the linked public key is released.
+
+  ## Examples
+
+      $ mix astarte_dev_tool.auth.key
+      $ mix astarte_dev_tool.auth.key "-----BEGIN EC PRIVATE KEY-----Base64-----END EC PRIVATE KEY-----"
+      $ mix astarte_dev_tool.auth.key "$(mix astarte_dev_tool.auth.key)"
+
+  ## Command line options
+    * `--log-level` - the level to set for `Logger`. This task
+      does not start your application, so whatever level you have configured in
+      your config files will not be used. If this is not provided, no level
+      will be set, so that if you set it yourself before calling this task
+      then this won't interfere. Can be any of the `t:Logger.level/0` levels
+  """
+
+  @impl true
+  def run(args) do
+    {opts, args} = OptionParser.parse!(args, strict: @switches, aliases: @aliases)
+
+    if Enum.count(args) > 1,
+      do: Mix.raise("Only one optional argument - the private key - is allowed")
+
+    if log_level = opts[:log_level],
+      do: Logger.configure(level: String.to_existing_atom(log_level))
+
+    case Keys.exec(Enum.at(args, 0)) do
+      {:ok, key} ->
+        # Result to stdout
+        IO.puts(key)
+        {:ok, key}
+
+      {:error, reason} ->
+        Mix.raise("Error generating key: #{reason}")
+    end
+  end
+end
--- a/tools/astarte_dev_tool/lib/utilities/auth.ex
+++ b/tools/astarte_dev_tool/lib/utilities/auth.ex
@@ -18,8 +18,29 @@
 
 defmodule AstarteDevTool.Utilities.Auth do
   alias Astarte.Client.Credentials
+  alias X509.PrivateKey
+  alias X509.PublicKey
 
   def gen_auth_token(private_key) when is_bitstring(private_key) do
     Credentials.dashboard_credentials() |> Credentials.to_jwt(private_key)
+  end
+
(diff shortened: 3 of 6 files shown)


tools/astarte_dev_tool/test/astarte_dev_tool_test.exs
```diff
@@ -1,8 +1,41 @@
...
+
 defmodule AstarteDevToolTest do
   use ExUnit.Case
-  doctest AstarteDevTool
+  alias AstarteDevTool.Utilities.Auth
+
+  # doctest AstarteDevTool
+  describe "unit test" do
+    test "key" do
+      assert {:ok, private} = Auth.pem_key()
```
eddbbt, line 26: AstarteDevTool.Utilities.Auth.pem_keys/0 and AstarteDevTool.Utilities.Auth.pem_keys/1 functions don't exist yes

tools/astarte_dev_tool/lib/commands/auth/keys.ex
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule AstarteDevTool.Commands.Auth.Keys do
+  @moduledoc false
+  alias AstarteDevTool.Utilities.Auth, as: AuthUtilities
```
eddbbt, line 21: I will not define an alias like this 
```
alias AstarteDevTool.Utilities.Auth, as: AuthUtilities
``` 
then below, alias the same module just as 
```
alias AstarteDevTool.Utilities.Auth
```

Since the only occurrence of Auth in this file is within the Utilities group, we can just alias as Auth
```suggestion
  alias AstarteDevTool.Utilities.Auth 
```

eddbbt (commented):
remember to sign before pushing

tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/auth/keys.ex
```diff
@@ -0,0 +1,69 @@
...
+
+    if Enum.count(args) > 1,
+      do: Mix.raise("Only one optional argument - the private key - is allowed")
+
+    if log_level = opts[:log_level],
+      do: Logger.configure(level: String.to_existing_atom(log_level))
+
+    case Keys.exec(Enum.at(args, 0)) do
+      {:ok, key} ->
+        IO.puts(key)
```
eddbbt, line 62: leftover

tools/astarte_dev_tool/lib/utilities/auth.ex
```diff
@@ -18,8 +18,26 @@
...
+  def pem_key(), do: {:ok, PrivateKey.new_ec(:secp256r1) |> PrivateKey.to_pem()}
+
+  def pem_key(priv) do
+    case PrivateKey.from_pem(priv) do
+      {:ok, result} -> {:ok, result |> PublicKey.derive() |> PublicKey.to_pem()}
+      error -> error
+    end
+  end
+
+  def pem_keys() do
```
eddbbt, line 37: looks like we're using this only for testing matter, so it's not really needed

tools/astarte_dev_tool/test/astarte_dev_tool_test.exs
```diff
@@ -1,8 +1,41 @@
...
+      assert {:ok, public} = Auth.pem_key(private)
+    end
+
+    test "pem_keys/0" do
```
eddbbt, line 30: as above, i'll not test against another function, stick with the ones that already exists (above test is enough, just change the name)

tools/astarte_dev_tool/lib/utilities/auth.ex
shinnokdisengir: nope, it will be used by the `astarte_dev_tool` stateful version

tools/astarte_dev_tool/lib/utilities/auth.ex
```diff
@@ -18,8 +18,29 @@
...
   end
+
+  def pem_key(), do: {:ok, PrivateKey.new_ec(:secp256r1) |> PrivateKey.to_pem()}
+
+  def pem_key(priv) do
+    case PrivateKey.from_pem(priv) do
+      {:ok, result} -> {:ok, result |> PublicKey.derive() |> PublicKey.to_pem()}
+      error -> error
+    end
+  end
```
Annopaolo, line 35: I'm not a fan of using the same approach as the CLI here. It is not clear what the `pem_key` function does. One option could be to use an option (pun intended) to pass in the private key, something like
```elixir
def pem_key(opts \\ []) do
case Keyword.fetch(:private_key) do
#...
```
but in my opinion it would be clearer to have two different functions, one for generating the private key and another for the public key.

tools/astarte_dev_tool/lib/commands/auth/keys.ex
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+#
+
+defmodule AstarteDevTool.Commands.Auth.Keys do
+  @moduledoc false
+  alias AstarteDevTool.Utilities.Auth
+
+  def exec(), do: Auth.pem_key()
+  def exec(nil), do: exec()
+  def exec(priv), do: Auth.pem_key(priv)
```
Annopaolo, line 25: See the comment to `tools/astarte_dev_tool/lib/utilities/auth.ex`

tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/auth/keys.ex
```diff
@@ -0,0 +1,70 @@
...
+
+  @moduledoc """
+  Astarte auth key generation.
+  If no `arg` is provided, a private key is released
+  If a private key is provided via `arg`, the linked public key is released.
+
+  ## Examples
+
+      $ mix astarte_dev_tool.auth.key
+      $ mix astarte_dev_tool.auth.key "-----BEGIN EC PRIVATE KEY-----Base64-----END EC PRIVATE KEY-----"
```
Annopaolo, line 39: The PR description mentions a `--from-key` that seems to be missing. Leftover?

Annopaolo (changes requested):

tools/astarte_dev_tool/lib/utilities/auth.ex
shinnokdisengir: I prefer the ‘erlang’ approach. lots of simple functions

tools/astarte_dev_tool/lib/commands/auth/keys.ex
shinnokdisengir: see the answer

tools/astarte_dev_tool/lib/mix/tasks/astarte_dev_tool/auth/keys.ex
shinnokdisengir: Ops! I've fixed PR description

tools/astarte_dev_tool/lib/utilities/auth.ex
eddbbt: imho, pem_key could make sense as a single function that does both, but a clearer invocation (so, an explicit field :private_key) surely help to better understand without necessarily checking the code

Like what 
```
OptionParser.parse!(args, strict: @switches, aliases: @aliases)
```
does with `strict:` and `aliases:`

Annopaolo: There is no way for a reader to understand what `pem_key` does without reading the code.

The `X509` library handles [private](https://hexdocs.pm/x509/X509.PrivateKey.html) and [public](https://hexdocs.pm/x509/X509.PublicKey.html) keys in two different modules (modules! not even functions); the same approach is used by Erlang's [`public_key`](https://www.erlang.org/doc/apps/public_key/public_key.html) library, while the [`:crypto`](https://www.erlang.org/doc/apps/crypto/crypto#key-api) library outputs both the private and public key in just one functions, and documents the fact that you can pass a private key as an optional argument using typespecs.

The current approach is suitable for a CLI command, but too opaque for implementation code. In my opinion, the way to go is using two functions, e.g. `new_ec_private_key/0` and `public_key_from/1`, but a number of different options is available.

shinnokdisengir: Although in my opinion the single function was the most elegant solution, it is fine. I used the `atom' approach to solve

Annopaolo: I am sorry, but this is still not a clear solution. In the same way as before, there is no way for a reader to understand what the function does without looking at implementation details (the atom, in this case).
To move out from this impasse, I suggest either changing function names (since the two functions do two different things) or generating both keys in the same function (the `:crypto` approach)

Annopaolo (commented):
LGTM, it just needs a rebase

Annopaolo (approved):

PR #995, opened 2024-09-23
DUP: move AMQP handling logic to Mississippi

Annopaolo:

Use the Mississippi framework to manage messages coming from VerneMQ. This allows to retain the same ordering and delivery guarantees, while moving to the external component all the intricacies of handling messages in a distributed setting.
This will also increase the ease of contributing to the core message handling logic, as now code only works on what to do once a message is arrived, instead of also including how to receive a message.

Note that e2e tests will fail until https://github.com/astarte-platform/astarte_vmq_plugin/pull/94 is merged.

(none)

* [ ] Yes
* [x] No

```docs

```


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/volatile_trigger_handler.ex
```diff
@@ -0,0 +1,98 @@
...
+  def install_volatile_trigger(%InstallVolatileTrigger{
+        realm_name: realm_name,
+        device_id: encoded_device_id,
+        object_id: object_id,
+        object_type: object_type,
+        parent_id: parent_id,
+        simple_trigger_id: simple_trigger_id,
+        simple_trigger: simple_trigger,
+        trigger_target: trigger_target
+      }) do
```
noaccOS, line 37:
```suggestion
  def install_volatile_trigger(install_volatile_trigger) do
    %InstallVolatileTrigger{
        realm_name: realm_name,
        device_id: encoded_device_id,
        object_id: object_id,
        object_type: object_type,
        parent_id: parent_id,
        simple_trigger_id: simple_trigger_id,
        simple_trigger: simple_trigger,
        trigger_target: trigger_target
      } = install_volatile_trigger
```

I'd suggest keeping the pattern matching complexity out of the function definition when possible (ie when we're not pattern matching on the function arguments)

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/volatile_trigger_handler.ex
```diff
@@ -0,0 +1,98 @@
...
+    with :ok <- verify_device_exists(realm_name, encoded_device_id),
+         {:ok, device_id} <- decode_device_id(encoded_device_id) do
+      sharding_key = {realm_name, device_id}
+
+      {:ok, du_pid} = DataUpdater.get_data_updater_process(sharding_key)
+
+      DataUpdater.handle_signal(
+        du_pid,
+        {:handle_install_volatile_trigger, object_id, object_type, parent_id, simple_trigger_id,
+         simple_trigger, trigger_target}
```
noaccOS, line 47: why not passing down the whole object? as a rule of thumb, when you have more than 4-5 distinct objects you're usually better off aggregating them in a single collection (keyword, struct, etc)

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -87,15 +97,78 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
...
+      _ ->
+        # ack to discard unhandled messages
+        {:ack, :ok, state}
+    end
+  end
+
+  @impl Handler
+  def handle_signal(signal, state) do
+    case signal do
+      {:handle_install_volatile_trigger, object_id, object_type, parent_id, simple_trigger_id,
```
noaccOS, line 154: this is the only `handle_` function where we pattern match on the argument inside the function, I'd keep this at function signature level for consistency but don't feel strongly about it

noaccOS (commented):
I see you have split the PR in many commits, but I think that
1. They don't really work by themselves because they remove functionality in order to implement them one at a time
2. They don't make the review process any easier because the commit diffs are gigantic with all the comment - uncomment procedure

Still, aside from a couple comments, the code looks good

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/application.ex
```diff
@@ -63,15 +64,16 @@ defmodule Astarte.DataUpdaterPlant.Application do
...
       amqp_consumer_options: [host: Config.amqp_consumer_host!()],
-      events_consumer_connection_number: Config.amqp_consumer_connection_number!(),
-      mississippi_queues_config: [
-        events_exchange_name: Config.events_exchange_name!(),
-        data_queue_prefix: Config.data_queue_prefix!(),
-        data_queue_range_start: Config.data_queue_range_start!(),
-        data_queue_range_end: Config.data_queue_range_end!(),
-        data_queue_total_count: Config.data_queue_total_count!()
-      ],
-      message_handler: Impl
```
noaccOS: Should this change be in 209fa627?

Annopaolo: Makes sense

noaccOS (approved):

rbino (approved):

PR #998, opened 2024-10-03
added cassandra port binding

shinnokdisengir:
In order to local develop astarte via `dev mode`, cassandra/scylla needs to expose the port used to connection

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -114,3 +114,6 @@ services:
       watch:
         - path: apps/astarte_trigger_engine/lib
           action: rebuild
+  scylla:
+    ports:
+      - "9042:9042"


docker-compose.dev.yml
```diff
@@ -114,3 +114,6 @@ services:
       watch:
         - path: apps/astarte_trigger_engine/lib
           action: rebuild
+  scylla:
+    ports:
+      - "9042:9042"
```
Annopaolo, line 119: Watch out for newlines!

Annopaolo (changes requested):
Some problems with DCO

Annopaolo (changes requested):
Code is good! Please, provide some more context in the commit (and PR title) and we're ready to go!

PR #999, opened 2024-10-03
Feature/housekeeping dataaccess

shinnokdisengir:
After porting all `realm` operations into `astarte_dataaccess`, `housekeeping` needs to be refactored using the newest api


docker-compose.dev.yml
Annopaolo, line 1: This change is not related, please remove from this PR

apps/astarte_housekeeping/mix.exs
```diff
@@ -72,7 +72,7 @@ defmodule Astarte.Housekeeping.Mixfile do
   defp astarte_required_modules(_) do
     [
       {:astarte_core, github: "astarte-platform/astarte_core"},
-      {:astarte_data_access, github: "astarte-platform/astarte_data_access"},
+      {:astarte_data_access, path: "/home/ghio/Documents/astarte_data_access"},
```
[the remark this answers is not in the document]
shinnokdisengir: sorry!

apps/astarte_housekeeping/lib/astarte_housekeeping/release_tasks.ex
```diff
@@ -84,7 +85,7 @@ defmodule Astarte.Housekeeping.ReleaseTasks do
   end
 
   defp wait_connection_and_check_astarte_keyspace(retries \\ 60) do
-    case Queries.is_astarte_keyspace_existing() do
+    case Keyspace.keyspace_existing?("astarte") do
```
eddbbt, line 88: could be shortened as 

```suggestion
    with {:error, :database_connection_error} <- Keyspace.keyspace_existing?("astarte") do
      if retries > 0 do
          :timer.sleep(1000)
          wait_connection_and_check_astarte_keyspace(retries - 1)
        else
          {:error, :database_connection_error}
        end
    
    end
```

cause when 'with' can not match, it returns the value (so {:ok, _} or {:error, _})

PR #1005, opened 2024-10-17
Fix unsupported data types in Astarte import

osmanhadzic:
The current import functionality fails when encountering unsupported data types, which can lead to incomplete or incorrect imports. This fix ensures that unsupported data types are handled gracefully.

- Added checks for unsupported data types in the import process.
- Implemented error handling to log and skip unsupported data types instead of failing the entire import.

### Testing
Manual testing has been conducted to ensure the import process completes successfully even when unsupported data types are encountered.

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md) and [CODE_OF_CONDUCT.md](../CODE_OF_CONDUCT.md)
- [x] I have added to [CHANGELOG.md](../CHANGELOG.md) relevant changes and any other user facing change
- [x] I have added or ran the appropriate tests

This PR addresses an issue with the unsupported data types during the import process in Astarte.

Fixes #

- [x] Yes
- [ ] No

```docs

```

### DIFF
--- a/tools/astarte_import/README.md
+++ b/tools/astarte_import/README.md
@@ -8,48 +8,400 @@ docker run -e CASSANDRA_DB_HOST=127.0.0.1 -e CASSANDRA_DB_PORT=9042 \
  --net=host astarte/astarte_import
 ```
 
-The following example is a valid Astarte Import XML file:
+Command to run data import:
+
+```bash
+mix astarte.import <realm> <xml file>
+```
+
 ```xml
-<?xml version="1.0" encoding="UTF-8"?>
 <astarte>
-  <devices>
-    <device device_id="yKA3CMd07kWaDyj6aMP4Dg">
-      <protocol revision="0" pending_empty_cache="false" />
-      <registration secret_bcrypt_hash="$2b$12$bKly9EEKmxfVyDeXjXu1vOebWgr34C8r4IHd9Cd.34Ozm0TWVo1Ve" first_registration="2019-05-30T13:49:57.045000Z" />
-      <credentials inhibit_request="false" cert_serial="324725654494785828109237459525026742139358888604" cert_aki="a8eaf08a797f0b10bb9e7b5dca027ec2571c5ea6" first_credentials_request="2019-05-30T13:49:57.355000Z" last_credentials_request_ip="198.51.100.1" />
-      <stats total_received_msgs="64" total_received_bytes="3960" last_connection="2019-05-30T13:49:57.561000Z" last_disconnection="2019-05-30T13:51:00.038000Z" last_seen_ip="198.51.100.89" />
-      <interfaces>
-        <interface name="org.astarteplatform.Values" major_version="0" minor_version="1" active="true">
-          <datastream path="/realValue">
-            <value reception_timestamp="2019-05-31T09:12:42.789379Z">0.1</value>
-            <value reception_timestamp="2019-05-31T09:13:29.144111Z">0.2</value>
-            <value reception_timestamp="2019-05-31T09:13:52.040373Z">0.3</value>
+<devices>
+<device device_id="yKA3CMd07kWaDyj6aMP4Dg">
+  <protocol pending_empty_cache="false" revision="0"></protocol>
+  <registration first_registration="2024-05-30T13:49:57.045Z" secret_bcrypt_hash="$2b$12$bKly9EEKmxfVyDeXjXu1vOebWgr34C8r4IHd9Cd.34Ozm0TWVo1Ve"></registration>
+  <credentials cert_aki="a8eaf08a797f0b10bb9e7b5dca027ec2571c5ea6" cert_serial="324725654494785828109237459525026742139358888604" first_credentials_request="2024-05-30T13:49:57.355Z" inhibit_request="false"></credentials>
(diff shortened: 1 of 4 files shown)


tools/astarte_import/README.md
```diff
@@ -8,48 +8,334 @@ docker run -e CASSANDRA_DB_HOST=127.0.0.1 -e CASSANDRA_DB_PORT=9042 \
...
+        <value reception_timestamp="2024-09-09T09:00:00.000Z">example string</value>
+      </datastream>
+      <datastream path="/a/binaryblob">
+        <value reception_timestamp="2024-09-09T09:00:00.000Z">aGVsbG8gd29ybGQ=</value>
+      </datastream>
+      <datastream path="/a/datetime">
+        <value reception_timestamp="2024-09-09T09:00:00.000Z">2024-10-17T13:25:19.130Z</value>
+      </datastream>
+      <datastream path="/a/doublearray">
+        <value reception_timestamp="2024-09-09T09:00:00.000Z">[123.45, 678.90]</value>
```
Annopaolo, line 43: Here (and in the following lines regarding arrays) XML and JSON are mixed. While this is technically valid, an inconsistency in the representation of data is introduced. I would suggest to use this representation instead, encoding arrays as lists of XML elements:
```suggestion
        <value reception_timestamp="2024-09-09T09:00:00.000Z">
          <element>123.45</element>
          <element>678.9</element>
        </value>
```

tools/astarte_import/lib/astarte/import/populatedb.ex
```diff
@@ -491,6 +491,83 @@ defmodule Astarte.Import.PopulateDB do
     end
   end
 
+  defp to_native_type(value_chars, :doublearray) do
+    with doublearray_string = to_string(value_chars),
+         {:ok, doublearray_list} <- Jason.decode(doublearray_string),
```
Annopaolo, line 496: We should not mix XML and JSON (see previous comment).
I'd suggest to take a look e.g. at how object `item`s are handled:
- When you start parsing the `item` element, you use the `:startElement` event to build the actual object item in memory, see [here](https://github.com/astarte-platform/astarte/blob/fe6af4f4d07bf3c663c7964823b76d4c0a99b0cd/tools/astarte_import/lib/astarte/import.ex#L260)
- When you're done with parsing the item, you can put it inside an accumulator, see [here](https://github.com/astarte-platform/astarte/blob/fe6af4f4d07bf3c663c7964823b76d4c0a99b0cd/tools/astarte_import/lib/astarte/import.ex#L291)
- When the full list (i.e. the actual value) is built, insert it in the database using a function (see [here](https://github.com/astarte-platform/astarte/blob/fe6af4f4d07bf3c663c7964823b76d4c0a99b0cd/tools/astarte_import/lib/astarte/import/populatedb.ex#L334). In this case, I think it would be actually possible not to write a new function, but just reuse [`got_end_of_value_fun`](https://github.com/astarte-platform/astarte/blob/fe6af4f4d07bf3c663c7964823b76d4c0a99b0cd/tools/astarte_import/lib/astarte/import/populatedb.ex#L304-L332).

Note that this approach would also allow for recycling the already existing `to_native_type` functions (as every element in the list would be parsed individually)

Annopaolo (changes requested):

tools/astarte_import/lib/astarte/import/populatedb.ex
Annopaolo: Edit: `got_end_of_value_fun` would probably need some little adjustment to handle casting to the correct type

Annopaolo: Edit 2: the values are casted AFTER being parsed, just before being inserted in the database. So the `to_native_type` function for arrays would probably become just something like `Enum.map(array_value, fn element, type -> to_native_type(element, type) end)`

tools/astarte_import/README.md
osmanhadzic: Should we make the same changes in the Property and Object interfaces? For example:
``` xml
<property path="/a/doublearray" reception_timestamp="2024-09-04T11:31:36.956Z">
       <element>123.45</element>
       <element>678.90</element>
</property>
```
``` xml
<item name="/doublearray">
      <element>123.45</element>
      <element>678.90</element>
</item>
```

Annopaolo: Yes @osmanhadzic , good catch!

tools/astarte_import/lib/astarte/import/populatedb.ex
```diff
@@ -487,6 +487,62 @@ defmodule Astarte.Import.PopulateDB do
     end
   end
 
+  defp to_native_type(value_chars, :doublearray) do
+    {:ok,
+     Enum.map(value_chars, fn element ->
+       {:ok, value} = to_native_type(element, :double)
+       value
+     end)}
```
Annopaolo, line 495: Nit: this is a bit difficult to read as the tuple adds an unnecessary additional level of indentation. Consider using
```suggestion
    value =  Enum.map(value_chars, fn element ->
       {:ok, value} = to_native_type(element, :double)
       value
     end)
   {:ok, value}
```

It applies also to the other functions below

Annopaolo (changes requested):

Annopaolo (approved):

PR #1050, opened 2025-01-21
AstarteExport: Add Support for Dynamic Cassandra Host/Port in Astarte Export

Jusuf95:
Enhancements to Astarte Export Functionality:

- Modified export_realm_data/3 to accept db_host_and_port as an argument, allowing for more flexible configuration of 
  Cassandra.
- Set Cassandra host and port as environment variables for better deployment flexibility.
- Updated the run/1 function to accommodate the new parameters for exporting data.
- Enhanced documentation to include the new db_host_and_port argument and its usage in the function.
- Updated the README to reflect these changes and provide clear instructions on the new configuration.

Code Maintenance:
- Refined the export process by including necessary parameters for Cassandra connectivity, ensuring a smoother setup.

This change enhances the Astarte export process by allowing more flexible configuration of Cassandra connectivity, improving the function's ability to work with various environments and setups.

Fixes #

Please review the changes in export/export_realm_data/3 and run/1 to ensure the new parameters are correctly implemented and the documentation is accurate.
* [ ] Yes
* [ ] No

```docs

```


tools/astarte_export/lib/astarte/export.ex
```diff
@@ -29,18 +29,25 @@ defmodule Astarte.Export do
...
-  @spec export_realm_data(String.t(), String.t(), keyword()) ::
+  @spec export_realm_data(String.t(), String.t(), String.t(), keyword()) ::
           :ok | {:error, :invalid_parameters} | {:error, any()}
 
-  def export_realm_data(realm, file, opts \\ []) do
+  def export_realm_data(db_host_and_port, realm, file, opts \\ []) do
+    [db_host, db_port] = String.split(db_host_and_port, ":")
+
+    System.put_env("CASSANDRA_DB_HOST", db_host)
+    System.put_env("CASSANDRA_DB_PORT", db_port)
```
Annopaolo, line 49: We shouldn't rely on setting env vars from the code. As an example, what if a user has set `CASSANDRA_DB_HOST` already for another task? This would override it.
Relevant variables should be passed down in the code, as we do e.g. for the `file`

Annopaolo (changes requested):

tools/astarte_export/README.md
eddbbt, line 1: Granting the user to use this tool with another instance other than the one set-upped within the system is a good idea, but IMHO we still want to have a backup option and ensure retro-compatibility (this way we do not break existent workflows).
To do so, cassandra host:port should be optional (even better if as an opts keyword list) and in case there is no value set, it should default to CASSANDRA_NODES env var

tools/astarte_export/lib/astarte/export.ex
```diff
@@ -29,32 +29,35 @@ defmodule Astarte.Export do
   """
 
   @doc """
-    The export_realm_data/2 function required 2 arguments to export
+    The export_realm_data/2 function required 3 arguments to export
```
eddbbt, line 32: rememeber to change the arity of export_realm_data even in doc sections

tools/astarte_export/lib/astarte/fetchdata/fetchdata.ex
```diff
@@ -13,8 +13,8 @@ defmodule Astarte.Export.FetchData do
     booleanarray: :boolean
   }
 
-  def db_connection_identifier() do
-    with {:ok, conn_ref} <- Queries.get_connection() do
+  def db_connection_identifier(db_host_and_port) do
+    with {:ok, conn_ref} <- Queries.get_connection(db_host_and_port) do
```
eddbbt, line 17: imho, it should be this function to split the parameter by ":"

tools/astarte_export/lib/astarte/fetchdata/queries/queries.ex
```diff
@@ -314,4 +324,8 @@ defmodule Astarte.Export.FetchData.Queries do
     {:ok, device_uuid, _} = Device.decode_extended_device_id(device_id)
     device_uuid
   end
+
+  defp split_db_host_and_port(db_host_and_port) do
+    String.split(db_host_and_port, ":")
+  end
```
eddbbt, line 330: It's not required to make another function with only one instruction, String.split is already expressive enough

tools/astarte_export/lib/astarte/fetchdata/queries/queries.ex
```diff
@@ -6,12 +6,22 @@ defmodule Astarte.Export.FetchData.Queries do
 
   require Logger
 
-  def get_connection() do
-    host = System.get_env("CASSANDRA_DB_HOST")
-    port = System.get_env("CASSANDRA_DB_PORT")
-    Logger.info("Connecting to #{inspect(host)}:#{inspect(port)} cassandra database.")
+  def get_connection(opts \\ []) do
+    db_host_and_port = Keyword.get(opts, :db_host_and_port)
```
eddbbt, line 10: this looks good, double check the README cause there you still have a string, not a keyword list

tools/astarte_export/README.md
```diff
@@ -3,7 +3,7 @@
 Astarte Export is an easy to use tool that allows to exporting all the devices and data from an existing Astarte realm to XML format.
 
 ```iex
-iex(astarte_export@127.0.0.1)6> Astarte.Export.export_realm_data("test", "test.xml")
+iex(astarte_export@127.0.0.1)6> Astarte.Export.export_realm_data("test", "test.xml", "localhost:9042" )
```
eddbbt, line 6: definig `export_realm_data(String.t(), String.t(), keyword())` implies that the third parameter should be a keyword list, not a string

tools/astarte_export/lib/astarte/export.ex
```diff
@@ -41,6 +41,7 @@ defmodule Astarte.Export do
           :ok | {:error, :invalid_parameters} | {:error, any()}
```
eddbbt, line 41: let's use a double tipespec, one with opts and one without, because it is not mandatory

tools/astarte_export/lib/mix/tasks/astarte_export.ex
```diff
@@ -19,13 +19,40 @@ defmodule Mix.Tasks.Astarte.Export do
             Logger.error("Cannot start applications: #{inspect(reason)}")
         end
 
+      [realm, file_name, db_host_and_port] ->
```
eddbbt, line 22: this pattern match will never match the below condition `[realm, file_name, device_id]` cause mix cannot understand if you are passing an address or a device id. please check if keyword lists are supported also for mix.tasks and if so, rewrite the pattern match to have a clear difference.
If not, this piece needs to be rethinked

eddbbt (changes requested):
It is not completely wrong to accept a keyword list inside the iex shell and a string parameter outside with mix.task, but be sure that we're not overriding existent behaviors

eddbbt (approved):

PR #1054, opened 2025-01-22
feature: arm64

shinnokdisengir:
Enhance arm64 Compatibility and Development Setup

## Key Changes
- Bumped the ScyllaDB container image to version 6.2 to resolve compatibility issues on `arm64`.
- `astarte` deps are now github based
- `re2` is forced >= 1.9.8 to fix arm64 build

### Minor Changes
- fixed code to build using new deps


apps/astarte_housekeeping/lib/astarte_housekeeping/c_system.ex
```diff
@@ -97,10 +97,12 @@ defmodule CSystem do
...
     case result do
       {:error, :timeout} ->
-        Xandra.Error.new(:agreement_timeout, "Schema agreement wait timeout.")
+        raise Xandra.Error, reason: :agreement_timeout, message: "Schema agreement wait timeout."
 
       {:error, :no_schema_change} ->
-        Xandra.Error.new(:no_schema_change, "Statement did not change the schema_version.")
+        raise Xandra.Error,
+          reason: :no_schema_change,
+          message: "Statement did not change the schema_version."
```
Annopaolo, line 105: `raise`s here would break the flow in some places where the function is called inside a `with` (e.g. [here](https://github.com/astarte-platform/astarte/blob/cbfd1b289cd9f5c8576f1273af595bf9732b9897/apps/astarte_housekeeping/lib/astarte_housekeeping/migrator.ex#L306)). Let's go back to use the `{:error, %Xandra.Error{}}` return value

Annopaolo (changes requested):

PR #1069, opened 2025-02-07
Fix/astarte import test

osmanhadzic:
- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md) and [CODE_OF_CONDUCT.md](../CODE_OF_CONDUCT.md)
- [ ] I have added to [CHANGELOG.md](../CHANGELOG.md) relevant changes and any other user facing change
- [x] I have added or ran the appropriate tests

Sets up an Astarte realm instance pre-populated with data (devices, interfaces, endpoints) to facilitate testing for data imports.
Drops the database after testing to maintain a clean environment.

Fixes #

- [ ] Yes
- [X] No

```docs

```


tools/astarte_import/test/test_helper.exs
```diff
@@ -1 +1,9 @@
 ExUnit.start()
+
+children = [
+  {Astarte.DataAccess, xandra_options: Astarte.DataAccess.Config.xandra_options!()}
+]
```
Annopaolo, line 5: It would be better to use a different Xandra pool just for tests, so that we're not using DataAccess' resource.
 ```suggestion
    xandra_options =
      Astarte.DataAccess.Config.xandra_options!()
      |> Keyword.put(:name, :test_xandra)

    children = [
      {Xandra.Cluster, xandra_options}
    ]
```

This would mean that all calls to `Xandra.Cluster` functions should use the new pool, see comment above for an example

File: tools/astarte_import/test/astarte/populatedb_test.exs [hunk 2]
```diff
@@ -19,63 +19,94 @@
...
+    end)
+  end
+
+  setup_all do
+    Xandra.Cluster.run(:astarte_data_access_xandra, fn conn ->
+      DatabaseTestHelper.create_test_keyspace(conn)
+    end)
+
+    on_exit(fn ->
+      Xandra.Cluster.run(:astarte_data_access_xandra, fn conn ->
```
Annopaolo, line 102:
```suggestion
      Xandra.Cluster.run(:test_xandra, fn conn ->
```

tools/astarte_import/test/astarte/populatedb_test.exs
```diff
@@ -19,63 +19,94 @@
...
+  </device>
+  </devices>
   </astarte>
   """
 
+  setup do
+    Xandra.Cluster.run(:astarte_data_access_xandra, fn conn ->
+      DatabaseTestHelper.seed_data(conn)
+    end)
+  end
```
Annopaolo, line 94: Usually `setup` blocks are below `setup_all`s, as `setup_all` runs before all tests in a case and `setup` runs before each test in a case. See https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#setup_all/2

tools/astarte_import/test/test_helper.exs
```diff
@@ -1 +1,9 @@
...
+
+Supervisor.start_link(children, strategy: :one_for_one)
```
Annopaolo, line 7: Consider whether this could be changed to a `start_supervised` call inside a `setup_all` block (see https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#start_supervised/2)

tools/astarte_import/test/support/database_test_helper.exs
```diff
@@ -0,0 +1,328 @@
...
+      end
+    )
+  end
+
+  def drop_test_keyspace(conn) do
+    Xandra.execute(conn, @drop_keyspace, [], [])
+    :timer.sleep(3000)
+
+    Xandra.execute(conn, @delete_realm_from_astarte_realms, [], [])
+    :timer.sleep(3000)
```
Annopaolo, line 326: Are `sleep` really needed here?

tools/astarte_import/test/test_helper.exs
osmanhadzic: Sorry, I tried to implement your suggestion, but in this case, it doesn’t work. I think this PR can resolve my problem with database connections. https://github.com/astarte-platform/astarte/pull/1061

PR #1077, opened 2025-02-21
DUP: introduce Ecto (deps+schemas)

Annopaolo:

Setup an Ecto Repo for queries (bring in dependencies, update Xandra,  create the basic scaffolding) and create Ecto Schemas which are are a straightforward translation from CQL.

Notable points:
-  This changes add 10 connections to the database, as the Xandra config is copied as-is for the Exandra adapter.
- Some utility `Ecto.Type`s have been introduced in order to comply with the current Astarte database schema (UUID, big and small int)
- Object-aggregated datastreams have a runtime-defined schema, so they have not been included

* [ ] Yes
* [x] No


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/bigint.ex
```diff
@@ -0,0 +1,46 @@
...
+  @max_bigint 2 ** 63 - 1
+
+  @doc false
+  @impl Ecto.Type
+  def type, do: :bigint
+
+  @impl Ecto.Type
+  @spec load(t() | any()) :: {:ok, t()} | :error
+  def load(n) when is_integer(n) and n >= @min_bigint and n <= @max_bigint, do: {:ok, n}
+  def load(_), do: :error
```
davidebriani, line 37: We should also handle the `nil` case, without returning an error.
Same thing for the SmallInt type.

davidebriani (approved):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/bigint.ex
davidebriani: Actually, Ecto's documentation reports:

> Note: nil values are always bypassed and cannot be handled by custom types.

So it's fine like this.

PR #1091, opened 2025-02-28
refactor(appengine): port Device delete queries to exandra

eddbbt:
refactor(appengine): port Device update queries to exandra
add log in case of record not found (affected rows = 0)
Clean up unused params

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
@@ -71,8 +71,7 @@ defmodule Astarte.AppEngine.API.Device do
   end
 
   def merge_device_status(realm_name, encoded_device_id, device_status_merge) do
-    with {:ok, client} <- Database.connect(realm: realm_name),
-         {:ok, device_id} <- Device.decode_device_id(encoded_device_id),
+    with {:ok, device_id} <- Device.decode_device_id(encoded_device_id),
          {:ok, device_status} <- Queries.retrieve_device_status(realm_name, device_id),
          changeset = DeviceStatus.changeset(device_status, device_status_merge),
          {:ok, updated_device_status} <- Ecto.Changeset.apply_action(changeset, :update),
@@ -80,8 +79,8 @@ defmodule Astarte.AppEngine.API.Device do
          :ok <- change_credentials_inhibited(realm_name, device_id, credentials_inhibited_change),
          aliases_change = Map.get(changeset.changes, :aliases, %{}),
          attributes_change = Map.get(changeset.changes, :attributes, %{}),
-         :ok <- update_aliases(realm_name, client, device_id, aliases_change),
-         :ok <- update_attributes(realm_name, client, device_id, attributes_change) do
+         :ok <- update_aliases(realm_name, device_id, aliases_change),
+         :ok <- update_attributes(realm_name, device_id, attributes_change) do
       # Manually merge aliases since changesets don't perform maps deep merge
       merged_aliases = merge_data(device_status.aliases, updated_device_status.aliases)
       merged_attributes = merge_data(device_status.attributes, updated_device_status.attributes)
@@ -95,36 +94,38 @@ defmodule Astarte.AppEngine.API.Device do
     end
   end
 
-  defp update_attributes(realm_name, client, device_id, attributes) do
-    Enum.reduce_while(attributes, :ok, fn
-      {"", _attribute_value}, _acc ->
-        Logger.warning("Attribute key cannot be an empty string.",
-          tag: :invalid_attribute_empty_key
-        )
+  defp update_attributes(realm_name, device_id, attributes) do
+    Enum.reduce_while(attributes, :ok, fn {attribute_key, attribute_value}, _ ->
+      case {attribute_key, attribute_value} do
+        {"", _attribute_value} ->
+          Logger.warning("Attribute key cannot be an empty string.",
+            tag: :invalid_attribute_empty_key
+          )
 
-        {:halt, {:error, :invalid_attributes}}
+          {:halt, {:error, :invalid_attributes}}
 
-      {attribute_key, nil}, _acc ->
-        case Queries.delete_attribute(realm_name, client, device_id, attribute_key) do
-          :ok ->
-            {:cont, :ok}
+        {attribute_key, nil} ->
+          case Queries.delete_attribute(realm_name, device_id, attribute_key) do
+            :ok ->
+              {:cont, :ok}
 
-          {:error, reason} ->
-            {:halt, {:error, reason}}
-        end
+            {:error, reason} ->
+              {:halt, {:error, reason}}
+          end
 
-      {attribute_key, attribute_value}, _acc ->
-        case Queries.insert_attribute(realm_name, device_id, attribute_key, attribute_value) do
-          :ok ->
-            {:cont, :ok}
+        {attribute_key, attribute_value} ->
+          case Queries.insert_attribute(realm_name, device_id, attribute_key, attribute_value) do
+            :ok ->
+              {:cont, :ok}
 
-          {:error, reason} ->
-            {:halt, {:error, reason}}
-        end
+            {:error, reason} ->
+              {:halt, {:error, reason}}
+          end
+      end
     end)
   end
 
-  defp update_aliases(realm_name, client, device_id, aliases) do
+  defp update_aliases(realm_name, device_id, aliases) do
     Enum.reduce_while(aliases, :ok, fn
       {_alias_key, ""}, _acc ->
         Logger.warning("Alias value cannot be an empty string.", tag: :invalid_alias_empty_value)
@@ -135,13 +136,13 @@ defmodule Astarte.AppEngine.API.Device do
         {:halt, {:error, :invalid_alias}}
 
       {alias_key, nil}, _acc ->
-        case Queries.delete_alias(realm_name, client, device_id, alias_key) do
+        case Queries.delete_alias(realm_name, device_id, alias_key) do
           :ok -> {:cont, :ok}
           {:error, reason} -> {:halt, {:error, reason}}
         end
 
       {alias_key, alias_value}, _acc ->
-        case Queries.insert_alias(realm_name, client, device_id, alias_key, alias_value) do
+        case Queries.insert_alias(realm_name, device_id, alias_key, alias_value) do
           :ok -> {:cont, :ok}
           {:error, reason} -> {:halt, {:error, reason}}
         end
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
@@ -33,7 +33,7 @@ defmodule Astarte.AppEngine.API.Device.Queries do
   alias Astarte.Core.InterfaceDescriptor
   alias CQEx.Query, as: DatabaseQuery
   alias CQEx.Result, as: DatabaseResult
-
+  alias Astarte.AppEngine.API.Realm
   alias Astarte.AppEngine.API.Devices.Device, as: DatabaseDevice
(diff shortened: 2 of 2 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -33,6 +33,12 @@ defmodule Astarte.AppEngine.API.Device.Queries do
   alias Astarte.Core.InterfaceDescriptor
   alias CQEx.Query, as: DatabaseQuery
   alias CQEx.Result, as: DatabaseResult
+  alias Astarte.AppEngine.API.Devices.Device, as: ApiDevice
```
davidebriani, line 36: This module is already aliased some lines below as `DatabaseDevice`

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -234,20 +240,33 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
-    DatabaseQuery.call!(db_client, unset_query)
+    q =
+      from v in storage,
+        prefix: ^keyspace_name,
+        where:
+          v.device_id == ^device_id and v.interface_id == ^interface_id and
+            v.endpoint_id == ^endpoint and v.path == ^path
 
-    :ok
+    case Repo.delete_all(q) do
```
noaccOS, line 253: delete all cannot return `nil` or an error tuple, see its signature here: https://hexdocs.pm/ecto/Ecto.Repo.html#c:delete_all/2

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -234,20 +240,33 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
+    %InterfaceDescriptor{interface_id: interface_id, storage: storage} = interface_descriptor
+    keyspace_name = Realm.keyspace_name(realm_name)
 
-    DatabaseQuery.call!(db_client, unset_query)
+    q =
+      from v in storage,
+        prefix: ^keyspace_name,
+        where:
+          v.device_id == ^device_id and v.interface_id == ^interface_id and
+            v.endpoint_id == ^endpoint and v.path == ^path
```
davidebriani, line 251:
```suggestion
    q =
      from v in storage,
        prefix: ^keyspace_name,
        where:
          v.device_id == ^device_id and v.interface_id == ^interface_id and
            v.endpoint_id == ^endpoint.endpoint_id and v.path == ^path
```

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -234,20 +240,33 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
+      {:error, reason} ->
+        _ =
+          Logger.warning(
+            "Could not unset value for  #{Device.encode_device_id(device_id)} in #{storage}}",
+            realm: "realm",
+            tag: "cant_unset"
+          )
+
+        {:error, reason}
+    end
```
davidebriani, line 269: `Repo.delete_all` always returns `{count, records}`, so the latter two cases will never match and can be removed.
On the first case instead, the previous function returned `:ok`, not `{:ok}`.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -234,20 +240,33 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
+      from v in storage,
+        prefix: ^keyspace_name,
+        where:
+          v.device_id == ^device_id and v.interface_id == ^interface_id and
+            v.endpoint_id == ^endpoint and v.path == ^path
 
-    :ok
+    case Repo.delete_all(q) do
+      {n, _} when is_number(n) ->
+        {:ok}
```
noaccOS, line 255: usually when a result tuple is returned, the elements are from: `:ok`, `{:ok, result}`,  `:error`, `{:error, reason}`
tuples with only one element are not usually used

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -574,30 +593,24 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
+          prefix: ^keyspace,
+          where: [device_id: ^device_id],
+          update: [set: [attributes: fragment("attributes - ?", ^map_new_attribute)]]
+
+      with {1, nil} <- Repo.update_all(query_delete_attributes, [], consistency: :each_quorum) do
+        :ok
+      else
         %{acc: _, msg: error_message} ->
           _ = Logger.warning("Database error: #{error_message}.", tag: "db_error")
           {:error, :database_error}
```
noaccOS, line 616: the else can be removed as this kind of result cannot be returned from `delete_all`

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -683,41 +696,28 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
-      with {:ok, _result} <- DatabaseQuery.call(client, delete_batch) do
+      sql_query_delete_in_name = Repo.to_sql(:delete_all, query_delete_in_name)
+
+      update_and_delete_batch =
+        %Exandra.Batch{queries: [sql_query_delete_alias, sql_query_delete_in_name]}
+
+      with :ok <- Exandra.execute_batch(Repo, update_and_delete_batch, consistency: :each_quorum) do
         :ok
       else
         %{acc: _, msg: error_message} ->
```
noaccOS, line 723:
```suggestion
      Exandra.execute_batch(Repo, update_and_delete_batch, consistency: :each_quorum)
```

apart from the fact that execute_batch returns [:ok or {:error, reason}](https://hexdocs.pm/exandra/0.10.2/Exandra.html#execute_batch/3) and not a map, the with can be removed as the returned value is the same

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -574,30 +593,24 @@ defmodule Astarte.AppEngine.API.Device.Queries do
     end
   end
 
-  def delete_attribute(realm_name, client, device_id, attribute_key) do
+  def delete_attribute(realm_name, _client, device_id, attribute_key) do
```
davidebriani, line 596: Can we remove `_client` parameters if they are not used anymore?

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -671,7 +684,7 @@ defmodule Astarte.AppEngine.API.Device.Queries do
     end
   end
 
-  def delete_alias(realm_name, client, device_id, alias_tag) do
+  def delete_alias(realm_name, _client, device_id, alias_tag) do
```
davidebriani, line 687: Can we remove `_client` parameters if they are not used anymore?

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
noaccOS, line 36: this was already aliased (see line 43)

noaccOS (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -574,30 +593,24 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
 
+      query_delete_attributes =
+        from ApiDevice,
+          prefix: ^keyspace,
+          where: [device_id: ^device_id],
+          update: [set: [attributes: fragment("attributes - ?", ^map_new_attribute)]]
+
+      with {1, nil} <- Repo.update_all(query_delete_attributes, [], consistency: :each_quorum) do
+        :ok
+      else
```
davidebriani, line 613: We're already verifying that the attribute exists with the first query above.
Here we can assert that the call to `Repo.update_all` must edit 1 row.
Also note that Repo.update_all always returns a `{count, rows}` tuple, so we don't need to handle other type of return values as before.

The whole with/else block can become:

```elixir
{1, nil} = Repo.update_all(query_delete_attributes, [], consistency: :each_quorum)

:ok
```

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -683,41 +696,28 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
+          where: [object_name: ^alias_value, object_type: 1]
         )
 
-      with {:ok, _result} <- DatabaseQuery.call(client, delete_batch) do
+      sql_query_delete_in_name = Repo.to_sql(:delete_all, query_delete_in_name)
+
+      update_and_delete_batch =
+        %Exandra.Batch{queries: [sql_query_delete_alias, sql_query_delete_in_name]}
+
+      with :ok <- Exandra.execute_batch(Repo, update_and_delete_batch, consistency: :each_quorum) do
```
davidebriani, line 720: `Exandra.execute_batch` should return `:ok`, or `{:error, exception}`.
If we can "let it crash" like in other API services, we can avoid handling the error case here, so the with/else block can be replaced with:

```elixir
Exandra.execute_batch(Repo, update_and_delete_batch, consistency: :each_quorum)
```

davidebriani (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
eddbbt: except that this is always passed as an id, probably it was a map a while ago but not rn. It's never checked except for an old log. I'm gonna rename it

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -234,18 +234,23 @@ defmodule Astarte.AppEngine.API.Device.Queries do
       # TODO: should we handle this situation?
     end
 
-    # TODO: :reception_timestamp_submillis is just a place holder right now
```
noaccOS, line 237: let's not delete this comment as it's still valid

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -561,38 +566,23 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
-      end
+      map_new_attribute = MapSet.new([attribute_key])
+
+      query_delete_attributes =
+        from DatabaseDevice,
+          prefix: ^keyspace,
+          where: [device_id: ^device_id],
+          update: [set: [attributes: fragment("attributes - ?", ^map_new_attribute)]]
+
+      {1, nil} = Repo.update_all(query_delete_attributes, [], consistency: :each_quorum)
```
noaccOS, line 584: the "error" handling is inconsistent: in `insert_value_into_db`, you log in case the value did not exist, while here you're asserting it should exist and crash if it didn't

noaccOS (changes requested):
just a couple things, in general it looks good

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -71,7 +71,7 @@ defmodule Astarte.AppEngine.API.Device do
   end
 
   def merge_device_status(realm_name, encoded_device_id, device_status_merge) do
-    with {:ok, client} <- Database.connect(realm: realm_name),
```
noaccOS, line 74: if we're not using the client anymore, it's safe to remove cqex's connection altogether

noaccOS (changes requested):

noaccOS (approved):
Good job

davidebriani (approved):

PR #1093, opened 2025-02-28
DUP: move UPDATE queries from CQEx to Ecto

Annopaolo:

Same old, same old: replace the unsupported CQEx driver with Ecto (specifically, to use the Exandra adapter).
This also implies some refactor, as we don't need to pass around the CQEx client anymore. 

The API of existing query functions is kept as it was.

* [ ] Yes
* [x] No

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
@@ -63,30 +63,28 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
     Repo.all(q)
   end
 
-  def set_pending_empty_cache(db_client, device_id, pending_empty_cache) do
-    pending_empty_cache_statement = """
-    UPDATE devices
-    SET pending_empty_cache = :pending_empty_cache
-    WHERE device_id = :device_id
-    """
+  def set_pending_empty_cache(realm, device_id, pending_empty_cache) do
+    keyspace_name = Realm.keyspace_name(realm)
 
-    update_pending =
-      DatabaseQuery.new()
-      |> DatabaseQuery.statement(pending_empty_cache_statement)
-      |> DatabaseQuery.put(:device_id, device_id)
-      |> DatabaseQuery.put(:pending_empty_cache, pending_empty_cache)
+    device =
+      from d in Device,
+        prefix: ^keyspace_name,
+        where: [device_id: ^device_id],
+        update: [set: [pending_empty_cache: ^pending_empty_cache]]
 
-    with {:ok, _result} <- DatabaseQuery.call(db_client, update_pending) do
-      :ok
-    else
-      %{acc: _, msg: error_message} ->
-        Logger.warning("Database error: #{error_message}.")
-        {:error, :database_error}
+    case Repo.safe_update_all(device, []) do
+      {1, _} ->
+        :ok
 
       {:error, reason} ->
-        # DB Error
-        Logger.warning("Failed with reason #{inspect(reason)}.")
-        {:error, :database_error}
+        _ =
+          Logger.warning(
+            "Cannot set pending empty cache for device #{CoreDevice.encode_device_id(device_id)}: #{inspect(reason)}",
+            realm: realm,
+            tag: "set_pending_empty_cache_fail"
+          )
+
+        {:error, reason}
     end
   end
 
@@ -436,19 +434,19 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
     }
   end
 
-  def set_device_connected!(db_client, device_id, timestamp_ms, ip_address) do
-    set_connection_info!(db_client, device_id, timestamp_ms, ip_address)
+  def set_device_connected!(realm, device_id, timestamp, ip_address) do
+    set_connection_info!(realm, device_id, timestamp, ip_address)
 
     ttl = heartbeat_interval_seconds() * 8
-    refresh_device_connected!(db_client, device_id, ttl)
+    refresh_device_connected!(realm, device_id, ttl)
   end
 
-  def maybe_refresh_device_connected!(db_client, realm, device_id) do
+  def maybe_refresh_device_connected!(realm, device_id) do
     with {:ok, remaining_ttl} <- get_connected_remaining_ttl(realm, device_id) do
       if remaining_ttl < heartbeat_interval_seconds() * 2 do
         Logger.debug("Refreshing connected status", tag: "refresh_device_connected")
         write_ttl = heartbeat_interval_seconds() * 8
-        refresh_device_connected!(db_client, device_id, write_ttl)
+        refresh_device_connected!(realm, device_id, write_ttl)
       else
         :ok
       end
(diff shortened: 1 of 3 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -850,6 +839,10 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
     end
   end
 
+  defp compute_device_update_changes(device_id, changes) do
+    Ecto.Changeset.cast(%Device{device_id: device_id}, changes, Map.keys(changes))
+  end
+
```
lusergit, line 845: More of a comment than an actual change request (provide feedback on this): this is sort of considered an anti-pattern: the point of cast is to cast the actual fields of a schema and ignore keys that are not in the schema. It would be better to define a `changeset/3` function in the ecto schema itself and use that to generate the changeset, we already have some examples of this (e.g. [realm_management#interface.ex](https://github.com/astarte-platform/astarte/blob/463df387477131d5322b371ac09ed5130f5e78cb/apps/astarte_realm_management/lib/astarte_realm_management/realms/interface.ex#L65-L69)). Maybe in the device schema something like 
```elixir
def changeset(device,  params) do
  device
  |> cast(params, @permitted_params)
  |> validate_required(@required_params)
end
```
might be useful now/in the future

lusergit (commented):
overall LGTM, one small question :point_down:

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
Annopaolo: The idea here is to use a changeset for each specific case, rather than define a  "global" changeset that will be used in many places and could be adjusted to fit a number of different use cases. But this is not religion.
You can find a more thorough description of the proposal here: https://medium.com/very-big-things/towards-maintainable-elixir-the-anatomy-of-a-core-module-b7372009ca6d (paragraph "Building a changeset")

lusergit (approved):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -560,59 +560,72 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
-      |> DatabaseQuery.put(:introspection, introspection)
-      |> DatabaseQuery.put(:introspection_minor, introspection_minor)
-      |> DatabaseQuery.consistency(:quorum)
+    changes = %{
+      introspection: introspection,
+      introspection_minor: introspection_minor
+    }
 
-    DatabaseQuery.call!(db_client, introspection_update_query)
+    changeset = compute_device_update_changes(device_id, changes)
```
davidebriani, line 571: IMO `compute_device_update_changes` just adds a level of indirection instead of doing in place `Ecto.Changeset.cast(%Device{device_id: device_id}, changes, Map.keys(changes))`.
I don't see a particular reason why we need that line to be a reusable function, nor do we need to always cast every parameters.

In addition, we're not dealing with unsanitized user input here, we're managing Astarte internal data.
We can skip casting and validating params by using [Ecto.Changeset.change/2](https://hexdocs.pm/ecto/Ecto.Changeset.html#change/2).

```elixir
%Device{device_id: device_id} # Identifying struct with the primary key
|> Ecto.Changeset.change( # Changes
  introspection: introspection,
  introspection_minor: introspection_minor
)
|> Repo.update!(prefix: keyspace_name, consistency: :quorum)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -459,39 +457,26 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
-      |> DatabaseQuery.put(:last_connection, timestamp_ms)
-      |> DatabaseQuery.put(:last_seen_ip, ip_address)
-      |> DatabaseQuery.consistency(:local_quorum)
+    changes = %{
+      last_connection: timestamp,
+      last_seen_ip: ip_address
+    }
 
-    DatabaseQuery.call!(db_client, device_update_query)
+    changeset = compute_device_update_changes(device_id, changes)
```
davidebriani, line 468: Same here.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -459,39 +457,26 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
-    """
+  defp refresh_device_connected!(realm, device_id, ttl) do
+    keyspace_name = Realm.keyspace_name(realm)
 
-    refresh_connected_query =
-      DatabaseQuery.new()
-      |> DatabaseQuery.statement(refresh_connected_statement)
-      |> DatabaseQuery.put(:device_id, device_id)
-      |> DatabaseQuery.consistency(:local_quorum)
+    changeset = compute_device_update_changes(device_id, %{connected: true})
```
davidebriani, line 476: Same here.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -523,37 +508,28 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
+    changes = %{
+      connected: false,
+      last_disconnection: timestamp_ms,
+      total_received_msgs: total_received_msgs,
+      total_received_bytes: total_received_bytes,
+      exchanged_bytes_by_interface: interface_exchanged_bytes,
+      exchanged_msgs_by_interface: interface_exchanged_msgs
+    }
+
+    changeset = compute_device_update_changes(device_id, changes)
```
davidebriani, line 530: Same here.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -459,39 +457,26 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
-    refresh_connected_query =
-      DatabaseQuery.new()
-      |> DatabaseQuery.statement(refresh_connected_statement)
-      |> DatabaseQuery.put(:device_id, device_id)
-      |> DatabaseQuery.consistency(:local_quorum)
+    changeset = compute_device_update_changes(device_id, %{connected: true})
 
-    DatabaseQuery.call!(db_client, refresh_connected_query)
+    # We use `insert` here becuase Exandra does not support ttl on updates. However, this is an upsert in Scylla.
+    Repo.insert!(changeset, prefix: keyspace_name, ttl: ttl, consistency: :local_quorum)
```
davidebriani, line 479: It's fine to use an insert instead of an update, but I would double check which data is being written by Ecto: it should not overwrite all fields of an existing device with `nil` just because we're building a shallow Ecto struct here with just two fields.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/repo.ex
```diff
@@ -48,6 +48,15 @@ defmodule Astarte.DataUpdaterPlant.Repo do
     end
   end
 
+  def safe_update(queryable, updates, opts \\ []) do
+    try do
+      update_all(queryable, updates, opts)
```
davidebriani, line 53: Nitpick: maybe `safe_update_all` would better match the original function name?

davidebriani (changes requested):

davidebriani (approved):

PR #1097, opened 2025-02-28
refactor(RM): move delete_devices_with_data_on_interface

lusergit:
moves `delete_devices_with_data_on_interface` to use `exandra` and `ecto`

* [ ] Yes
* [x] No

### DIFF
--- a/apps/astarte_realm_management/lib/astarte_realm_management/queries.ex
+++ b/apps/astarte_realm_management/lib/astarte_realm_management/queries.ex
@@ -558,22 +558,15 @@ defmodule Astarte.RealmManagement.Queries do
     Repo.fetch_all(query, prefix: keyspace, consistency: :quorum)
   end
 
-  def delete_devices_with_data_on_interface(client, interface_name) do
-    devices_statement = "DELETE FROM kv_store WHERE group=:group_name"
+  def delete_devices_with_data_on_interface(realm_name, interface_name) do
+    keyspace = Realm.keyspace_name(realm_name)
+    group_name = "devices-with-data-on-interface-#{interface_name}-v0"
 
-    devices_query =
-      DatabaseQuery.new()
-      |> DatabaseQuery.statement(devices_statement)
-      |> DatabaseQuery.put(:group_name, "devices-with-data-on-interface-#{interface_name}-v0")
-      |> DatabaseQuery.consistency(:each_quorum)
+    query = from KvStore, where: [group: ^group_name]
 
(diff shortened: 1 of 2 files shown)


apps/astarte_realm_management/lib/astarte_realm_management/queries.ex
```diff
@@ -558,22 +558,14 @@ defmodule Astarte.RealmManagement.Queries do
...
+    query = from KvStore, where: [group: ^group_name]
 
-    with {:ok, _result} <- DatabaseQuery.call(client, devices_query) do
-      :ok
-    else
-      {:error, reason} ->
-        _ = Logger.warning("Database error: #{inspect(reason)}.", tag: "db_error")
-        {:error, :database_error}
-    end
+    with {:ok, _, _} <- Repo.match_delete(query, prefix: keyspace, consistency: :each_quorum),
```
davidebriani, line 567: Are we sure that there should be some match on the database?
I'm not sure we can make this assumption.

We can probably keep a simple call to `Repo.delete_all`, with the same behavior as the previous function.
That way we can also avoid "inventing" new creative Repo functions that might not be as straightforward as calling well known Repo functions directly.

apps/astarte_realm_management/lib/astarte_realm_management/queries.ex
```diff
@@ -558,21 +558,15 @@ defmodule Astarte.RealmManagement.Queries do
...
 
-    with {:ok, _result} <- DatabaseQuery.call(client, devices_query) do
-      :ok
-    else
-      {:error, reason} ->
-        _ = Logger.warning("Database error: #{inspect(reason)}.", tag: "db_error")
-        {:error, :database_error}
+    case Repo.delete_all(query, prefix: keyspace, consistency: :each_quorum) do
+      {0, _} -> {:error, :not_found}
+      {_n, _} -> :ok
```
davidebriani, line 569: My previous point was: are we sure that there is a match in the key value store?
If it's ok that no match exist, this code is returning an error even though interface deletion was successful.

davidebriani (changes requested):

apps/astarte_realm_management/lib/astarte_realm_management/queries.ex
lusergit: Ah you're right, re-looking at previous code it did not check for a match, in the last commit I solved this

davidebriani (approved):

PR #1103, opened 2025-03-03
refactor(RM): move `install_simple_trigger`

lusergit:
Moves `install_simple_trigger` to `exandra` and `ecto`

* [ ] Yes
* [x] No


apps/astarte_realm_management/lib/astarte_realm_management/queries.ex
```diff
@@ -920,52 +920,57 @@ defmodule Astarte.RealmManagement.Queries do
...
         trigger_target
       ) do
+    keyspace = Realm.keyspace_name(realm_name)
+
     insert_simple_trigger_statement = """
-    INSERT INTO simple_triggers
+    INSERT INTO #{keyspace}.simple_triggers
     (object_id, object_type, parent_trigger_id, simple_trigger_id, trigger_data, trigger_target)
-    VALUES (:object_id, :object_type, :parent_trigger_id, :simple_trigger_id, :simple_trigger_data, :trigger_target_data);
+    VALUES (?, ?, ?, ?, ?, ?);
```
Annopaolo, line 936: Why raw CQL here?

apps/astarte_realm_management/lib/astarte_realm_management/queries.ex
```diff
@@ -920,52 +920,57 @@ defmodule Astarte.RealmManagement.Queries do
...
-      not_ok ->
-        _ = Logger.warning("Database error: #{inspect(not_ok)}.", tag: "db_error")
-        {:error, :cannot_install_simple_trigger}
-    end
+    Exandra.execute_batch(
+      Repo,
+      %Exandra.Batch{
+        queries: [{insert_simple_trigger_statement, insert_params}, kv_insert]
+      }
+    )
```
Annopaolo, line 973: :eyes: you're changing the ordering of queries here: before it was `insert_simple_trigger_query` and then `insert_simple_trigger_by_uuid_query`, now it's the opposite

Annopaolo: Also, is it really necessary to use a batch for a single insert?

Annopaolo, line 973: What about an [`Ecto.Multi`](https://hexdocs.pm/ecto/Ecto.Multi.html)? It allows to group `insert_all` statements, and you could write everything in pure Ecto

Annopaolo (changes requested):

apps/astarte_realm_management/lib/astarte_realm_management/queries.ex
Annopaolo: Also, take into account the fact that multi-partition batches (as in this case) are recommended against due to the increrased load on the coordinator node (see https://docs.datastax.com/en/cql-oss/3.x/cql/cql_using/useBatchBadExample.html). Maybe going back to the two-line `with` is the best?

lusergit: Apparently `Ecto.Multi`s do not work with exandra, I rolled back to the two `with` solution :+1: 
(also did the same on the whole stack)

noaccOS: Just to add on this, Ecto transaction in general do not work. This also applies to [Changeset.prepare_changes](https://hexdocs.pm/ecto/Ecto.Changeset.html#prepare_changes/2) for instance

Annopaolo (approved):

PR #1118, opened 2025-03-04
refactor(appengine): centralize insert to sql definition

noaccOS:
using the standard ecto functions, there is no way to get the sql and params for an insertion from the adapter, which is instead the case for every other query type.

this pr defines a function, `insert_to_sql`, which behaves a bit like `to_sql` for insertions.

instead of using `Exandra.Connection.insert`, the logic was re-implemented here as for the little effort of copy-pasting the logic here, we get the added benefit of being able to declare custom syntax for the parameters, like using `intAsBlob(?)` for an insertion in the KvStore, while exandra is limited to just `?` directly.

Fixes #

* [ ] Yes
* [X] No

```docs

```

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/kv_store.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/kv_store.ex
@@ -54,12 +54,6 @@ defmodule Astarte.AppEngine.API.KvStore do
           Keyword.t()
         ) :: :ok | {:error, Exception.t()}
   def insert(kv_store_map, opts \\ []) do
-    %{
-      group: group,
-      key: key,
-      value: value
-    } = kv_store_map
-
     value_type = Map.get(kv_store_map, :value_type, :binary)
 
     value_expr =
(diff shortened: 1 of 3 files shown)


lusergit (approved):

apps/astarte_appengine_api/lib/astarte_appengine_api/kv_store.ex
```diff
@@ -54,13 +54,7 @@ defmodule Astarte.AppEngine.API.KvStore do
...
         ) :: :ok | {:error, Exception.t()}
   def insert(kv_store_map, opts \\ []) do
-    %{
-      group: group,
-      key: key,
-      value: value
-    } = kv_store_map
-
-    value_type = Map.get(kv_store_map, :value_type, :binary)
+    value_type = Map.pop(kv_store_map, :value_type, :binary)
```
davidebriani, line 57: Note that `Map.pop` returns `{value(), updated_map :: map()}` and not just the value directly.

davidebriani (changes requested):

davidebriani (approved):

PR #1124, opened 2025-03-05
ci(CI): refactor Astarte CI

lusergit:
Up to now build test and coverage workflows were triggered on every app on every code change. This was quite wasteful in terms of resources, as unnecessary checks were run on unchanged code.

This PR separates the workflows of every app into its own workflow, that gets triggered only when there are some changes in the relevant application.

`api` application's workflows get also triggered when there is a change in the non-`api` counterpart (to ensure that the communication between the two is consistent).

### DIFF
--- a/.github/workflows/astarte-housekeeping-api-workflow.yaml
+++ b/.github/workflows/astarte-housekeeping-api-workflow.yaml
@@ -1,0 +1,37 @@
+name: Build and Test Astarte Housekeeping API
+
+on:
+  # Run when pushing to stable branches
+  push:
+    paths:
+    - 'apps/astarte_housekeeping'
+    - 'apps/astarte_housekeeping_api'
+    - '.github/workflows/astarte-apps-build-workflow.yaml'
+    - '.github/workflows/astarte-housekeeping-api-workflow.yaml'
+    branches:
+    - 'master'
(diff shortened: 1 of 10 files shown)


.github/workflows/astarte-housekeeping-api-workflow.yaml
```diff
@@ -0,0 +1,35 @@
+name: Build and Test Astarte Housekeeping API
+
+on:
+  # Run when pushing to stable branches
+  push:
+    paths:
+    - 'apps/astarte_housekeeping_api'
+    - '.github/workflows/astarte-apps-build-workflow.yaml'
+    - '.github/workflows/astarte-housekeeping-api-workflow.yaml'
```
noaccOS, line 9: We should run the tests for the API modules even when only the backend module was modified

```suggestion
    paths:
    - 'apps/astarte_housekeeping'
    - 'apps/astarte_housekeeping_api'
    - '.github/workflows/astarte-apps-build-workflow.yaml'
    - '.github/workflows/astarte-housekeeping-api-workflow.yaml'
```

same for all other api modules

davidebriani (approved):

davidebriani:
❤️

PR #1127, opened 2025-03-07
refactor(RM): move `create` of individual datastream table

lusergit:
Previously when installing a new interface we ensured the `individual_datastream` table was present by creating it if needed with a custom `CREATE TABLE IF EXISTS` query. This pr removes the custom CQL code and insted runs an ecto migration when ths same action is needed/possible (only in one case, since in the others the query is heavily parametric and cannot be ported to an ecto query).


noaccOS (approved):
:1st_place_medal:

davidebriani (approved):

PR #1129, opened 2025-03-07
refactor(appengine): use datetimes for data insertion

noaccOS:
using integer timestamps, especially when doing calculations on them,
is error prone and we've just recently resolved a bug that was caused
by a double "microsecond to millisecond" conversion.

elixir has something more solid than that: `DateTime`s.

default Ecto types only allow full second or microsecond precision,
but cql timestamps use millisecond precision. To avoid precision errors,
a new Ecto DateTime type has been created, which allows millisecond
precision (althogh no check is made: Xandra handles that itself).

although currently unused, the submillis should now use the correct
number of digits.

an unused and nonfunctioning function in `IndividualProperty` has been
removed

Fixes submillis precision probably?

* [ ] Yes
* [X] No

```docs

```


noaccOS:
I've refactored `split_submillis` to make it more similar to the work done in https://github.com/astarte-platform/astarte/pull/1142 after the comments in https://github.com/astarte-platform/astarte/pull/1134

Annopaolo (approved):

PR #1130, opened 2025-03-07
Move keyspace translations to public functions in query files

eddbbt:
Fix astarte keyspace encoding, check and fix for functions that used to require a plain or encoded realm name in order to generally have:
def functions -> accepts plain realm name, encode into keyspace and then use that
defp -> accepts only keyspaces

There are actually one or two occurrences that require a rework to match the above case(s), but this would affect the entire application logic (maybe in another PR?)

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/groups/queries.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/groups/queries.ex
@@ -70,7 +70,7 @@ defmodule Astarte.AppEngine.API.Groups.Queries do
     with {:ok, %{device_id: device_id}} <-
            Ecto.Changeset.apply_action(device_changeset, :insert),
          {:group_exists?, true} <-
-           {:group_exists?, group_exists?(realm_name, group_name)},
+           {:group_exists?, group_exists?(keyspace, group_name)},
          :ok <- check_valid_device_for_group(keyspace, group_name, device_id),
          :ok <- add_to_group(keyspace, group_name, [device_id]) do
       :ok
@@ -94,9 +94,10 @@ defmodule Astarte.AppEngine.API.Groups.Queries do
   end
 
   def remove_device(realm_name, group_name, device_id) do
+    keyspace = DataAccessRealm.keyspace_name(realm_name)
+
     with {:group_exists?, true} <-
-           {:group_exists?, group_exists?(realm_name, group_name)},
-         keyspace = DataAccessRealm.keyspace_name(realm_name),
+           {:group_exists?, group_exists?(keyspace, group_name)},
          :ok <- remove_from_group(keyspace, group_name, device_id) do
       :ok
     else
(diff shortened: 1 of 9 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/groups/queries.ex
```diff
@@ -387,12 +381,8 @@ defmodule Astarte.AppEngine.API.Groups.Queries do
     end
   end
 
-  defp prepare_with_realm(conn, realm_name, query) do
-    keyspace_name =
-      CQLUtils.realm_name_to_keyspace_name(realm_name, Config.astarte_instance_id!())
-
-    with {:valid, true} <- {:valid, Realm.valid_name?(realm_name)},
```
noaccOS, line 394: this makes me think: are we actually not checking for valid realm names anymore? Are we allowing the user to query on the `system` keyspace?

apps/astarte_appengine_api/lib/astarte_appengine_api/health.ex
```diff
@@ -17,23 +17,21 @@
...
-  alias Astarte.AppEngine.API.Realm
   alias Astarte.AppEngine.API.Queries
 
   require Logger
 
   def get_health do
-    astarte_keyspace = Realm.keyspace_name("astarte")
-
-    case Queries.check_astarte_health(astarte_keyspace, :quorum) do
+    # TODO shall we check for an encoded astarte realm or it's fine to use the generic "astarte one"?
```
noaccOS, line 25: `astarte` is fine, as we're checking for astarte's health. Astarte is healthy even if there are no installed realms, but it needs the `astarte` realm to function

apps/astarte_appengine_api/lib/astarte_appengine_api/groups/queries.ex
eddbbt: This is actually a good point, let me fix this asap

apps/astarte_appengine_api/lib/astarte_appengine_api/realms/realm.ex
```diff
@@ -29,7 +30,15 @@ defmodule Astarte.AppEngine.API.Realm do
...
+    CQLUtils.realm_name_to_keyspace_name("astarte", Config.astarte_instance_id!())
+  end
+
   def keyspace_name(realm_name) do
-    CQLUtils.realm_name_to_keyspace_name(realm_name, Config.astarte_instance_id!())
+    keyspace = CQLUtils.realm_name_to_keyspace_name(realm_name, Config.astarte_instance_id!())
+
+    with {:valid, true} <- {:valid, Realm.valid_name?(keyspace)} do
+      keyspace
+    end
```
noaccOS, line 42: This is a bit all over the place:
- the spec says the return type is a string, but it now is `String.t() | {:valid, false}`. I suggest we either change this function to a `{:ok, keyspace_name} | {:error, :invalid_realm_name}` or crash instead.
- `"astarte"` is not a valid realm name, and as such you declared its own body to make sure calls to it still succeed. We do not want users querying on the `astarte` keyspace, I suggest we make another ad-hoc function used only for the queries actually requiring the astarte keyspace, something like `Realm.astarte_keyspace_name()`

davidebriani (approved):

PR #1132, opened 2025-03-10
refactor(trigger_engine): move xandra select to exandra

emirh1:

Porting all Xandra SELECT queries to Exandra

- retrieve_policy_data 
- get_astarte_health
- list_policies
- list_realms

link:
https://www.notion.so/seco-mind/Move-all-SELECT-queries-with-Xandra-to-Exandra-1a0c537dba7f808596c9efc047a251a0

* [ ] Yes
* [x] No


apps/astarte_trigger_engine/lib/astarte_trigger_engine/health/queries.ex
```diff
@@ -19,17 +19,22 @@
...
-    with {:ok, %Xandra.Page{} = page} <-
-           Xandra.Cluster.execute(:xandra, query, %{}, consistency: consistency),
-         {:ok, _} <- Enum.fetch(page, 0) do
+    keyspace_name =
+      CQLUtils.realm_name_to_keyspace_name("astarte", Config.astarte_instance_id!())
+
+    query =
+      from r in Realms,
+        prefix: ^keyspace_name,
+        select: count(r.realm_name)
```
noaccOS, line 35: nit: you don't need to specify the primary key

```suggestion
        select: count()
```

better yet, you can also use [`Repo.aggregate`](https://hexdocs.pm/ecto/Ecto.Repo.html#c:aggregate/3).

apps/astarte_trigger_engine/lib/astarte_trigger_engine/policy/queries.ex
```diff
@@ -18,33 +18,34 @@
...
+        prefix: ^keyspace_name,
+        where: k.group == "trigger_policy" and k.key == ^policy_name,
+        select: k.value
 
-    with {:ok, prepared} <-
-           Xandra.prepare(conn, retrieve_statement),
-         {:ok, %Xandra.Page{} = page} <-
-           Xandra.execute(conn, prepared, %{"policy_name" => policy_name}),
-         [%{"value" => policy}] <- Enum.to_list(page) do
+    with policy when policy != nil <- Repo.one!(query) do
```
noaccOS, line 48: see above comment about `nil` and `one!`

apps/astarte_trigger_engine/lib/astarte_trigger_engine/amqp_consumer/queries.ex
```diff
@@ -19,73 +19,52 @@
...
-         {:ok, %Xandra.Page{} = page} <-
-           Xandra.execute(conn, prepared, %{}),
-         policy_list <- Enum.map(page, &extract_name_and_data/1) do
-      {:ok, policy_list}
-    else
-      {:error, %Xandra.Error{} = err} ->
-        _ = Logger.warning("Database error: #{inspect(err)}.", tag: "database_error")
-        {:error, :database_error}
+    case Repo.safe_fetch_all(query, []) do
+      policies ->
```
noaccOS, line 39: This will not work as expected:

`case` blocks select the first branch which matches the result of the argument (in this case `Repo.safe_fetch_all(...)`)
as the first branch is just an anonymous variable, it will always match regardless of the return type of `safe_fetch_all`.

You can approach this issue in two ways:
1. by declaring the `:error` branch first, it will match whenever the argument returns an error, and then it will fallback to the `:ok` variant in other situations
  ```elixir
  case_result_first = fn value ->
    case value do
      result -> {:ok, result}
      {:error, reason} -> {:error, reason}
    end
  end
  #Function<42.105768164/1 in :erl_eval.expr/6>
  
  case_error_first = fn value ->
    case value do
      {:error, reason} -> {:error, reason}
      result -> {:ok, result}
    end
  end
  #Function<42.105768164/1 in :erl_eval.expr/6>
  
  case_result_first.([1,2,3]) # => {:ok, [1, 2, 3]}
  case_result_first.({:error, :database_error}) # => {:ok, {:error, :database_error}}
  
  case_error_first.([1,2,3]) # => {:ok, [1, 2, 3]}
  case_error_first.({:error, :database_error}) # => {:error, :database_error}
  ```
2. by editing `Repo.safe_fetch_all` to return a result tuple, like so
  ```elixir
    def safe_fetch_one(queryable, opts \\ []) do
      try do
        {:ok, one(queryable, opts)}
      catch
        error ->
          handle_xandra_error(error)
      end
    end
  ```
  you can then match the result tuple on this `case`:
  ```elixir
  case Repo.safe_fetch_all(query, []) do
    {:ok, policies} -> # ...
    {:error, reason} -> # ...
  end
  ```

the second method also has the advantage of keeping the typing consistent!

apps/astarte_trigger_engine/lib/astarte_trigger_engine/health/queries.ex
```diff
@@ -19,42 +19,21 @@
...
-         {:ok, _} <- Enum.fetch(page, 0) do
-      :ok
-    else
-      :error ->
-        _ =
-          Logger.warning("Cannot retrieve count for astarte.realms table.",
-            tag: "health_check_error"
-          )
+    case Repo.aggregate(Realms, :count, prefix: keyspace_name, consistency: consistency) do
+      :ok ->
```
noaccOS, line 31: look out for types! An aggregate of type `:count` always return an integer, not `:ok | {:error, reason}`

you can tackle this a couple ways
1. by adding a `safe_aggregate` function to repo, like `safe_all` and friends
2. you could just use your old query with `Repo.safe_all`

apps/astarte_trigger_engine/lib/astarte_trigger_engine/policy/queries.ex
```diff
@@ -18,46 +18,40 @@
...
+        where: k.group == "trigger_policy" and k.key == ^policy_name,
+        select: k.value
 
-      {:error, %Xandra.ConnectionError{} = err} ->
-        _ =
-          Logger.warning("Database connection error: #{inspect(err)}.",
-            tag: "database_connection_error"
-          )
+    case Repo.safe_fetch_one(query, []) do
+      policy ->
```
noaccOS, line 49: the comment I left for `safe_fetch_all` applies here as well

noaccOS (changes requested):
Already looking much better! Remember to also try running the tests locally to catch errors early!

apps/astarte_trigger_engine/lib/astarte_trigger_engine/policy/queries.ex
```diff
@@ -18,46 +18,40 @@
...
   alias Astarte.TriggerEngine.Config
+  alias Astarte.TriggerEngine.KvStore
+  alias Astarte.TriggerEngine.Repo
+
   require Logger
 
-  alias Astarte.Core.Realm
+  import Ecto.Query
 
   def retrieve_policy_data(realm_name, policy_name) do
```
Annopaolo, line 30: This function seem to be unused. If that's the case, it can be removed

Annopaolo: Edit: the whole file seems to be unused

Annopaolo (changes requested):

PR #1133, opened 2025-03-10
refactor(dup): move xandra select to exandra

shinnokdisengir:

Move Xandra Select to Exandra:
- retrieve_policy_name

* [ ] Yes
* [x] No

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/trigger_policy/queries.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/trigger_policy/queries.ex
@@ -18,36 +18,30 @@
 
 defmodule Astarte.DataUpdaterPlant.TriggerPolicy.Queries do
   require Logger
-  alias Astarte.Core.CQLUtils
-  alias Astarte.DataUpdaterPlant.Config
+
+  import Ecto.Query
+
+  alias Astarte.DataUpdaterPlant.Repo
+  alias Astarte.DataUpdaterPlant.DataUpdater.KvStore
+  alias Astarte.DataUpdaterPlant.DataUpdater.Realm
 
   def retrieve_policy_name(realm_name, trigger_id) do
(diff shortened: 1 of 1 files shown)


lusergit (approved):
LGTM! :thumbsup:

eddbbt (approved):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/trigger_policy/queries.ex
```diff
@@ -18,36 +18,30 @@
...
-        [%{"value" => policy_name}] -> {:ok, policy_name}
-        [] -> {:error, :policy_not_found}
-      end
-    else
-      {:error, error} ->
-        Logger.warning("Database error #{inspect(error)}", tag: "retrieve_policy_name_db_error")
-        {:error, error}
+    case Repo.all(query) do
+      [%{"value" => policy_name}] -> {:ok, policy_name}
+      [] -> {:error, :policy_not_found}
```
noaccOS, line 44:
```suggestion
    case Repo.one(query) do
      nil -> {:error, :policy_not_found}
      value -> {:ok, value}
```

unfortunately it's not obvious from the tests as it always returns `{:error, :policy_not_found}`, but when you `select:` a single field from the capture it returns just that value and not the whole struct (it would have had to be `value:` (the atom) even if it was the whole struct)

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/trigger_policy/queries.ex
shinnokdisengir: Ok! I understood that it was better to use `Repo.all` (it lifts less times)

noaccOS (approved):

Annopaolo (approved):

PR #1152, opened 2025-03-18
(DUP) Update with data_access schemas

eddbbt:
Update dependencies (astarte_core, data_access) due to new functionalities not yet published as releases
Replace schemas with ones from data_access
Remove unused schema files


lusergit (approved):

apps/astarte_data_updater_plant/mix.exs
```diff
@@ -68,7 +68,8 @@ defmodule Astarte.DataUpdaterPlant.Mixfile do
 
   defp astarte_required_modules(_) do
     [
-      {:astarte_core, "~> 1.2"},
+      {:astarte_core, github: "astarte-platform/astarte_core", branch: "release-1.2"},
+      # {:astarte_core, "~> 1.2"},
```
Annopaolo, line 72: Leftover

apps/astarte_data_updater_plant/mix.exs
```diff
@@ -82,7 +83,7 @@ defmodule Astarte.DataUpdaterPlant.Mixfile do
       {:cyanide, "~> 2.0"},
       {:excoveralls, "~> 0.15", only: :test},
       {:mox, "~> 1.0", only: :test},
-      {:exandra, github: "vinniefranco/exandra"},
+      {:exandra, github: "vinniefranco/exandra", override: true},
```
Annopaolo, line 86: Let's use exandra 0.13 from hex.pm, which is at the same commit as master right now

Annopaolo (changes requested):

noaccOS (approved):

PR #1169, opened 2025-03-25
fix(TE): Delegate health check query to standardized one on DataAccess and fix a crash loop caused by incorrect health reporting

OmarBrbutovic:

- Delegates the health check query to the standardized one on `DataAccess`, and in the process fix a crash loop caused by incorrect health reporting.
- Fixed incorrect DateTime format conversion by replacing `DateTime.from_unix/2` with `DateTime.to_unix/2`.

* [ ] Yes
* [x] No


apps/astarte_trigger_engine/lib/astarte_trigger_engine/health/queries.ex
```diff
@@ -32,12 +32,14 @@ defmodule Astarte.TriggerEngine.Health.Queries do
         select: count()
 
     case Repo.safe_fetch_one(query, consistency: consistency) do
-      {:ok, count} ->
-        {:ok, count}
```
eddbbt, line 36: do the callers expect back a tuple or `:ok` ?

eddbbt, line 36: do the callers expect back a tuple or `:ok` ?

OmarBrbutovic: The caller of `get_astarte_health/1` expects a simple `:ok` on success, rather than a tuple like `{:ok, count}`. This aligns with the existing health check logic in `get_health/0`, which directly matches against `:ok`.

eddbbt (approved):

Annopaolo (approved):

PR #1176, opened 2025-03-28
Erlang clustering in docker compose

davidebriani:

This change introduces support for clustering DUP and Appengine services in the Docker Compose environment.
This is a prerequisite for reworking Astarte's RPC system to use native Erlang clustering and message passing.

This setup allows for clustering both DUP and Appengine together, but also replicas of the same service.
As of now, replicating DUP is not straightfoward as it is not purely stateless and it requires different sets of environment variables for each replica to configure AMQP message consumption.
For Appengine, clustering between multiple replicas can be tested out by specifying the number of replicas in the docker-compose.yaml file.

```yaml
  astarte-appengine-api:
    deploy:
      replicas: 2
```

Then run the services:

```sh
docker compose pull
docker compose build astarte-appengine-api astarte-data-updater-plant
docker compose up -d
docker compose exec -it astarte-appengine-api ./bin/astarte_appengine_api remote
```

and inspect the Erlang nodes from the remote iEX session:

```elixir
Node.self() # the current Erlang node
Node.list() # other Erlang nodes in the cluster
Node.get_cookie() # the Erlang cookie, that must be shared between nodes
```

* [ ] Yes
* [x] No


apps/astarte_appengine_api/lib/astarte_appengine_api/config.ex
```diff
@@ -279,6 +279,26 @@ defmodule Astarte.AppEngine.API.Config do
...
+              node_basename: "astarte_data_updater_plant"
+            ]
+          ],
+          appengine_api: [
+            strategy: Elixir.Cluster.Strategy.DNSPoll,
+            config: [
+              polling_interval: 5_000,
+              query: "astarte-appengine-api",
+              node_basename: "astarte_appengine_api"
+            ]
```
lusergit, line 299: I think appengine api instances only need to find data updater plants to make RPC requests, no need to find other appengine instances

lusergit (changes requested):
Overall LGTM :+1: Just a couple of nits

Annopaolo (approved):

PR #1177, opened 2025-03-31
fix: kubernetes clustering

lusergit:
Kubernetes clustering strategies were misconfigured and required to an unnecessary update from the operator. This PR introduces the necessary fixes to solve 2 main issues:
- DUP clustering strategy should only be responsible for finding other DUP instances. Up to not a general clustering approach was used to allow all nodes to find each other. What we actually want to do is allow DUP instances to find each other and leave to appegine to find DUP instances to talk to (that logic is in fact moved into appengine). DUP clustering strategy now only tries to find other DUP nodes and by default leverages already available endpoint labels to query the kubernetes API.
- As anticipated, AppEngine instances should only be responsible for finding DUP instances. With this changes they not try anymore to find other appengine instances (as it is not necessary). They also leverage already existing endpoint labels, this requires less spin/changes with the operator.

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/config.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/config.ex
@@ -140,16 +140,16 @@ defmodule Astarte.AppEngine.API.Config do
           :astarte_data_updater_plant,
           :clustering_strategy,
           os_env: "CLUSTERING_STRATEGY",
-          type: Astarte.AppEngine.API.ClusteringStrategy,
+          type: Astarte.AppEngine.API.Config.ClusteringStrategy,
           default: "none"
 
-  @envdoc "The Kubernetes selector to use when `kubernetes` Erlang clustering strategy is used. Defaults to `app=astarte-data-updater-plant`."
-  app_env :clustering_kubernetes_selector,
+  @envdoc "The Endpoint label to use to query Kubernetes to find data updater plant instances. Defaults to `app=astarte-data-updater-plant`."
+  app_env :dup_clustering_kubernetes_selector,
           :astarte_data_updater_plant,
-          :clustering_kubernetes_selector,
-          os_env: "CLUSTERING_KUBERNETES_SELECTOR",
+          :dup_clustering_kubernetes_selector,
+          os_env: "DATA_UPDATER_PLANT_CLUSTERING_KUBERNETES_SELECTOR",
           type: :binary,
-          default: "clustering=astarte"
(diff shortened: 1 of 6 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/config.ex
```diff
@@ -140,16 +140,16 @@ defmodule Astarte.AppEngine.API.Config do
...
 
-  @envdoc "The Kubernetes selector to use when `kubernetes` Erlang clustering strategy is used. Defaults to `app=astarte-data-updater-plant`."
-  app_env :clustering_kubernetes_selector,
+  @envdoc "The Endpoint label to use to query Kubernetes to find data updater plant instances. Defaults to `app=astarte-data-updater-plant`."
+  app_env :dup_clustering_kubernetes_selector,
           :astarte_data_updater_plant,
-          :clustering_kubernetes_selector,
-          os_env: "CLUSTERING_KUBERNETES_SELECTOR",
+          :dup_clustering_kubernetes_selector,
+          os_env: "DUP_CLUSTERING_KUBERNETES_SELECTOR",
```
Annopaolo, line 150: Nit: you are looking for a DUP cluster and we're using full names in env vars, so I'd suggest
```suggestion
          os_env: "DATA_UPDATER_PLANT_CLUSTER_KUBERNETES_SELECTOR",
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/config.ex
```diff
@@ -254,16 +254,16 @@ defmodule Astarte.DataUpdaterPlant.Config do
...
 
-  @envdoc "The Kubernetes selector to use when `kubernetes` Erlang clustering strategy is used. Defaults to `clustering=astarte`."
-  app_env :clustering_kubernetes_selector,
+  @envdoc "The endpoint label to query to get other data updater plant instances. Defaults to `app=astarte-data-updater-plant`."
+  app_env :dup_clustering_kubernetes_selector,
           :astarte_data_updater_plant,
-          :clustering_kubernetes_selector,
-          os_env: "CLUSTERING_KUBERNETES_SELECTOR",
+          :dup_clustering_kubernetes_selector,
+          os_env: "DUP_CLUSTERING_KUBERNETES_SELECTOR",
```
Annopaolo, line 264: See comment above

Annopaolo (changes requested):

davidebriani (approved):

PR #1190, opened 2025-04-04
Add libcluster topology for VerneMQ discovery in DUP

davidebriani:

This PR introduces a libcluster topology to enable the DUP service to automatically discover and connect to all VerneMQ replicas across different environments.

- Docker-compose: a topology tailored for this environment is added, relying on DNS facilities provided by Docker to discover VerneMQ nodes at known domain names, i.e. the name of the VerneMQ service as defined in docker-compose.yaml.
- Kubernetes: for production scenarios, VerneMQ is deployed by the Astarte Operator as a StatefulSet. As such, the libcluster strategy attempts first attempts to discover node IPs by querying for app labels, as in the other Kubernetes strategy already defined. Then it assumes specific names for the Erlang nodes using the discovered IPs and the fact that nodes in the StatefulSet have deterministic names that include a progressive counter such as `VerneMQ@<service_name>-0.<service_name>.<namespace>.svc.<cluster_domain>`

This change is a prerequisite to allow DUP to communicate with VerneMQ for RPC operations, using Erlang's native message passing.

* [ ] Yes
* [x] No

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/config.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/config.ex
@@ -265,6 +265,14 @@ defmodule Astarte.DataUpdaterPlant.Config do
           type: :binary,
           default: "app=astarte-data-updater-plant"
 
+  @envdoc "The Endpoint label to use to query Kubernetes to find vernemq instances. Defaults to `app=astarte-vernemq`."
+  app_env :vernemq_clustering_kubernetes_selector,
+          :astarte_data_updater_plant,
+          :vernemq_clustering_kubernetes_selector,
+          os_env: "VERNEMQ_CLUSTERING_KUBERNETES_SELECTOR",
+          type: :binary,
+          default: "app=astarte-vernemq"
+
   @envdoc "The Kubernetes namespace to use when `kubernetes` Erlang clustering strategy is used. Defaults to `astarte`."
   app_env :clustering_kubernetes_namespace,
           :astarte_data_updater_plant,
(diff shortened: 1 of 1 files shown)


lusergit (approved):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/config.ex
```diff
@@ -514,6 +522,18 @@ defmodule Astarte.DataUpdaterPlant.Config do
...
               polling_interval: 10_000
             ]
+          ],
+          vernemq_k8s: [
+            strategy: Elixir.Cluster.Strategy.Kubernetes,
+            config: [
+              mode: :hostname,
+              kubernetes_service_name: "astarte-vernemq",
+              kubernetes_node_basename: "VerneMQ",
+              kubernetes_ip_lookup_mode: :pods,
```
Annopaolo, line 532: Couldn't we use the default value (i.e. `:endpoints`) here?

davidebriani: Currently, we cannot.
Libcluster is looking for endpoints with `kubectl get endpoints -l app=astarte-vernemq`.
However the endpoints for VerneMQ do not contain labels: we should change the Astarte Operator to rectify that.

Instead of endpoints, looking for pods already works out of the box.

lusergit: The vernemq endpoint does not have labels (a part for `service.kubernetes.io/headless`, set to `""`) but I dont think it is reliable. If looking for pods, the pods with label `app=astarte-vernemq` uniquely identify vernemq pods in the cluster

PR #1191, opened 2025-04-04
feat(AE): VerneMQ node discovery

lusergit:
Similar to #1190 

Adds support for VerneMQ node discovery in an astarte clustering, leveraging `libcluster` configuration

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/config.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/config.ex
@@ -151,6 +151,14 @@ defmodule Astarte.AppEngine.API.Config do
           type: :binary,
           default: "app=astarte-data-updater-plant"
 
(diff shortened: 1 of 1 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/config.ex
```diff
@@ -151,6 +151,14 @@ defmodule Astarte.AppEngine.API.Config do
           type: :binary,
           default: "app=astarte-data-updater-plant"
 
+  @envdoc "The Endpoint label to use to query Kubernetes to find data updater plant instances. Defaults to `app=astarte-vernemq`."
```
davidebriani, line 154:
```suggestion
  @envdoc "The Endpoint label to use to query Kubernetes to find vernemq instances. Defaults to `app=astarte-vernemq`."
```

davidebriani (changes requested):

PR #1194, opened 2025-04-07
refactor(AE): VerneMQ RPC

lusergit:
Based on #1191
Requires astarte-platform/astarte_vmq_plugin#102

This PR moves the communication between AppEngine and the Astarte VerneMQ Plugin outside of `astarte_rpc`. The new strategy consists in having an rpc server shared among all VerneMQ Plugin replicas (see astarte-platform/astarte_vmq_plugin#102 for context) and making RPC calls to shuch server. AppEngine calls have to be reworked to accomodate for the changes

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/rpc/vmq_plugin.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/rpc/vmq_plugin.ex
@@ -22,38 +22,22 @@ defmodule Astarte.AppEngine.API.RPC.VMQPlugin do
   """
   require Logger
 
-  alias Astarte.RPC.Protocol.VMQ.Plugin, as: Protocol
-
-  alias Astarte.RPC.Protocol.VMQ.Plugin.{
-    Call,
-    GenericErrorReply,
-    GenericOkReply,
-    Publish,
-    PublishReply,
-    Reply
-  }
-
-  alias Astarte.AppEngine.API.Config
-
-  require Logger
-
-  @rpc_client Config.rpc_client!()
-  @destination Protocol.amqp_queue()
+  @rpc_behaviour Application.compile_env(
+                   :astarte_appengine_api,
+                   :vernemq_plugin_rpc_client,
+                   Astarte.AppEngine.API.RPC.VMQPlugin.Client
+                 )
 
   def publish(topic, payload, qos)
       when is_binary(topic) and is_binary(payload) and is_integer(qos) and qos >= 0 and qos <= 2 do
     with {:ok, tokens} <- split_topic(topic) do
-      _ = Logger.debug("Going to publish value on MQTT.")
-
-      %Publish{
+      data = %{
         topic_tokens: tokens,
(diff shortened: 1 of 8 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/rpc/vmq_plugin.ex
```diff
@@ -22,38 +22,22 @@ defmodule Astarte.AppEngine.API.RPC.VMQPlugin do
...
-
-  @rpc_client Config.rpc_client!()
-  @destination Protocol.amqp_queue()
+  @rpc_behaviour Application.compile_env(
+                   :astarte_appengine_api,
+                   :vernemq_plugin_rpc_client,
+                   Astarte.AppEngine.API.RPC.VMQPlugin.Client
+                 )
 
   def publish(topic, payload, qos)
```
davidebriani, line 31: The previous implementation of `publish/3` returned `:ok`, while now it returns:
- `:ok` in tests, since the `@rpc_behaviour` is mocked to return `:ok`.
- `{:ok, %{...}}` in production, when using the actual implementation for `@rpc_behavior`.

I wonder if callers of `publish/3` expect a `:ok` return value, and so tests pass but the code may break in production.
Let's double check this. And even if it's not the case, I would strive to have a single specific return value: `:ok` is fine, since it already was fine before.
Arguably, we could even have VerneMQ just return `:ok` if what is currently returning (`%{local_matches: ..., remote_matches: ...}`) is not of interest, but that's a separate issue.

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_test.exs
```diff
@@ -153,9 +153,16 @@ defmodule Astarte.AppEngine.API.DeviceTest do
...
 
+  setup :verify_on_exit!
+
   setup_all do
     {:ok, _client} = DatabaseTestHelper.create_test_keyspace()
 
+    Mox.stub_with(
+      Astarte.AppEngine.API.RPC.VMQPlugin.ClientMock,
+      Astarte.AppEngine.API.RPC.VMQPlugin.Client
+    )
```
davidebriani, line 164: Do we really need to stub the mock with the actual module implementation?

apps/astarte_appengine_api/lib/astarte_appengine_api/rpc/vmq_plugin.ex
Annopaolo: I would leave `{:ok, %{local_matches: integer(), remote_matches: integer()}}` as it was before, because the number of local/remote matches might become relevant in case of a netsplit

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_test.exs
lusergit: No ! fixing :+1:

davidebriani (approved):

PR #1207, opened 2025-04-14
test(RM): add astarte_generators

lusergit:
Based on #1202

Adds property based testing for the engine module, testing `interface` operations and `trigger_policy` operations.

## Interfaces
- A valid interface gets installed
- A deletion is performed only if the major version is 0
- An update is performed only on interfaces with major version 0
- Only valid updates are performed

## Trigger policies
- valid trigger policies get installed
- trigger policies can be deleted


apps/astarte_realm_management/test/astarte_realm_management/new_engine_test.exs
```diff
@@ -33,5 +33,35 @@ defmodule Astarte.RealmManagement.EngineTestv2 do
...
+        assert interface.name in interfaces
+      end
+    end
+
+    property "is deleted if the major version is 0", %{realm: realm} do
+      check all(interface <- Astarte.Core.Generators.Interface.interface(major_version: 0)) do
+        json_interface = Jason.encode!(interface)
+
+        _ = Engine.install_interface(realm, json_interface)
+        {:ok, interfaces} = Engine.get_interfaces_list(realm)
```
noaccOS, line 61: leftover line? I think you're not using this result

lusergit: removed :+1:

apps/astarte_realm_management/test/astarte_realm_management/new_engine_test.exs
```diff
@@ -0,0 +1,37 @@
...
+
+  describe "Test interface" do
+    property "is installed properly", %{realm: realm} do
+      check all(interface <- Astarte.Core.Generators.Interface.interface()) do
+        json_interface = Jason.encode!(interface)
+
+        _ = Engine.install_interface(realm, json_interface)
+
+        {:ok, interfaces} = Engine.get_interfaces_list(realm)
+        assert interface.name in interfaces
```
Annopaolo, line 35: Why not the stricter check?
```elixir
interface_source = Engine.get_interface_source(realm, interface.name, interface.major_version)
# might need some more fiddling to match mappings order
assert interface_source = json_interface
```

Annopaolo (commented):
Nit: commit `dd870` does not `test engine properties`, it only tests one specific property

apps/astarte_realm_management/test/astarte_realm_management/new_engine_test.exs
```diff
@@ -146,4 +148,28 @@ defmodule Astarte.RealmManagement.EngineTestv2 do
...
+
+  describe "Test trigger policy" do
+    @describetag :trigger_policy
+
+    property "is installed correctly", %{realm: realm} do
+      check all(policy <- Astarte.Core.Generators.Triggers.Policy.policy()) do
+        policy_json = Jason.encode!(policy)
+        :ok = Engine.install_trigger_policy(realm, policy_json)
+
+        assert {:ok, _source} = Engine.trigger_policy_source(realm, policy.name)
```
Annopaolo, line 160: Nit: this does not check that what was installed is actually the same policy passed to `Engine.install_trigger_policy`, only that there exists a policy with the same name

Annopaolo (changes requested):

Annopaolo (approved):

PR #1209, opened 2025-04-15
test(RM): add astarte_generators

lusergit:
Based on #1207

Adds tests for `Trigger`s in Realm Management. It also implements some basic generators for triggers from fixtures and name generators, to be substituted wiht an actual trigger generator once available in [astarte-platform/astarte_generators](https://github.com/astarte-platform/astarte_generators)

Signed-off-by: Luca Zaninotto <luca.zaninotto@secomind.com>


apps/astarte_realm_management/test/astarte_realm_management/new_engine_test.exs
```diff
@@ -213,6 +217,59 @@ defmodule Astarte.RealmManagement.EngineTestv2 do
...
+          end)
+
+        simple_trigger =
+          unless simple_trigger.interface_major,
+            do: Map.put(simple_trigger, :interface_major, 0),
+            else: simple_trigger
+
+        assert serialized_tagged_simple_triggers == [simple_trigger]
+
+        _ = Engine.delete_trigger(realm, trigger.name)
```
Annopaolo, line 268: This would fit better in the `on_exit` callback:
```suggestion
        on_exit(fn -> Engine.delete_trigger(realm, trigger.name) end)
```

apps/astarte_realm_management/coveralls.json
Annopaolo, line 1: :fire:

Annopaolo (changes requested):

apps/astarte_realm_management/test/astarte_realm_management/new_engine_test.exs
lusergit: With an `on_exit` the function is called only after the whole property is tested, while that call has to be made at each `check`. The generated triggers might have the same name, so at the end of each run if there are no errors is better to delete the newly installed trigger.

Annopaolo: OK!

PR #1210, opened 2025-04-16
HK Rewrite queries tests

eddbbt:
Rationalize and expand queries test
Add new test dependencies
Clean up queries file: Db errors handling have been moved to the parent function

### DIFF
--- a/apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
+++ b/apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
@@ -161,6 +161,18 @@ defmodule Astarte.Housekeeping.Queries do
            :ok <- execute_realm_deletion(conn, realm_name, keyspace_name) do
         :ok
       else
+        {:error, %Xandra.Error{} = err} ->
+          _ = Logger.warning("Database error: #{inspect(err)}.", tag: "database_error")
+          {:error, :database_error}
+
+        {:error, %Xandra.ConnectionError{} = err} ->
+          _ =
+            Logger.warning("Database connection error: #{inspect(err)}.",
+              tag: "database_connection_error"
+            )
+
+          {:error, :database_connection_error}
+
         {:error, reason} ->
           _ =
             Logger.warning("Cannot delete realm: #{inspect(reason)}.",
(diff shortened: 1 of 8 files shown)


apps/astarte_housekeeping/lib/astarte_housekeeping/queries.ex
```diff
@@ -1099,6 +888,18 @@ defmodule Astarte.Housekeeping.Queries do
...
+          {:error, :database_error}
+
+        {:error, %Xandra.ConnectionError{} = err} ->
+          _ =
+            Logger.warning("Database connection error: #{inspect(err)}.",
+              tag: "database_connection_error"
+            )
+
+          {:error, :database_connection_error}
+
```
lusergit, line 902: Removed everywhere but here?

eddbbt: this is one of the "top" functions, basically works as a catcher for errors from row 844-847

apps/astarte_housekeeping/test/helpers/db_test_helper.ex
Annopaolo, line 1: Most of these queries should be moved to Data Access because they are common to other Astarte services (e.g. device insertion is used in almost all tests)

apps/astarte_housekeeping/test/test_helper.exs
```diff
@@ -1 +1,3 @@
+Mimic.copy(Xandra)
```
Annopaolo, line 1: Is `Xandra` actually omcked somewhere or just `Xandra.Cluster`?

Annopaolo (changes requested):

apps/astarte_housekeeping/test/helpers/db_test_helper.ex
eddbbt: I think the long term point is exactly this?

apps/astarte_housekeeping/test/test_helper.exs
eddbbt: yes row 248 for example, queries test file

Annopaolo (approved):

Annopaolo (approved):

PR #1211, opened 2025-04-16
test(RM): parallel testing

lusergit:
Based on #1209 

Allows for parallel testing in Realm Management:
- Introduces [mimic](https://github.com/edgurgel/mimic) the mock library to mock the `Astarte.Config` module to give each and every test module a separate astarte instance id, to allow for astarte DB manipulation in parallel
- Splits tests in separate files based on funcionality testing: one to test interfaces, one for triggers, and one for trigger policies, removing the old single `new_engine_test` test.

PR #1212, opened 2025-04-16
test(RM): test realm and jwt functionalities in RM

lusergit:
Based on #1211 

Adds property based testing for realm and jwt functionalities in Realm management:
- RM shoud be able to retrieve and update the `JWT PEM` certificate
- RM should be able to retrieve the `device_registration_limit` and `datastream_maximum_storage_retention` of a realm

### DIFF
--- a/apps/astarte_realm_management/test/astarte_realm_management/jwt_test.exs
+++ b/apps/astarte_realm_management/test/astarte_realm_management/jwt_test.exs
@@ -1,0 +1,45 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Astarte.RealmManagement.JwtTest do
+  alias Astarte.RealmManagement.Engine
+
+  use Astarte.RealmManagement.DataCase, async: true
+  use ExUnitProperties
+
+  describe "Test JWT PEM" do
+    @describetag :jwt
+    setup %{realm: realm} do
+      {:ok, jwt} = Engine.get_jwt_public_key_pem(realm)
+      %{jwt: jwt}
+    end
+
+    test "is correctly updated and fetched", %{realm: realm, jwt: old_jwt} do
+      new_jwt =
+        :ascii
+        |> StreamData.string()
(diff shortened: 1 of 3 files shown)


apps/astarte_realm_management/test/astarte_realm_management/jwt_test.exs
```diff
@@ -0,0 +1,41 @@
...
+      {:ok, old_jwt} = Engine.get_jwt_public_key_pem(realm)
+
+      new_jwt = "not_exactly_a_jwt_but_will_do"
+      :ok = Engine.update_jwt_public_key_pem(realm, new_jwt)
+
+      {:ok, jwt} = Engine.get_jwt_public_key_pem(realm)
+
+      assert new_jwt == jwt
+
+      _ = Engine.update_jwt_public_key_pem(realm, old_jwt)
```
Annopaolo, line 38: Since this is running in a separate keyspace (as there's  the `Astarte.RealmManagement.DataCase`) and there are no other tests in this case, could we ignore this teardown step?

lusergit: Yess

Annopaolo (approved):

apps/astarte_realm_management/test/support/helpers/database.ex
```diff
@@ -224,6 +226,12 @@ defmodule Astarte.Test.Helpers.Database do
   -----END PUBLIC KEY-----
   """
 
+  @insert_device_registration_limit """
+      INSERT INTO :keyspace.realms
+      (realm_name, device_registration_limit)
+      VALUES (:realm, :limit)
+  """
+
```
noaccOS, line 234: where is this used?

apps/astarte_realm_management/test/astarte_realm_management/jwt_test.exs
```diff
@@ -0,0 +1,40 @@
...
+    test "is correctly updated and fetched", %{realm: realm} do
+      {:ok, old_jwt} = Engine.get_jwt_public_key_pem(realm)
+
+      new_jwt = "not_exactly_a_jwt_but_will_do"
+      :ok = Engine.update_jwt_public_key_pem(realm, new_jwt)
+
+      {:ok, jwt} = Engine.get_jwt_public_key_pem(realm)
+
+      assert new_jwt == jwt
+    end
```
noaccOS, line 38: The `old_jwt` ís not used, and shouldn't really be part of the _test_.
We also have no guarantee this is not equal to `new_jwt`

I suggest moving things up like so

```suggestion
    setup do
      {:ok, jwt} = Engine.get_jwt_public_key_pem(realm)
      %{jwt: jwt}    
    end

    test "is correctly updated and fetched", %{realm: realm, jwt: old_jwt} do
      new_jwt = StreamData.string(:ascii) |> StreamData.filter(& &1 != old_jwt) |> Enum.at(0)
      :ok = Engine.update_jwt_public_key_pem(realm, new_jwt)

      assert {:ok, ^new_jwt} = Engine.get_jwt_public_key_pem(realm)
    end
```

noaccOS (changes requested):

apps/astarte_realm_management/test/support/helpers/database.ex
lusergit: nowhere!

PR #1204, opened 2025-04-11
fix(te): incoming introspection + exposed __unknown_fields__

noaccOS:
- avoid exposing `__unknown_fields__` to mustache templates
- fix incoming introspection events with introspection_map not being handled correctly

depends on #1199

* [ ] I have read [CONTRIBUTING.md](../CONTRIBUTING.md) and [CODE_OF_CONDUCT.md](../CODE_OF_CONDUCT.md)
* [ ] I have added to [CHANGELOG.md](../CHANGELOG.md) relevant changes and any other user facing change
* [ ] I have added or ran the appropriate tests

##### Does this PR introduce a user-facing change?
* [ ] Yes
* [ ] No

#### Additional documentation e.g. usage docs, diagrams, etc.:

```docs

```

PR #1214, opened 2025-04-16
refactor(te): increase coverage through unconventional means

noaccOS:
- use `Repo` from `DataAccess` to remove untested lines
- Avoid checking `test/` and `deps/` with excoveralls (yes it did find matches in `deps/`)

depends on #1204

* [ ] Yes
* [X] No

```docs

```


apps/astarte_trigger_engine/config/config.exs
```diff
@@ -37,8 +37,6 @@ config :astarte_trigger_engine, :events_consumer, Astarte.TriggerEngine.EventsCo
 
 config :astarte_trigger_engine, :amqp_adapter, ExRabbitPool.RabbitMQ
 
-config :astarte_trigger_engine, ecto_repos: [Astarte.TriggerEngine.Repo]
-
-config :astarte_trigger_engine, Astarte.TriggerEngine.Repo, []
+config :astarte_trigger_engine, ecto_repos: [Astarte.DataAccess.Repo]
```
Annopaolo, line 40: Makes quite a lot of sense! But we should make this uniform in all Astarte services, so that having two different `Repo`s would be the exception (that has to be justified), not the rule.

noaccOS: That's a great take, but unlike `capture_log: true`, this is a lot of work as not all repos work quite the same way.
For instance in this PR I had to change a match on `{:ok, result}` to a `result when is_list(result)`

lusergit: :+1:  for @noaccOS on this, I think that changing it asynchronously is actually best, instead of doing one separate PR later for all astarte services. Probably in this batch of PRs to move to the new test approach all services will converge to this solution of deleting unused files and make `Repo` files obsolete

lusergit (approved):
:rocket:

Annopaolo (approved):

PR #1219, opened 2025-04-17
Hk new engine test

eddbbt:
Rationalize and expand engine test
Add new test helper

Depends on https://github.com/astarte-platform/astarte/pull/1210


apps/astarte_housekeeping/test/astarte_housekeeping/engine_test.exs
```diff
@@ -20,26 +20,120 @@ defmodule Astarte.Housekeeping.EngineTest do
...
+      assert {:error,
+              {:invalid_replication,
+               "replication_factor 10 is >= 1 nodes in datacenter datacenter1"}} =
+               Engine.create_realm("testrealm", "test1publickey", 10, 1, 1, [])
+    end
 
-    Queries.initialize_database()
+    test "async creations returns ok" do
+      on_exit(fn ->
+        Process.sleep(1000)
```
noaccOS, line 73: sleeps are an automatic smell in tests. This test could arguably not even be needed as that functionality should already be tested in queries tests, but still.
one way to go about this is by placing an expectation for `Queries.do_delete_realm`, so that we make sure it is called once (and we may block until it is called)

This is not a blocker for this PR as the Queries tests have already been merged, but could be a nice improvement!

apps/astarte_housekeeping/test/astarte_housekeeping/engine_test.exs
```diff
@@ -20,26 +20,120 @@ defmodule Astarte.Housekeeping.EngineTest do
...
+        Database.destroy_test_keyspace!(:xandra, "testrealm")
+      end)
+
+      assert :ok = Engine.create_realm("testrealm", "test1publickey", 1, 1, 1, async: true)
+    end
+
+    test "creations returns an error" do
```
noaccOS, line 80: test names should give insight on what went wrong

```suggestion
    test "returns an error in case of database error" do
```

so that the whole error message (with the describe) appears as

`creation returns an error in case of database error`

apps/astarte_housekeeping/test/astarte_housekeeping/engine_test.exs
```diff
@@ -20,26 +20,120 @@ defmodule Astarte.Housekeeping.EngineTest do
...
+        Database.destroy_test_astarte_keyspace!(:xandra)
+      end)
+
+      Queries.initialize_database()
+    end
+
     test "succeeds when realm exists and valid update values are given" do
-      :ok = Queries.create_realm(@realm1, "test1publickey", 1, 1, 1, [])
+      on_exit(fn ->
+        Database.destroy_test_keyspace!(:xandra, @realm1)
```
noaccOS, line 133: here we're using `@realm1` instead of "testrealm", which is weird, but my point of generating it with the generator remains

apps/astarte_housekeeping/test/astarte_housekeeping/engine_test.exs
```diff
@@ -20,26 +20,120 @@ defmodule Astarte.Housekeeping.EngineTest do
...
+      end)
+
+      :ok = Engine.create_realm(@realm1, "test1publickey", 1, 1, 1, [])
```
noaccOS, line 136: as this `describe` is testing for realm update, I would expect a realm to be created in the setup process, so that the tests only _update_ this realm

apps/astarte_housekeeping/test/astarte_housekeeping/engine_test.exs
```diff
@@ -20,26 +20,120 @@ defmodule Astarte.Housekeeping.EngineTest do
...
+      end)
+
+      assert :ok = Engine.create_realm("testrealm", "test1publickey", 1, 1, 0, [])
+    end
+
+    test "creations returns error with bigger than nodes replication factor" do
+      assert {:error,
+              {:invalid_replication,
+               "replication_factor 10 is >= 1 nodes in datacenter datacenter1"}} =
+               Engine.create_realm("testrealm", "test1publickey", 10, 1, 1, [])
```
noaccOS, line 68: this may be too strict of an assertion: it relies on the fact that we're creating a keyspace with replication 1

something like
```suggestion
      assert {:error, {:invalid_replication, error_message}} =
               Engine.create_realm("testrealm", "test1publickey", 10, 1, 1, [])
      assert error_message =~ ">="
```
 may be enough

apps/astarte_housekeeping/test/astarte_housekeeping/engine_test.exs
```diff
@@ -20,26 +20,120 @@ defmodule Astarte.Housekeeping.EngineTest do
...
   doctest Astarte.Housekeeping.Engine
 
-  alias Astarte.Housekeeping.DatabaseTestHelper
   alias Astarte.Housekeeping.Engine
   alias Astarte.Housekeeping.Queries
   alias Astarte.RPC.Protocol.Housekeeping.UpdateRealm
   alias Astarte.Housekeeping.Helpers.Database
-
+  use Mimic
+  setup :set_mimic_private
```
noaccOS, line 28: this is redundant as it is the default behavior

noaccOS (approved):
The tests are a little disorganized, and a lot should be randomized and/or be put in a setup block

Still, the tests themselves look good!

apps/astarte_housekeeping/test/test_helper.exs
```diff
@@ -1,3 +1,5 @@
 Mimic.copy(Xandra)
 Mimic.copy(Xandra.Cluster)
 ExUnit.start(capture_log: true)
+Mimic.copy(Astarte.Housekeeping.Queries)
+ExUnit.start()
```
Annopaolo, line 5: Repeated `ExUnit.start()` from line 3, let's keep the one with `capture_log: true`

Annopaolo (changes requested):
GG, just a small detail

Annopaolo (approved):

PR #1223, opened 2025-04-17
test(RM): clean test files from old test approach

lusergit:
Based on #1228 

Removes old unnecessary files from the `test` section of `RealmManagement`:
- Removes old `database_test_helper`
- Removes old `engine_test`
- Removes old `queries_test`
- Removes old `database_fixtures`

Moreover moves the `Repo` management to `DataAccess`, removing the `RealmManagement.Repo` file and increasing overall coverage of the project


apps/astarte_realm_management/lib/astarte_realm_management/config.ex
```diff
@@ -42,11 +42,5 @@ defmodule Astarte.RealmManagement.Config do
   @doc """
   Returns Cassandra nodes formatted in the Xandra format.
   """
-  defdelegate xandra_nodes, to: DataAccessConfig
-  defdelegate xandra_nodes!, to: DataAccessConfig
-
   defdelegate xandra_options!, to: DataAccessConfig
-
-  defdelegate astarte_instance_id!, to: DataAccessConfig
-  defdelegate astarte_instance_id, to: DataAccessConfig
```
Annopaolo, line 51: As agreed upon IRL, let's make really sure that this is not called anymore before removing it

Annopaolo (changes requested):

Annopaolo:
Apart from one nit, looks very good to me!
image

OmarBrbutovic (approved):

apps/astarte_realm_management/lib/astarte_realm_management/config.ex
lusergit: Actually upon further inverstigation this does not seem to happen :+1: 

Apparently `astarte_instance_id` is only called from inside `DataAccess`, `RealmManagement` does not call it directly

PR #1216, opened 2025-04-16
test(RM): test device deletion

lusergit:
Tests device deletion in Realm Management `Engine` module with property based testing. The actual deletion is carried out by another module (`DeviceRemover`) hence here the only thing to test is that RM correctly puts a new entry in the `deletion_in_progress` table in the DB, so that the device remover can then query the database and handle the deletion asyncronously (with DUP). Tests of that module will appear in another PR.

PR #1228, opened 2025-04-22
test(RM): test device remover core logic and scheduler

lusergit:
Based on #1216 

Adds property based testing for the device deletion:
- Moves the device remover logic to a new `Core` module so that functions in that module can be tested separately
- Tests the scheduler to call the device remover in a separate supervised task
- Removes cumbersome tests in the device remover


apps/astarte_realm_management/lib/astarte_realm_management/migrations/create_datastream_individual_multi_interface.ex
```diff
@@ -27,8 +27,8 @@ defmodule Astarte.RealmManagement.Migrations.CreateDatastreamIndividualMultiInte
       add :interface_id, :uuid, primary_key: true
       add :endpoint_id, :uuid, primary_key: true
       add :path, :varchar, primary_key: true
-      add :value_timestamp, :timestamp, partition_key: true
-      add :reception_timestamp, :timestamp, partition_key: true
+      add :value_timestamp, :utc_datetime_usec, partition_key: true
+      add :reception_timestamp, :utc_datetime_usec, partition_key: true
```
Annopaolo, line 31: As agreed upon IRL, let's move this to another commit

apps/astarte_realm_management/test/astarte_realm_management/device_removal/core_test.exs
```diff
@@ -0,0 +1,242 @@
...
+        %IndividualDatastream{
+          device_id: device.device_id,
+          interface_id: interface.interface_id,
+          endpoint_id: endpoint_id,
+          path: "/path/",
+          value_timestamp: value_timestamp,
+          reception_timestamp: reception_timestamp,
+          reception_timestamp_submillis: reception_timestamp_submillis
+        }
+        |> Repo.insert!(prefix: keyspace)
```
Annopaolo, line 69: As agreed upon IRL, it would be great to test this by seeding the table with ([list of](https://hexdocs.pm/stream_data/StreamData.html#list_of/1)) fully generated data, so that we're actually testing the property.

apps/astarte_realm_management/test/astarte_realm_management/device_removal/core_test.exs
```diff
@@ -0,0 +1,242 @@
...
+
+        %IndividualProperty{
+          device_id: device.device_id,
+          interface_id: interface.interface_id,
+          endpoint_id: endpoint_id,
+          path: "/example/path"
+        }
+        |> Repo.insert!(prefix: keyspace)
+
+        Core.delete_individual_properties!(realm, device.device_id)
```
Annopaolo, line 94: See above

apps/astarte_realm_management/test/astarte_realm_management/device_removal/core_test.exs
```diff
@@ -0,0 +1,242 @@
...
+          value_timestamp,
+          reception_timestamp,
+          reception_timestamp_submillis
+        ])
+
+        %Device{
+          device_id: device.device_id,
+          introspection: %{interface.name => interface.major_version}
+        }
+        |> Repo.insert!(prefix: keyspace)
```
Annopaolo, line 140: Nit: as agreed upon IRL, consider moving this (and similar) to a `defp` so that the scope of the test is clearer

apps/astarte_realm_management/test/astarte_realm_management/device_removal/core_test.exs
```diff
@@ -0,0 +1,242 @@
...
+      check all(
+              device_alias <- string(:utf8, length: 1..100),
+              object_type <- integer(0..256),
+              device <- Astarte.Core.Generators.Device.device(interfaces: [])
+            ) do
+        %Name{
+          object_uuid: device.device_id,
+          object_name: device_alias,
+          object_type: object_type
+        }
```
Annopaolo, line 164: Consider moving this to a generator (even just a private one with `defp`) so that the test is clearer, as agreed upon IRL, It also holds for other data to be seeded

apps/astarte_realm_management/test/astarte_realm_management/device_removal/scheduler_test.exs
```diff
@@ -0,0 +1,87 @@
...
+                                 [restart: :transient] ->
+        assert %DeletionInProgress{
+                 device_id: ^device_id,
+                 vmq_ack: true,
+                 dup_start_ack: true,
+                 dup_end_ack: true
+               } = deletion_in_progress
+
+        :ok
+      end)
```
Annopaolo, line 81: As agreed upon IRL, this is the interesting part of the test, while the rest is just checking that OTP's basic functionalities work. So let's just test the query that lists devices to be deleted, and trust OTP for the rest

Annopaolo (changes requested):

apps/astarte_realm_management/test/astarte_realm_management/device_removal/core_test.exs
```diff
@@ -0,0 +1,242 @@
...
+        all: true
+      )
+
+      check all(
+              interface <- Astarte.Core.Generators.Interface.interface(),
+              device <- Astarte.Core.Generators.Device.device(interfaces: interface),
+              endpoint_id = Ecto.UUID.bingenerate(),
+              value_timestamp = DateTime.utc_now(),
+              reception_timestamp = DateTime.utc_now(),
+              reception_timestamp_submillis <- integer(0..10)
```
noaccOS, line 58: this is a bit weird, you're property testing on the submillis but not on the timestamp
I can see why you wouldn't need the timestamps to be property tested but doing it halfway seems a bit weird
You could have an helper function split the submillis from the reception_timestamp, or alternatively generate the reception timestamp using `Astarte.Common.Generators.Timestamp`

apps/astarte_realm_management/test/astarte_realm_management/device_removal/core_test.exs
noaccOS, line 140: nit: this is all part of the setup, it would be great having this in an helper function to avoid polluting the test _too_ too much

apps/astarte_realm_management/test/astarte_realm_management/device_removal/core_test.exs
```diff
@@ -0,0 +1,242 @@
...
+
+    property "delete_aliases/2 removes all aliases of a valid device", %{
+      realm: realm
+    } do
+      keyspace = Realm.keyspace_name(realm)
+
+      check all(
+              device_alias <- string(:utf8, length: 1..100),
+              object_type <- integer(0..256),
+              device <- Astarte.Core.Generators.Device.device(interfaces: [])
```
noaccOS, line 158:
```suggestion
              device_id <- Astarte.Core.Generators.Device.id()
```

apps/astarte_realm_management/test/astarte_realm_management/device_removal/scheduler_test.exs
```diff
@@ -0,0 +1,87 @@
...
+
+      send(scheduler, :delete_devices)
+      Process.sleep(100)
```
noaccOS, line 84:
```suggestion
      test_process = self()

      Task.Supervisor
      |> allow(self(), scheduler)
      |> expect(:start_child, fn Astarte.RealmManagement.DeviceRemoverSupervisor,
                                 DeviceRemover,
                                 :run,
                                 [deletion_in_progress],
                                 [restart: :transient] ->
        assert %DeletionInProgress{
                 device_id: ^device_id,
                 vmq_ack: true,
                 dup_start_ack: true,
                 dup_end_ack: true
               } = deletion_in_progress
        
        send(test_process, :device_deleted)

        :ok
      end)

      send(scheduler, :delete_devices)
      assert_receive :device_deleted
```

noaccOS (changes requested):

apps/astarte_realm_management/lib/astarte_realm_management/migrations/create_datastream_individual_multi_interface.ex
lusergit: This is no longer necessary as @noaccOS pointed out in [the comment below](https://github.com/astarte-platform/astarte/pull/1228#discussion_r2055677549)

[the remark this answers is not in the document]
lusergit: Yeah you're right, I think it was a temporary misconfiguration of my local DB for testing

apps/astarte_realm_management/test/astarte_realm_management/device_removal/scheduler_test.exs
lusergit: As agreed upon IRL discussion the shceduler remover will only test the relevant logic contained in it, i.e., it will only test that the `retrieve_devices_to_delete/1` function actually returns all devices with an `ack`ed deletion in progress structure

apps/astarte_realm_management/codecov.yml
```diff
@@ -0,0 +1,8 @@
+ignore:
+  - "test"
+  - "deps"
+  - "_build"
+  # Ignoring since they do not contain much logic, the acual logic is separated
+  # and tested individually
+  - "lib/astarte_realm_management/device_removal/scheduler.ex"
+  - "lib/astarte_realm_management/device_removal/device_remover.ex"
```
noaccOS, line 8: I don't agree with this change: those files are still part of your program, if they're simple enough that they don't _need_ testing (right now) it's fine but that means they should account for very little coverage decrease and it **should** be evident

apps/astarte_realm_management/coveralls.json
```diff
@@ -2,6 +2,8 @@
     "skip_files": [
         "test",
         "deps",
-        "_build"
+        "_build",
+        "lib/astarte_realm_management/device_removal/scheduler.ex",
+        "lib/astarte_realm_management/device_removal/device_remover.ex"
```
noaccOS, line 7: same as above

apps/astarte_realm_management/test/support/generators/deletion_in_progress.ex
```diff
@@ -0,0 +1,49 @@
...
+  @moduledoc """
+  Generator for `Astarte.DataAccess.Device.DeletionInProgress`
+  """
+  alias Astarte.DataAccess.Device.DeletionInProgress
+  alias Astarte.Generators.Utilities.ParamsGen
+
+  use ExUnitProperties
+  import ParamsGen
+
+  def deletion_in_progress(device_id_gen, params \\ []) do
```
noaccOS, line 31: I believe we have decided not to add structured generators to module tests: all generators should eventually live inside `astarte_generators`, and for this reason **if** you need to take a shortcut you can do so but it should be an ugly helper function

noaccOS (changes requested):

Annopaolo (approved):

PR #1229, opened 2025-04-22
test(RM API): Expand and refactor trigger policy testing

OmarBrbutovic:

- Improve testing coverage for realm management api, by removing unused trigger policy view as it is actually handled by `trigger_policy_already_present` and not `already_installed_policy`
- Fix extension mistake in `trigger_policy_controller_test.ex` file as it was not being executed since it was named with `.ex` extension instead of `.exs`
- Add property based testing for trigger policies
- Refactor context tests to test functions in isolation
- Update `execute_rpc` function to properly support synchronous calls

* [ ] Yes
* [x] No

### DIFF
--- a/apps/astarte_realm_management_api/test/astarte_realm_management_api/triggers/policies/policy_test.exs
+++ b/apps/astarte_realm_management_api/test/astarte_realm_management_api/triggers/policies/policy_test.exs
@@ -1,7 +1,7 @@
 #
 # This file is part of Astarte.
 #
-# Copyright 2021 Ispirata Srl
+# Copyright 2021 - 2025 SECO Mind Srl
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


apps/astarte_realm_management_api/test/astarte_realm_management_api/triggers/policies/policy_test.exs
```diff
@@ -39,46 +43,140 @@ defmodule Astarte.RealmManagement.API.Triggers.Policies.PolicyTest do
...
+                                  } ->
+              on_map =
+                case on do
+                  %ErrorKeyword{keyword: keyword} -> %{on: keyword}
+                  %ErrorRange{error_codes: codes} -> %{on: codes}
+                end
+
+              Map.put(on_map, :strategy, strategy)
+            end)
+          end)
```
noaccOS, line 67: This is fine as it is now (although maybe an helper function would have made this cleaner), but I think for the future it would be best to have some kind of logic in Astarte Generators to cover for this issue.
We already have code there to do exactly this, which is used internally for testing, but I think what we should actually do is expose a generator of valid changeset parameters and not only a complete struct, as I think there may be many more instances where we need to test functions which pass the parameters to the changesets

emirh1 (approved):
lgtm

Annopaolo (approved):

PR #1252, opened 2025-05-07
test(appengine): update_interface_value setups

noaccOS:
cases and helpers for interface values tests

generators are also added, which can temporarily be used by multiple tests
in appengine before being moved to astarte_generators

~~depends on https://github.com/astarte-platform/astarte/pull/1250~~

* [ ] Yes
* [X] No

```docs

```

### DIFF
--- a/apps/astarte_appengine_api/test/support/cases/device.ex
+++ b/apps/astarte_appengine_api/test/support/cases/device.ex
@@ -1,0 +1,87 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Cases.Device do
+  alias Astarte.Core.Generators.Device, as: DeviceGenerator
+  alias Astarte.Core.Generators.Interface, as: InterfaceGenerator
+  use ExUnit.CaseTemplate
+  use ExUnitProperties
+
+  import Astarte.Helpers.Device
+
+  setup_all %{realm_name: realm_name} do
+    interfaces = interfaces_for_update()
+    device = DeviceGenerator.device(interfaces: interfaces) |> Enum.at(0)
+
+    insert_device_cleanly(realm_name, device, interfaces)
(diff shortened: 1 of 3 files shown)


apps/astarte_appengine_api/test/support/cases/device.ex
```diff
@@ -0,0 +1,96 @@
...
+      :string,
+      :binaryblob,
+      :doublearray,
+      :integerarray,
+      :booleanarray,
+      :longintegerarray,
+      :stringarray,
+      :binaryblobarray,
+      :datetimearray
+    ])
```
OmarBrbutovic, line 94: Since the `Device Helper` module already includes a function named `fallible_value_types` that returns this list of types, it would be more efficient and maintainable to call that function instead of duplicating the list here.

apps/astarte_appengine_api/test/support/interface_update_generators.ex
```diff
@@ -0,0 +1,390 @@
...
+      end
+
+    value_types =
+      interface.mappings
+      |> Map.new(fn mapping ->
+        key = String.replace_prefix(mapping.endpoint, prefix_with_slash, "")
+        value_type = mapping.value_type
+
+        {key, value_type}
+      end)
```
OmarBrbutovic, line 47: The PR looks solid overall! One potential area for improvement could be extracting some common logic into private helper functions to enhance readability and maintainability. For instance, the following function could encapsulate the repeated logic for creating `value_types`:

```elixir
defp extract_value_types(interface, prefix_with_slash) do
  interface.mappings
  |> Map.new(fn mapping ->
    key = String.replace_prefix(mapping.endpoint, prefix_with_slash, "")
    value_type = mapping.value_type
    {key, value_type}
  end)
end
```

That said, there’s a `TODO` comment about moving these generators to `astarte_generators`. It’s worth considering whether this refactoring should be done now or deferred until the code is relocated. Either way, introducing such a helper function would be a valuable improvement when the time is right.

OmarBrbutovic (approved):

eddbbt (approved):

apps/astarte_appengine_api/test/support/interface_update_generators.ex
noaccOS: I've extracted the logic for value types, endpoint and reliability of object interface values into their own private functions

apps/astarte_appengine_api/test/support/cases/device.ex
```diff
@@ -0,0 +1,83 @@
...
+    )
+  end
+
+  defp object_datastream(ownership) do
+    InterfaceGenerator.interface(ownership: ownership, aggregation: :object, type: :datastream)
+  end
+
+  defp individual_datastream(ownership) do
+    InterfaceGenerator.interface(ownership: ownership, aggregation: :object, type: :datastream)
+  end
```
davidebriani, line 78:
```suggestion
  defp individual_datastream(ownership) do
    InterfaceGenerator.interface(ownership: ownership, aggregation: :individual, type: :datastream)
  end
```

davidebriani (changes requested):

davidebriani (approved):

PR #1257, opened 2025-05-08
fix(RM): upgrade data_access

lusergit:
Upgrades data access to comply with the common signature of `fetch_*` functions, refactors the necessary code to comply with the new `fetch_all` signature


apps/astarte_realm_management/lib/astarte_realm_management/queries.ex
```diff
@@ -579,14 +579,13 @@ defmodule Astarte.RealmManagement.Queries do
...
-      query
-      |> Repo.fetch_all(prefix: keyspace, consistency: consistency)
-      |> Enum.map(fn property ->
+    with {:ok, properties} <- Repo.fetch_all(query, prefix: keyspace, consistency: consistency) do
+      Enum.map(properties, fn property ->
         [endpoint_id: property.endpoint_id, path: property.path]
       end)
 
-    {:ok, properties}
+      {:ok, properties}
```
davidebriani, line 587: The `Enum.map` line seems to produce a value that is not used: you forgot to rebind `properties` to the newly computed value

PR #1258, opened 2025-05-08
chore(AE): update data_access

lusergit:
Upgrade data_access on all astarte services, the new version complies with the usual elixir `fetch_*` syntax


eddbbt (approved):

apps/astarte_trigger_engine/lib/astarte_trigger_engine/amqp_consumer/queries.ex
```diff
@@ -51,13 +51,9 @@ defmodule Astarte.TriggerEngine.AMQPConsumer.Queries do
...
-    case Repo.fetch_all(query, consistency: Consistency.domain_model(:read)) do
-      realms when is_list(realms) ->
-        {:ok, realms}
-
-      {:error, reason} ->
-        _ = Logger.warning("Could not list realms, reason: #{inspect(reason)}")
-        {:error, reason}
+    with {:error, reason} <- Repo.fetch_all(query, consistency: Consistency.domain_model(:read)) do
+      _ = Logger.warning("Could not list realms, reason: #{inspect(reason)}")
+      {:error, reason}
```
OmarBrbutovic, line 56: I’d suggest rewording this a bit to make it easier to read and understand at a glance. Your implementation works fine, but at first glance, it looks like it only returns errors 😅, until you read it more carefully. Something like this might help:

```elixir
with {:ok, realms} <- Repo.fetch_all(query, consistency: Consistency.domain_model(:read)) do
  {:ok, realms}
else
  {:error, reason} ->
    Logger.warning("Could not list realms, reason: #{inspect(reason)}")
    {:error, reason}
end
```

OmarBrbutovic (approved):

apps/astarte_trigger_engine/lib/astarte_trigger_engine/amqp_consumer/queries.ex
noaccOS: This is not the only instance in Astarte (or our products) where we use `with` to match on an error, and imo it is very readable as it is

Moreover, if we were to use an approach like the one you suggested, I'd say that it would need to be a `case` statement, as `with` blocks only make sense when matching multiple clauses sequentially (yes I'm aware we have `with` blocks with only one match in astarte but that doesn't mean it's the best practice)

eddbbt (approved):

davidebriani (approved):

PR #1261, opened 2025-05-09
WIP Test/ae interface reading test

eddbbt:


apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -104,7 +104,33 @@ defmodule Astarte.Helpers.Device do
...
+  def valid_result?(result, interface, value) do
+    similar?(result, value)
+  end
+
   defp similar?(nil = _result, [] = _value), do: true
+  defp similar?("" = _result, nil = _value), do: true
+  defp similar?(result, result), do: true
+  defp similar?(result, [result]), do: true
+  defp similar?([result], [result]), do: true
+  defp similar?([result], result), do: true
```
noaccOS, line 120: these cases should all already be covered

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -104,7 +104,33 @@ defmodule Astarte.Helpers.Device do
...
+
+  defp similar?(%{"" => nil} = _result, [] = _value), do: true
```
noaccOS, line 122: what is this

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -104,7 +104,33 @@ defmodule Astarte.Helpers.Device do
...
+
+  defp similar?([~U[1970-01-01 00:00:00.000Z]], [0]), do: true
+  defp similar?(~U[1970-01-01 00:00:00.000Z], 0), do: true
+  defp similar?([~U[1970-01-01 00:00:00.000Z]], 0), do: true
+  defp similar?(~U[1970-01-01 00:00:00.000Z], [0]), do: true
```
noaccOS, line 127:
```suggestion
  defp similar?(datetime, timestamp) when is_struct?(datetime, DateTime) and is_integer(timestamp), do: DateTime.to_unix(datetime) == timestamp
```
might need to double check the unit

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -104,7 +104,33 @@ defmodule Astarte.Helpers.Device do
...
+  defp similar?([%DateTime{} = date], [date2]), do: DateTime.to_string(date) == date2
+  defp similar?(%DateTime{} = date, [date2]), do: DateTime.to_string(date) == date2
+  defp similar?([%DateTime{} = date], date2), do: DateTime.to_string(date) == date2
+  defp similar?(%DateTime{} = date, date2), do: DateTime.to_string(date) == date2
```
noaccOS, line 132: already covered at line 144

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -104,7 +104,33 @@ defmodule Astarte.Helpers.Device do
...
+  defp similar?([%{"timestamp" => _, "value" => ~U[1970-01-01 00:00:00.000Z]}], 0), do: true
+  defp similar?([%{"timestamp" => _, "value" => value}], value), do: true
+  defp similar?(%{"timestamp" => _, "value" => value}, value), do: true
+
+  defp similar?(%{"timestamp" => _, "value" => result}, value) when is_binary(result),
+    do: result == to_string(value)
```
noaccOS: already covered at line 147

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+        astarte_instance_id: astarte_instance_id,
+        interfaces: interfaces,
+        device: device
+      } = context
+
+      clean_devices_saved_data(astarte_instance_id, realm_name)
+
+      on_exit(fn ->
+        clean_devices_saved_data(astarte_instance_id, realm_name)
+      end)
```
noaccOS, line 45: the rule should be that tests leave the state clean, so each test can assume to find a clean state

I suggest adding this to a setup

```elixir
setup context do
  on_exit(fn -> clean_device_saved_data(context.astarte_instance_id, context.realm_name) end)
end
```

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+        {:ok, %InterfaceValues{data: result}} =
+          Device.get_interface_values!(
+            realm_name,
+            device.encoded_id,
+            interface_to_update.name,
+            %{}
+          )
+
+        result = get_in(result, read_path)
+        assert valid_result?(result, interface_to_update, expected_read_value)
```
noaccOS, line 91: how does this test differ from [this one](https://github.com/astarte-platform/astarte/blob/d3aebdef64cc331f8e7a3cbf0e2fe0ac2c5c9199/apps/astarte_appengine_api/test/astarte_appengine_api/device/device_v2_test.exs#L30)?

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+        clean_devices_saved_data(astarte_instance_id, realm_name)
+      end)
+
+      valid_interfaces_for_update = interfaces |> Enum.filter(&(&1.ownership == :server))
+
+      check all(
+              interface_to_update <- member_of(valid_interfaces_for_update),
+              mapping_update <- valid_mapping_update_for(interface_to_update)
+            ) do
+        update_value = nil
```
noaccOS, line 115: unused?

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+          Device.get_interface_values!(
+            realm_name,
+            device.encoded_id,
+            interface_to_update.name,
+            %{}
+          )
+
+        result = get_in(result, read_path)
+
+        assert valid_result?(result, interface_to_update, nil)
```
noaccOS, line 133: if you don't insert data, isn't the `nil` value from the `get_in` and not necessarily from astarte?

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+            interface_to_update.name,
+            %{}
+          )
+
+        result = get_in(result, read_path)
+        assert valid_result?(result, interface_to_update, expected_read_value)
+      end
+    end
+
+    property "returns nil", context do
```
noaccOS, line 95: this does not explain the behavior
A better `property` would be "returns nil when the value is not present"

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+
+      check all(
+              interface_to_update <- member_of(valid_interfaces_for_update),
+              mapping_update <- valid_mapping_update_for(interface_to_update)
+            ) do
+        update_value = nil
+        path_tokens = String.split(mapping_update.path, "/")
+        expected_token = [realm_name, device.encoded_id, interface_to_update.name | path_tokens]
+
+        expected_qos = expected_qos_for!(mapping_update)
```
noaccOS, line 119: unused?

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+
+      valid_interfaces_for_update = interfaces |> Enum.filter(&(&1.ownership == :server))
+
+      check all(
+              interface_to_update <- member_of(valid_interfaces_for_update),
+              mapping_update <- valid_mapping_update_for(interface_to_update)
+            ) do
+        update_value = nil
+        path_tokens = String.split(mapping_update.path, "/")
+        expected_token = [realm_name, device.encoded_id, interface_to_update.name | path_tokens]
```
noaccOS, line 117: unused?

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+        update_value = mapping_update.value
+        path_tokens = String.split(mapping_update.path, "/")
+        expected_token = [realm_name, device.encoded_id, interface_to_update.name | path_tokens]
+
+        expected_published_value =
+          expected_published_value!(mapping_update.value_type, update_value)
+
+        expected_qos = expected_qos_for!(mapping_update)
+
+        expected_read_value = update_value
```
noaccOS, line 166:
```suggestion
        expected_read_value = expected_read_value!(update_value)
```

this was probably your issue with the many extra `similar?` instances

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+        {:ok, %InterfaceValues{data: result}} =
+          Device.get_interface_values!(
+            realm_name,
+            device.encoded_id,
+            interface_to_update.name,
+            mapping_update.path,
+            %{limit: 1}
+          )
+
+        # result = get_in(result, mapping_update.path)
```
noaccOS, line 195: leftover

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+        assert valid_result?(result, interface_to_update, expected_read_value)
+      end
+    end
+
+    property "returns with limit", context do
```
noaccOS, line 200: what does this mean? that it respects the limit opt?

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+        clean_devices_saved_data(astarte_instance_id, realm_name)
+      end)
+
+      valid_interfaces_for_update = interfaces |> Enum.filter(&(&1.ownership == :server))
+
+      check all(
+              interface_to_update <- member_of(valid_interfaces_for_update),
+              mapping_update <- valid_mapping_update_for(interface_to_update)
+            ) do
+        limit_n = integer(0..400) |> Enum.at(0)
```
noaccOS, line 220: should this maybe be part of the `check all` block?

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+        {:ok, %InterfaceValues{data: result}} =
+          Device.get_interface_values!(
+            realm_name,
+            device.encoded_id,
+            interface_to_update.name,
+            mapping_update.path,
+            %{limit: limit_n}
+          )
+
+        excluded_values = [[], [0], [""], 0, "", ["1970-01-01T00:00:00Z"], "1970-01-01T00:00:00Z"]
```
noaccOS, line 265: ???

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+
+        result =
+          cond do
+            is_list(result) && mapping_update.reliability != :unique ->
```
noaccOS, line 269: I'm not sure how the reliability matters in this case?

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+              assert length(result) == limit_n
+              [first | _] = result
+              first
+
+            true ->
+              if is_list(result) do
+                result |> Enum.at(0)
+              else
+                result
+              end
```
noaccOS, line 279: this is literally why `valid_result?` exists, you shouldn't need to manually check it

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
noaccOS: I'm honestly not sure what this is testing... if it's just the read option, you should pre-populate your data in a setup?

noaccOS (changes requested):

apps/astarte_appengine_api/test/support/helpers/device.ex
eddbbt: a kind of valid case not covered yet, i'm not sure how or why the generators created this but this also apply to most of the cases added

I agree with you, doesn't seems too right

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+        astarte_instance_id: astarte_instance_id,
+        interfaces: interfaces,
+        device: device
+      } = context
+
+      clean_devices_saved_data(astarte_instance_id, realm_name)
+
+      on_exit(fn ->
+        clean_devices_saved_data(astarte_instance_id, realm_name)
+      end)
```
Annopaolo, line 107: Isn't this in the `setup` block already?

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+
+    property "returns nil", context do
+      %{
+        realm_name: realm_name,
+        astarte_instance_id: astarte_instance_id,
+        interfaces: interfaces,
+        device: device
+      } = context
+
+      clean_devices_saved_data(astarte_instance_id, realm_name)
```
Annopaolo, line 103: As @noaccOS said, tests should leave db in a clean state, so doing this here is discouraged

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+
+      on_exit(fn ->
+        clean_devices_saved_data(astarte_instance_id, realm_name)
+      end)
```
Annopaolo, line 149: See comments above

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,299 @@
...
+    case mapping_update.reliability do
+      :unreliable -> 0
+      :guaranteed -> 1
+      :unique -> 2
+    end
+  end
+
+  defp clean_devices_saved_data(astarte_instance_id, realm_name) do
+    Repo.query!("TRUNCATE #{astarte_instance_id <> realm_name}.individual_properties")
+    Repo.query!("TRUNCATE #{astarte_instance_id <> realm_name}.individual_datastreams")
```
Annopaolo, line 297: The keyspace name might not always be just instance id and realm name juxtaposed, so let's stick to use `Realm.keyspace_name(realm_name)` from DataAccess

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -104,7 +108,64 @@ defmodule Astarte.Helpers.Device do
     Enum.any?(result, &valid_result?(&1, interface, value))
   end
 
+  def valid_result?(result, interface, value) do
+    similar?(result, value)
+  end
```
Annopaolo, line 113: I would argue that there should be different tests for each of this cases, rather than trying to fit so-called similar values in a single result. We should know that e.g. some interfaces return a list and other an object, for instance. This holds for all `similar?/2` heads

apps/astarte_appengine_api/test/support/helpers/device.ex
eddbbt: I'm probably missing something, but looks like they're not 

valid_result?/3 has guards only for  is_map(result) and is_map(value), not for just one of them

```
 Expected truthy, got false
     code: assert valid_result?(result, interface_to_update, expected_read_value)
     arguments:

         # 1
         false

         # 2
         %Astarte.Core.Interface{
           interface_id: <<59, 211, 53, 211, 248, 56, 71, 88, 141, 237, 32, 150, 211, 247, 65, 164>>,
           name: "fM.Df.x",
           major_version: 6,
           minor_version: 248,
           type: :properties,
           ownership: :server,
           aggregation: :individual,
           description: nil,
           doc: nil,
           mappings: [
             %Astarte.Core.Mapping{
               endpoint: "/%{g}/individual_9090",
               value_type: :booleanarray,
               reliability: :unreliable,
               retention: :discard,
               expiry: 0,
               database_retention_policy: :no_ttl,
               database_retention_ttl: nil,
               allow_unset: true,
               explicit_timestamp: false,
               description: "o",
               doc: nil,
               endpoint_id: nil,
               interface_id: nil,
               path: nil,
               type: nil
             }
           ],
           quality: nil,
           aggregate: nil,
           interface_name: "fM.Df.x",
           version_major: 6,
           version_minor: 248
         }

         # 3
         [false]
```

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
eddbbt: there's no assert on the update value part, also those two tests are (logically and by tasks) superimposable

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
noaccOS: This means you don't have `Mimic`'s expectations working in your test process for `Realm.keyspace_name`

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -0,0 +1,232 @@
...
+        []
+      )
+
+    %{
+      interface_to_update: interface_to_update,
+      read_path: read_path,
+      expected_read_value: expected_read_value,
+      mapping_update: mapping_update
+    }
+  end
```
noaccOS, line 231: No hard feelings if you don't want to but now we have merged setups to pre-populate interfaces [here](https://github.com/astarte-platform/astarte/blob/e24d66829bfa8f2db76047754d92b0aa021c4bb0/apps/astarte_appengine_api/test/support/cases/device.ex#L46), so this block could be switched for a `setup_all :populate_interfaces`

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -187,6 +191,15 @@ defmodule Astarte.Helpers.Device do
     |> Enum.all?(fn {key, result_value} -> similar?(result_value, Map.fetch!(value, key)) end)
   end
 
+  def valid_result?(result, _interface, value) when is_map(result) do
+    similar?(result, value)
+  end
+
```
noaccOS, line 197: this is the same as the base case, if you're hitting something else first that's a problem

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -187,6 +191,15 @@ defmodule Astarte.Helpers.Device do
...
+  def valid_result?(result, interface, value) when is_list(result) and is_list(value) do
+    [result | _] = result
+    valid_result?(result, interface, value)
+  end
```
noaccOS, line 201: this doesn't look right. if they're both lists, I'd expect to check both for equality
still, this should be covered by current line 203

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -195,30 +208,123 @@ defmodule Astarte.Helpers.Device do
     similar?(result, value)
   end
 
-  defp similar?(nil = _result, [] = _value), do: true
+  defp similar?(nil, []), do: true
```
noaccOS, line 211: let's not change this line

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -195,30 +208,123 @@ defmodule Astarte.Helpers.Device do
...
+  defp similar?(%{} = _result, nil = _value), do: true
   defp similar?("" = _result, nil = _value), do: true
-  defp similar?(result, result), do: true
-  defp similar?(result, [result]), do: true
-  defp similar?([result], [result]), do: true
-  defp similar?([result], result), do: true
+  defp similar?(%{"" => nil}, nil), do: true
+  defp similar?(%{"" => nil}, []), do: true
```
noaccOS, line 215: still baffled on this

lusergit (changes requested):
Almost there! pay attention to the second commit description, it contains the old descriptions of all squashed commits

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -195,30 +208,123 @@ defmodule Astarte.Helpers.Device do
...
 
   defp similar?(%{"" => nil} = _result, [] = _value), do: true
+  defp similar?(nil = _result, [] = _value), do: true
```
noaccOS, line 218: this is literally line 212

apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -195,30 +208,123 @@ defmodule Astarte.Helpers.Device do
...
+         %{"reception_timestamp" => _, "timestamp" => _, "value" => [val]},
+         [val]
+       ),
+       do: true
+
+  defp similar?(
+         %{"reception_timestamp" => _, "timestamp" => _, "value" => val},
+         [val]
+       ),
+       do: true
```
noaccOS, line 266: these (and many others) are all already covered by (current) line 340
you just need to specify similar for datetimes and strings/integers (there already are some base cases specified)

noaccOS (changes requested):
Some changes in the first commit which are never used and overridden in the second
Let's try having more structured commits so that it's easier to navigate the history of the project

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
eddbbt: @noaccOS mimic expectations shall be called by Astarte.Cases.Data, that's already in use

lusergit: Yes, but `on_exit` function is called inside another process, so in your setup (line 33) you should have something like
```elixir
    import Astarte.Helpers.DatabaseV2
    # ...
    setup context do
      %{astarte_instance_id: astarte_instance_id, realm_name: realm_name} = context
      
      on_exit(fn -> 
        setup_database_access(astarte_instance_id)
        clean_device_saved_data(astarte_instance_id, realm_name) 
      end)
    end

```

apps/astarte_appengine_api/test/support/helpers/device.ex
eddbbt: I'm sorry but

```
     Expected truthy, got false
     code: assert valid_result?(result, interface_to_update, expected_read_value)
     arguments:

         # 1
         %{"" => nil}
         
         ....
         
         
         # 3
         []
```

apps/astarte_appengine_api/test/support/helpers/device.ex
eddbbt: not completely sure, but I'll try

lusergit (approved):
:rocket:

noaccOS (approved):
pepe

davidebriani (approved):

Annopaolo (approved):

lusergit (approved):

PR #1262, opened 2025-05-09
fix: make `unset_property` work

lusergit:
Astarte documentation reports an optional field of properties interfaces called `allow_unset`.This field should allow or forbid for interface value unset.
Up to now this field was not checked, and interfaces with `allow_unset: false` or `allow_unset: nil` were able to unset such properties.

This PR introduces such checks both in server owned interfaces (trough appengine) and in device owned interfaces (trough data updater plant).

Tests in both services are also added/edited accordingly:
- In appengine a deletion property is checked:  when unsetting properties with `allow_uset: true` the API should return `204` (no content) as HTML code. Instead, when `allow_unset` is `false` the server should answer with a `400` status code (bad request)
- In DUP the old _happy path_ test is updated to allow to unset the property of the `com.test.LCDMonitor` test interface

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
@@ -236,35 +236,36 @@ defmodule Astarte.AppEngine.API.Device do
             [ttl: db_max_ttl]
         end
 
-      Queries.insert_value_into_db(
-        realm_name,
-        device_id,
-        interface_descriptor,
-        endpoint_id,
-        mapping,
-        path,
-        value,
-        now,
-        opts
-      )
+      with :ok <-
+             Queries.insert_value_into_db(
+               realm_name,
+               device_id,
+               interface_descriptor,
+               endpoint_id,
+               mapping,
+               path,
+               value,
+               now,
+               opts
+             ) do
+        if interface_descriptor.type == :datastream do
+          Queries.insert_path_into_db(
+            realm_name,
+            device_id,
+            interface_descriptor,
+            endpoint_id,
+            path,
+            now,
+            now,
+            opts
+          )
+        end
 
-      if interface_descriptor.type == :datastream do
-        Queries.insert_path_into_db(
-          realm_name,
-          device_id,
-          interface_descriptor,
-          endpoint_id,
-          path,
-          now,
-          now,
-          opts
-        )
+        {:ok,
+         %InterfaceValues{
+           data: raw_value
+         }}
       end
-
-      {:ok,
-       %InterfaceValues{
-         data: raw_value
-       }}
     else
       {:error, :endpoint_guess_not_allowed} ->
         _ = Logger.warning("Incomplete path not allowed.", tag: "endpoint_guess_not_allowed")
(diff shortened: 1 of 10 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -684,24 +686,27 @@ defmodule Astarte.AppEngine.API.Device do
...
-        path,
-        nil,
-        nil,
-        []
-      )
-
-      case interface_descriptor.type do
-        :properties ->
-          unset_property(realm_name, device_id, interface, path)
+      now = DateTime.utc_now()
```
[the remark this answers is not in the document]
lusergit: It was a leftover form an old solution, removed :+1:

apps/astarte_appengine_api/lib/astarte_appengine_api_web/controllers/fallback_controller.ex
```diff
@@ -205,6 +205,13 @@ defmodule Astarte.AppEngine.APIWeb.FallbackController do
     |> render(:"422_unexpected_object_key")
   end
 
+  def call(conn, {:error, :unset_not_allowed}) do
+    conn
+    |> put_status(:bad_request)
+    |> put_view(Astarte.AppEngine.APIWeb.ErrorView)
+    |> render(:"400_unset_not_allowed")
```
Annopaolo, line 212: I would argue that this is more akin to a 422 error, as
>  The request was well-formed but was unable to be followed due to semantic errors.

([source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#client_error_responses)) In the same fashion as we do for unexpected object keys or invalid device attributes.
Moreover, it would be great to document this new error in the [API docs](https://github.com/astarte-platform/astarte/blob/release-1.2/apps/astarte_appengine_api/priv/static/astarte_appengine_api.yaml)

apps/astarte_appengine_api/lib/astarte_appengine_api_web/views/error_view.ex
```diff
@@ -82,6 +82,10 @@ defmodule Astarte.AppEngine.APIWeb.ErrorView do
     %{errors: %{detail: "Alias already in use"}}
   end
 
+  def render("400_unset_not_allowed.json", _assigns) do
```
Annopaolo, line 85: See comment above

apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/interface_controller_test.exs
```diff
@@ -217,6 +217,64 @@ defmodule Astarte.AppEngine.APIWeb.InterfaceControllerTest do
...
+          )
+
+        conn = delete(conn, request_path)
+
+        case mapping_to_unset.allow_unset do
+          false ->
+            assert response(conn, 400)
+
+          true ->
+            assert response(conn, 204)
```
Annopaolo, line 260: Nit: would it be possible to write two separate tests, one for `allow_unset: true` and one for `allow_unset: false`? This way, we clearly separate the two different flows

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -659,7 +659,40 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
...
+          }
+
+          execute_device_error_triggers(
+            new_state,
+            "unset_not_allowed",
+            error_metadata,
+            timestamp
+          )
+
+          raise "Forbidden"
```
Annopaolo, line 691: Once we've handled the message, sent the error trigger and updated the telemetry, there's no need to crash the DataUpdater process too. After all, DUP does not expose an API

Annopaolo (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
lusergit: :+1: from what I understand the execution should stop there, right? Otherwise an ack is sent and telemetry is updated for the correct processing of the message.

apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/interface_controller_test.exs
lusergit: It would be pretty difficult as we're using the device case to install a device and its interfaces and then we select one among those to perform an update, we do not generate the interfaces and devices directly here

lusergit: Done :+1:

noaccOS (approved):

eddbbt (approved):

Annopaolo (approved):

PR #1264, opened 2025-05-12
test(appengine): add `populate_interface` function

noaccOS:
this adds a function to the Device case which is used to pre-populate the database with random data for a random interface of each type

the chosen interfaces are then exposed in the context so they can be accessed by tests to verify the data.

to use this function, add `setup_all :populate_interfaces` in a module which `use`s `Cases.Device`

update values functions from Device had to be made public to avoid duplicating the code, as the pre-existing public interface would only allow server owned values to be inserted

* [ ] Yes
* [ ] No

```docs

```

### DIFF
--- a/apps/astarte_appengine_api/test/support/helpers/device.ex
+++ b/apps/astarte_appengine_api/test/support/helpers/device.ex
@@ -16,12 +16,15 @@
 # limitations under the License.
 
 defmodule Astarte.Helpers.Device do
+  alias Astarte.AppEngine.API.Device, as: Core
   alias Astarte.AppEngine.API.Device.InterfaceValue
   alias Astarte.AppEngine.API.Repo
   alias Astarte.DataAccess.Devices.Device
   alias Astarte.DataAccess.Realms.Interface
   alias Astarte.DataAccess.Realms.Realm
   alias Astarte.RealmManagement.Engine, as: RealmManagement
+  alias Astarte.Common.Generators.Timestamp, as: TimestampGenerator
+
   import ExUnit.CaptureLog
   import StreamData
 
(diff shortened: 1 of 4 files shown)


apps/astarte_appengine_api/test/support/helpers/device.ex
```diff
@@ -67,6 +69,72 @@ defmodule Astarte.Helpers.Device do
...
+              interface_descriptor,
+              &1.path,
+              &1.value
+            )
+          )
+      end
+    end)
+    |> Task.await()
+  end
+
```
OmarBrbutovic, line 137: The logic for handling `insert_value` and `insert_values` functions is duplicated. To improve maintainability and reduce redundancy, you could refactor this shared logic into a private helper function. For example:

```elixir
def insert_value(realm_name, device_id, interface_descriptor, mapping_update) do
  do_insert(realm_name, device_id, interface_descriptor, [mapping_update])
end

def insert_values(realm_name, device_id, interface_descriptor, mapping_updates) do
  do_insert(realm_name, device_id, interface_descriptor, mapping_updates)
end

defp do_insert(realm_name, device_id, interface_descriptor, mapping_updates) do
  Task.async(fn ->
    Mox.stub(Astarte.AppEngine.API.RPC.VMQPlugin.ClientMock, :publish, fn _ ->
      {:ok, %{local_matches: 1, remote_matches: 0}}
    end)

    case interface_descriptor.aggregation do
      :individual ->
        Enum.each(mapping_updates, &Core.update_individual_interface_values(realm_name, device_id, interface_descriptor, &1.path, &1.value))

      :object ->
        Enum.each(mapping_updates, &Core.update_object_interface_values(realm_name, device_id, interface_descriptor, &1.path, &1.value))
    end
  end)
  |> Task.await()
end
```

OmarBrbutovic (approved):

eddbbt (approved):

davidebriani (approved):

PR #1268, opened 2025-05-14
test(AE): test for optional `keep_milliseconds` parameter

lusergit:
Adds tests for interface update and retrieval with optional `keep_milliseconds` parameter

### DIFF
--- a/apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/interface_controller_test.exs
+++ b/apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/interface_controller_test.exs
@@ -217,6 +217,124 @@ defmodule Astarte.AppEngine.APIWeb.InterfaceControllerTest do
                )
       end
     end
+
+    test "accepts and understands `keep_milliseconds` optional parameter", context do
+      %{realm_name: realm_name, interfaces: interfaces, device: device, auth_conn: conn} = context
+
+      interface =
+        interfaces
+        |> Enum.filter(fn interface ->
+          interface.ownership == :server && Enum.all?(interface.mappings, & &1.explicit_timestamp)
+        end)
+        |> Enum.random()
+
+      mapping_update = valid_mapping_update_for(interface) |> Enum.at(0)
+      path = path_tokens(mapping_update.path)
+
+      request_path =
+        interface_values_path(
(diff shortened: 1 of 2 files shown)


emirh1 (approved):

apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/interface_controller_test.exs
```diff
@@ -217,6 +217,109 @@ defmodule Astarte.AppEngine.APIWeb.InterfaceControllerTest do
...
+          path
+        )
+
+      publish_result_ok(valid_interface, mapping_update, fn _ -> :ok end)
+
+      expected_time = DateTime.utc_now()
+
+      DateTime
+      |> allow(conn.owner, self())
+      |> expect(:utc_now, 9, fn -> expected_time end)
```
noaccOS, line 248: instead of using a "semi random" number of calls, I'd say it makes more sense to just stub it

```suggestion
      |> stub(:utc_now, fn -> expected_time end)
```

apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/interface_controller_test.exs
```diff
@@ -217,6 +217,109 @@ defmodule Astarte.AppEngine.APIWeb.InterfaceControllerTest do
...
+          device.encoded_id,
+          valid_interface.name,
+          path,
+          %{"keep_milliseconds" => true}
+        )
+
+      conn = get(conn, request_path)
+
+      assert %{"data" => [%{"timestamp" => timestamp}]} =
+               json_response(conn, 200)
```
noaccOS, line 266: looks like this assertion is failing in ci
I'm afraid you may need an helper function to discard values without the timestamp

noaccOS (commented):
A couple comments but the tests are good :rocket:

apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/interface_controller_test.exs
lusergit: I directly filter out interfaces without `explicit_timestamp`

PR #1270, opened 2025-05-19
refactor(DUP): `Core` modules

lusergit:
Initializes `Astarte.DataUpdaterPlant.DataUpdater.Core` modules to unpack `impl.ex`, initially adding `Core.Interface` module for interface related helper functions.


noaccOS (approved):
I don't think this should be the final form of DUP's code.
In my opinion, we should strive to have many independent modules, each representing a context, and not bound by `Impl`'s state

Still, this is a step forward and as it removes some clutter from Impl's module and exposing core functions makes everything more testable, so here's my approval

eddbbt (approved):

PR #1274, opened 2025-05-21
Dup async tests

eddbbt:
make dup tests async
change db initialization to accept realm
move non asynchronous test

### DIFF
--- a/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater_test.exs
+++ b/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater_test.exs
@@ -53,21 +53,16 @@ defmodule Astarte.DataUpdaterPlant.DataUpdaterTest do
 
   setup :verify_on_exit!
 
-  setup do
+  setup_all do
     realm = "autotestrealm#{System.unique_integer([:positive])}"
-    DatabaseTestHelper.destroy_local_test_keyspace(realm)
     {:ok, _keyspace_name} = DatabaseTestHelper.create_test_keyspace(realm)
 
     on_exit(fn ->
       DatabaseTestHelper.destroy_local_test_keyspace(realm)
     end)
 
-    %{realm: realm}
-  end
-
-  setup_all do
     {:ok, _pid} = AMQPTestHelper.start_link()
-    :ok
+    %{realm: realm}
   end
 
   test "simple flow", %{realm: realm} do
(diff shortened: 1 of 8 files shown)


apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater_test.exs
```diff
@@ -52,20 +53,26 @@ defmodule Astarte.DataUpdaterPlant.DataUpdaterTest do
 
   setup :verify_on_exit!
 
-  setup_all do
-    DatabaseTestHelper.destroy_local_test_keyspace()
-    {:ok, _keyspace_name} = DatabaseTestHelper.create_test_keyspace()
-    {:ok, _pid} = AMQPTestHelper.start_link()
+  setup do
```
noaccOS, line 56: why was this changed to a `setup`?

I suggest keeping it a `setup_all` instead, to avoid having to setup a keyspace for each test

still, this should be done with a `Case` and we already have a pr for that so I'll stop nitpicking

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater_test.exs
```diff
@@ -52,20 +53,26 @@ defmodule Astarte.DataUpdaterPlant.DataUpdaterTest do
...
+    realm = "autotestrealm#{System.unique_integer([:positive])}"
+    DatabaseTestHelper.destroy_local_test_keyspace(realm)
```
noaccOS, line 58: there probably is no need to destroy it beforehand anymore, as this is now cleanly assigned to each module/test

lusergit (approved):
nit: the `DeviceDeleteTest` can be made asyncronous: one can first get the message tracker and data updater using `fetch_message_tracker/2` and `fetch_data_updater_process/3` and allow `Mox` on those `pid`s, but for now I think this is good, only one test is not asyncronous, so lgtm :rocket:

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/dup_device_delete_test.exs
```diff
@@ -0,0 +1,261 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2017 - 2025 SECO Mind Srl
```
noaccOS, line 4:
```suggestion
# Copyright 2025 SECO Mind Srl
```

noaccOS (approved):
:rocket:

davidebriani (approved):

PR #1271, opened 2025-05-20
fix(appengine): crash with empty result + disjoint_tables

noaccOS:
according to the history, the author of c8994b8 (me) broke this function
by returning `[]` for empty values while the code using this result
expects a map with a "timestamp" column

the result has been changed to match the previous behavior

add tests for the happy path of `since`, `since_after`, `to` options for datastreams

Signed-off-by: Francesco Noacco <francesco.noacco@secomind.com>
* [ ] I have read [CONTRIBUTING.md](../CONTRIBUTING.md) and [CODE_OF_CONDUCT.md](../CODE_OF_CONDUCT.md)
* [ ] I have added to [CHANGELOG.md](../CHANGELOG.md) relevant changes and any other user facing change
* [ ] I have added or ran the appropriate tests

#### Which issue(s) this PR fixes:

Fixes #

##### Does this PR introduce a user-facing change?
* [ ] Yes
* [ ] No

#### Additional documentation e.g. usage docs, diagrams, etc.:

```docs

```

PR #1275, opened 2025-05-21
test(appengine): add downsampling to interface value options gen

noaccOS:
allows downsampling options if:
- an individual interface is downsampable
- an object interface's downsample keys are overridden from params

the final step would be allowing all downsampable object interfaces
without overrides from params, but currently this would mean checking
that an object key does not contain any `nil` value.

this would become easier to implement if we implemented nil handling
in exlttb

remove warnings by dropping unused alias and module attribute

~~depends on #1271~~

* [ ] Yes
* [X] No

```docs

```


lusergit (approved):
:rocket:

OmarBrbutovic (approved):

davidebriani (approved):

PR #1278, opened 2025-05-21
doc: document 422 error

lusergit:
based on #1262 , adds documentation for the new `422` error on property unset when endpoint has `allow_unset=false`


apps/astarte_appengine_api/priv/static/astarte_appengine_api.yaml
```diff
@@ -671,6 +673,8 @@ paths:
                       detail: Endpoint not found
         '405':
           description: Invalid object
+        '422':
+          description: Tried to unset a property with `allow_unset` false
```
davidebriani, line 677: If I am not mistaken, the POST method is meant for publishing datastream values.
So this code path shouldn't involve (un)setting properties, and the 422 error here is not needed.

davidebriani (changes requested):

eddbbt (approved):
lgtm

PR #1283, opened 2025-05-22
Dup repo

eddbbt:
, Remove DUP repo file
Fix return types in order to make them work with dataAccess Repo 

About row 779 of queries.ex, the whole astarte works with the idea that, if no `datastream_maximum_storage_retention` is found, the resulting return value MUST be nil, not error (old repo.fetch_one behavior, see here https://hexdocs.pm/ecto/Ecto.Repo.html#c:one/2 ) so a `with` has been put in place


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/repo.ex
```diff
@@ -1,108 +0,0 @@
...
-  end
-
-  def fetch_one(queryable, opts \\ []) do
-    try do
-      one(queryable, opts)
-    catch
-      error ->
-        handle_xandra_error(error)
-    end
-  end
```
noaccOS, line 49: unfortunately our naming is all over the place. in data access you can see that

```elixir
  def fetch_one(queryable, opts \\ []) do
    {error, opts} = Keyword.pop_first(opts, :error, :not_found)

    case all(queryable, opts) do
      [item] -> {:ok, item}
      [] -> {:error, error}
      other -> raise Ecto.MultipleResultsError, queryable: queryable, count: length(other)
    end
  end
```

although `fetch_all` is equivalent (not in its return typing tho) to the one from DUP

I believe a `safe_one`/`safe_fetch_one` is still missing from the Data Access' Repo, so it should probably be added there

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -722,7 +722,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
       |> put_query_prefix(keyspace_name)
 
     case Repo.fetch_one(query, consistency: Consistency.device_info(:read)) do
-      device_id when is_binary(device_id) -> {:ok, true}
+      {:ok, device_id} when is_binary(device_id) -> {:ok, true}
       nil -> {:ok, false}
```
noaccOS, line 726: when you implement the suggestion from the previous comment, I would expect this piece of code to be either

```suggestion
      {:ok, _} -> {:ok, true}
      {:error, :not_found} -> {:ok, false}
```

or

```suggestion
      {:ok, device_id} -> {:ok, device_id != nil}
```

based on whether you decide to go with `safe_fetch_one` or `safe_one`

apps/astarte_data_updater_plant/test/test_helper.exs
```diff
@@ -15,9 +15,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
-Code.require_file("support/database_test_helper.exs", __DIR__)
-Code.require_file("support/amqp_test_helper.exs", __DIR__)
-Code.require_file("support/amqp_events_consumer.exs", __DIR__)
+# Code.require_file("support/amqp_test_helper.exs", __DIR__)
+# Code.require_file("support/amqp_events_consumer.exs", __DIR__)
+# Code.require_file("support/database_test_helper.exs", __DIR__)
```
noaccOS, line 20: leftover

noaccOS (changes requested):
This is a good start, but I'd expect there to be many more instances of refactoring needed
For instance, `fetch_all` from DUP returns `list() | {:error, reason}`, while the one in Data Access returns `{:ok, list()} | {:error, reason}`

OmarBrbutovic (approved):
@noaccOS provided a thorough review, and I agree with his assessment. I have no further comments.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -776,8 +775,9 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
 
     consistency = Consistency.domain_model(:read)
 
-    with n when is_number(n) or is_nil(n) <- Repo.fetch_one(query, consistency: consistency) do
-      {:ok, n}
+    with {:ok, n} when is_number(n) or is_nil(n) <-
+           Repo.fetch_one(query, consistency: consistency) do
+      n
     end
```
noaccOS, line 781:
```suggestion
    Repo.fetch_one(query, consistency: consistency)
```

let's not delete the `:ok | :error` tuple returned from this fetch

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -79,7 +79,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
       Queries.retrieve_device_stats_and_introspection!(new_state.realm, device_id)
 
     # TODO this could be a bang!
-    {:ok, ttl} = Queries.fetch_datastream_maximum_storage_retention(new_state.realm)
+
+    ttl = Queries.fetch_datastream_maximum_storage_retention(new_state.realm)
```
noaccOS, line 83: I'd expect this not to be modified

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/time_based_actions.ex
```diff
@@ -196,20 +196,20 @@ defmodule Astarte.DataUpdaterPlant.TimeBasedActions do
...
-
         {:error, _reason} ->
           Logger.warning(
             "Failed to load last_datastream_maximum_retention_refresh, keeping old one",
             tag: "last_datastream_maximum_retention_refresh_fail"
           )
 
           state
+
+        ttl ->
```
noaccOS, line 207: I'd expect this not to be changed

noaccOS (changes requested):
With the comments addressed this could be considered ready to be merged.
Still, PR actually introduces a regression, in that
- `Astarte.DataUpdaterPlant.Repo.fetch_one` is in fact not a fetch but a `safe_one`: returns an error in case of xandra error but not in case of missing values (it would return `{:ok, nil}`)
- `Astarte.DataAccess.Repo.fetch_one` does not handle xandra errors

One solution would be adding a `safe_one` or `safe_fetch_one` to DataAccess in order to handle xandra errors, and change all the `fetch_one` in this pr with that function. This could be addressed in a follow up PR, but there would need to be `TODO`s in this pr or an open issue to keep track of this

PR #1292, opened 2025-05-26
DUP: fix expiry check and add tests

OmarBrbutovic:

- Fix the expiry logic in `fetch/2` and `has_key?/2` by introducing a clearly named `expired?/1` predicate. This reverses the previous logic, making the intent explicit and avoiding double negatives

- Add complete test suite for the `Cache.ex` module, which previously lacked any test coverage.

* [ ] Yes
* [x] No

### DIFF
--- a/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/cache_test.exs
+++ b/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/cache_test.exs
@@ -1,0 +1,195 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.DataUpdaterPlant.DataUpdater.CacheTest do
+  use ExUnit.Case, async: true
+  use ExUnitProperties
+  import Mimic
+
+  alias Astarte.DataUpdaterPlant.DataUpdater.Cache
+
+  setup :verify_on_exit!
+
+  describe "put/4 and get/3" do
+    property "stores and retrieves the value for any key and value" do
+      check all(
+              key <- term(),
+              value <- term(),
+              size <- positive_integer(),
+              ttl <- one_of([nil, integer()])
+            ) do
+        cache = Cache.new(size)
+        cache = Cache.put(cache, key, value, ttl)
+
+        if is_integer(ttl) and ttl <= 0 do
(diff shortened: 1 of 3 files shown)


apps/astarte_data_updater_plant/test/astarte_data_updater_plant/cache_test.exs
```diff
@@ -0,0 +1,188 @@
...
+      cache = Cache.new(1)
+      assert Cache.get(cache, :missing, :default) == :default
+    end
+
+    test "put with negative TTL does stores the value but it is not returned" do
+      cache = Cache.new(1)
+      cache = Cache.put(cache, :foo, :bar, -1)
+      assert Cache.get(cache, :foo) == nil
+      {_size, map} = cache
+      assert Map.has_key?(map, :foo)
```
noaccOS, line 92: this is an implementation detail and I would not include this in the test

```suggestion
```

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/cache_test.exs
```diff
@@ -0,0 +1,188 @@
...
+      assert Cache.fetch(cache, :missing) == :error
+    end
+
+    test "expired key is not returned but remains in map" do
+      cache = Cache.new(1)
+      cache = Cache.put(cache, :foo, :bar, 1)
+      :timer.sleep(1100)
+      assert Cache.fetch(cache, :foo) == :error
+      {_size, map} = cache
+      assert Map.has_key?(map, :foo)
```
noaccOS, line 135: same as above

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/cache_test.exs
```diff
@@ -0,0 +1,188 @@
...
+      cache = Cache.put(cache, :c, 3, nil)
+      assert {2, map} = cache
+      assert map_size(map) == 2
+    end
+
+    test "stores value with TTL and expires after TTL" do
+      cache = Cache.new(1)
+      cache = Cache.put(cache, :foo, :bar, 1)
+      assert Cache.get(cache, :foo) == :bar
+      :timer.sleep(1100)
```
noaccOS, line 73: I'd rather avoid introducing additional `sleep`s in the tests
One way around this is by `Mimic`ing the `System.system_time` call in order to return an expired timestamp, but you're free to look for alternatives. There are a few blog posts and forum threads about testing time-dependent code
This also applies to the other `sleep` calls in this module

lusergit (commented):
@noaccOS already made a thoughtful review for this PR and I agree with him :+1:

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/cache_test.exs
```diff
@@ -0,0 +1,202 @@
...
+  import Mimic
+
+  alias Astarte.DataUpdaterPlant.DataUpdater.Cache
+
+  setup :verify_on_exit!
+
+  describe "new/1" do
+    test "returns a new cache tuple with specified size" do
+      cache = Cache.new(3)
+      assert cache == {3, %{}}
```
lusergit, line 31: nit: I think that testing this might be testing an implementation detail. If tomorrow the cache implementation changes this test has to be rewritten. I think the scope of `new/1` is no narrow that we don't actually need this test

davidebriani (approved):

PR #1307, opened 2025-06-03
[SDTT-374] test(dup): queries tests part 1

noaccOS:
init tests for untested functions in the Queries module. Part 1/2

depends on #1306

* [ ] I have read [CONTRIBUTING.md](../CONTRIBUTING.md) and [CODE_OF_CONDUCT.md](../CODE_OF_CONDUCT.md)
* [ ] I have added to [CHANGELOG.md](../CHANGELOG.md) relevant changes and any other user facing change
* [ ] I have added or ran the appropriate tests

##### Does this PR introduce a user-facing change?
* [ ] Yes
* [X] No

#### Additional documentation e.g. usage docs, diagrams, etc.:

```docs

```

PR #1308, opened 2025-06-03
[SDTT-374] test(dup): queries tests part 2

noaccOS:
init tests for untested functions in the Queries module. Part 2/2

depends on #1307

* [ ] Yes
* [X] No

```docs

```

### DIFF
--- a/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/queries_test.exs
+++ b/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/queries_test.exs
@@ -1,0 +1,47 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.DataUpdaterPlant.DataUpdater.QueriesTest do
+  use Astarte.Cases.Data, async: true
+  use Astarte.Cases.Device
+  use ExUnitProperties
+  alias Astarte.Core.Generators.Realm, as: RealmGenerator
+  alias Astarte.DataUpdaterPlant.DataUpdater.Queries
+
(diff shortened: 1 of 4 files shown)


apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/queries_test.exs
```diff
@@ -245,4 +309,55 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.QueriesTest do
...
+
+  def registered_on_interface?(realm_name, device, interface) do
+    group = "devices-with-data-on-interface-#{interface.name}-v#{interface.major_version}"
+    key = device.encoded_id
+
+    case KvStore.fetch_value(group, key, :binary, prefix: Realm.keyspace_name(realm_name)) do
+      {:ok, _} -> true
+      {:error, _} -> false
+    end
+  end
```
OmarBrbutovic: Both `registered_by_interface?/3` and `registered_on_interface?/3` are almost identical except for the group name. A suggestion, would be to refactor them into a single helper that takes the group name as a parameter:

```elixir
def registered_in_kv_store?(realm_name, device, group_prefix, interface) do
  group = "#{group_prefix}-#{interface.name}-v#{interface.major_version}"
  key = device.encoded_id

  case KvStore.fetch_value(group, key, :binary, prefix: Realm.keyspace_name(realm_name)) do
    {:ok, _} -> true
    {:error, _} -> false
  end
end
```

OmarBrbutovic (approved):

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/queries_test.exs
noaccOS: I disagree: the similarity you are trying to abstract is only there due to an implementation detail and not a conceptual correlation

apps/astarte_data_updater_plant/test/support/helpers/database.ex
```diff
@@ -356,14 +358,13 @@ defmodule Astarte.Helpers.Database do
...
     astarte_keyspace = Realm.astarte_keyspace_name()
     execute!(astarte_keyspace, @drop_keyspace)
     :ok
   end
 
-  def teardown_realm_keyspace!(realm_name) do
+  def teardown!(realm_name) do
     realm_keyspace = Realm.keyspace_name(realm_name)
     execute!(realm_keyspace, @drop_keyspace)
     :ok
```
eddbbt: I'm not completely sure this is the right way, but if it does not break anything it's fine

noaccOS: yeah I kept it as it was to keep api compatibility with what was already being used but it probably makes more sense to keep the old name

davidebriani (approved):

PR #1313, opened 2025-06-05
refactor(DUP): drastically reduce `impl.ex`

lusergit:
Attempting to refactoring DUP we noticed that a huge amount of the file is part of the data handling of a device (the `handle_data/5` function)

This PR moves the logic to a separate module and makes some cosmetic changes and minor refactoring intervention in other parts of data updater trying to reduce `impl.ex` size


lusergit:
small comment, this PR does not include tests because it is already very large. Another PR to test data handler will be made in the near future

eddbbt (approved):

PR #1314, opened 2025-06-05
DUP move gather_interface_property_paths and reduce_interface_mapping…

eddbbt:
move `gather_interface_property_paths` and `reduce_interface_mapping` into `core.interface` from `impl.ex`

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
@@ -1770,16 +1770,6 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
     end
   end
 
-  defp reduce_interface_mapping(mappings, interface_descriptor, initial_acc, fun) do
-    Enum.reduce(mappings, initial_acc, fn {_endpoint_id, mapping}, acc ->
-      if mapping.interface_id == interface_descriptor.interface_id do
-        fun.(mapping, acc)
-      else
-        acc
-      end
-    end)
-  end
-
   defp send_control_consumer_properties(state) do
     Logger.debug("Device introspection: #{inspect(state.introspection)}.")
 
@@ -1789,7 +1779,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
 
         case Core.Interface.maybe_handle_cache_miss(descriptor, interface, state) do
           {:ok, interface_descriptor, new_state} ->
-            gather_interface_property_paths(new_state.realm, interface_descriptor)
+            Core.Interface.gather_interface_property_paths(new_state.realm, interface_descriptor)
 
           {:error, :interface_loading_failed} ->
             Logger.warning("Failed #{interface} interface loading.")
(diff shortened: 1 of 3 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -1789,7 +1780,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
 
         case Core.Interface.maybe_handle_cache_miss(descriptor, interface, state) do
           {:ok, interface_descriptor, new_state} ->
-            gather_interface_property_paths(new_state.realm, interface_descriptor)
+            CoreInterface.gather_interface_property_paths(new_state.realm, interface_descriptor)
```
noaccOS: I'd say let's be consistent with other calls where the existing alias to `Core` was used, eg in `execute_incoming_data_triggers`

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -31,6 +31,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
   alias Astarte.Core.Triggers.SimpleTriggersProtobuf.Utils, as: SimpleTriggersProtobufUtils
   alias Astarte.DataAccess.Data
   alias Astarte.DataAccess.Interface, as: InterfaceQueries
+  alias Astarte.DataUpdaterPlant.DataUpdater.Core.Interface, as: CoreInterface
```
noaccOS: let's use the existing alias to `Core`

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/interface_test.exs
```diff
@@ -236,6 +236,72 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.InterfaceTest do
...
+      } = context
+
+      valid_interfaces =
+        interfaces |> Enum.filter(&(&1.type == :properties and &1.ownership == :server))
+
+      path_to_check_with = []
+
+      check all interface <- member_of(valid_interfaces),
+                mapping_update <- valid_mapping_update_for(interface) do
+        Helpers.Database.insert_values(realm_name, device, interface, [mapping_update])
```
noaccOS: This works. However, I would suggest using the `:populate_interfaces` setup so that you don't need to insert values inside a `check all`

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/interface_test.exs
```diff
@@ -236,6 +236,72 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.InterfaceTest do
...
+
+        path_to_check_with = path_to_check_with ++ mapping_update.path
```
noaccOS: I'm not sure what you're trying to do with this?

```elixir
iex(1)> [] ++ "/some/path"
"/some/path"
```

noaccOS (changes requested):
A few comments, but the organization is nice!

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/interface_test.exs
```diff
@@ -236,6 +238,65 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.InterfaceTest do
...
+
+    property "gather_interface_property_paths/2 returns always empty list, for device owned interfaces",
+             context do
+      %{
+        state: state,
+        interfaces_with_data: interfaces
+      } = context
+
+      valid_interfaces =
+        interfaces |> Enum.filter(&(&1.ownership == :device))
```
noaccOS: nit: I think we can also include server owned datastreams here

```suggestion
        interfaces |> Enum.reject(&(&1.type == :properties and &1.ownership == :server))
```

lusergit (approved):
Apart from @noaccOS comment lgtm :+1:

PR #1315, opened 2025-06-05
refactor(DUP): move `ask_clean_session` to dedicated device module

OmarBrbutovic:
* [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md) and [CODE_OF_CONDUCT.md](../CODE_OF_CONDUCT.md)
* [x] I have added to [CHANGELOG.md](../CHANGELOG.md) relevant changes and any other user facing change
* [x] I have added or ran the appropriate tests

#### What this PR does / why we need it:

Move `ask_clean_session/2` away from `impl.ex` and test separately.

##### Does this PR introduce a user-facing change?
* [ ] Yes
* [x] No

#### Additional documentation e.g. usage docs, diagrams, etc.:

```docs

```

PR #1317, opened 2025-06-05
refactor(DUP): better (and more testable) error handling

lusergit:
Based on #1315 ~and #1313~
Up to now errors were handled in `handle_data/5` directly, with little to no testability

This refactors the code by
- providing error handling function testable separately
- move error handling where errors happen instead of the `else` branch of `handle_data/5`
- provide a general framework for message context and errors in data handling

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
@@ -20,7 +20,6 @@
 
 defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
   alias Astarte.Core.Mapping.ValueType
-  alias Astarte.DataUpdaterPlant.DataUpdater.Impl
   alias Astarte.DataUpdaterPlant.DataUpdater.State
   alias Astarte.DataUpdaterPlant.DataUpdater.CachedPath
   alias Astarte.DataUpdaterPlant.DataUpdater.Cache
@@ -65,7 +64,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
           do: value_timestamp,
           else: div(timestamp, 10000)
 
-      Impl.execute_incoming_data_triggers(
+      Core.DataTrigger.execute_incoming_data_triggers(
         state,
         device_id_string,
         interface_descriptor.name,
@@ -78,7 +77,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
       )
 
       {has_change_triggers, change_triggers} =
-        Impl.get_value_change_triggers(state, interface_id, endpoint_id, path, value)
+        Core.Interface.get_value_change_triggers(state, interface_id, endpoint_id, path, value)
 
       previous_value =
         with {:has_change_triggers, :ok} <- {:has_change_triggers, has_change_triggers},
@@ -101,7 +100,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
 
       if has_change_triggers == :ok do
         :ok =
-          Impl.execute_pre_change_triggers(
+          Core.Trigger.execute_pre_change_triggers(
             change_triggers,
             state.realm,
             device_id_string,
@@ -182,7 +181,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
             "base64_payload" => base64_payload
           }
 
-          Impl.execute_device_error_triggers(
+          Core.Trigger.execute_device_error_triggers(
             state,
             "unset_on_datastream",
             error_metadata,
@@ -231,7 +230,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
             "base64_payload" => base64_payload
           }
 
-          Impl.execute_device_error_triggers(
+          Core.Trigger.execute_device_error_triggers(
             state,
             "unset_not_allowed",
             error_metadata,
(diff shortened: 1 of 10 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/error.ex
```diff
@@ -0,0 +1,108 @@
...
+
+    %{
+      message: message,
+      tag: tag,
+      error_name: error_name
+    } = error
+
+    update_stats = Map.get(error, :update_stats, true)
+
+    Logger.warning(message, tag: tag)
```
noaccOS: not sold on this abstraction
logging functions allow specifying custom [metadata](https://hexdocs.pm/logger/Logger.html#module-metadata), which we don't currently do much but we probably should, so that we can give more context to our errors.
I suggest using something more generic, like `logger_metadata` instead of allowing only `tag` to be set

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/error.ex
```diff
@@ -0,0 +1,108 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.Error do
```
noaccOS, line 21: nit: this module should be initialized in the following commit, where it is also used

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -253,365 +250,198 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
...
+      ok ->
+        ok
+    end
+  end
 
-      {:error, :mapping_not_found} ->
-        Logger.warning("Mapping not found for #{interface}#{path}. Maybe outdated introspection?",
-          tag: "mapping_not_found"
-        )
+  defp decode_bson_payload(context) do
```
noaccOS, line 326: this function returns `{value, timestamp, metadata} | {:error, reason}`
I suggest using the standard result tuple `{:ok, {value, timestamp, metadata}} | {:error, reason}`

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -253,365 +250,198 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
...
-        base64_payload = Base.encode64(payload)
+  defp validate_interface(%{interface: interface} = context),
+    do: valid_interface_or_error(context, String.valid?(interface))
 
-        error_metadata = %{
-          "interface" => inspect(interface),
-          "path" => inspect(path),
-          "base64_payload" => base64_payload
-        }
+  defp valid_interface_or_error(_context, true), do: :ok
```
noaccOS: nit: usually having the result of a check as the input parameter of a function is confusing.
As an alternative, you could consider moving the branching to the main function and having more well-defined functions where needed, eg

```elixir
defp validate_interface(context) do
  case String.valid?(context.interface) do
    true -> :ok
    false -> invalid_interface_error(context)
  end
end
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -253,365 +250,198 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
...
+        Core.Error.handle_error(context, error)
 
-        :telemetry.execute(
-          [:astarte, :data_updater_plant, :data_updater, :discarded_message],
-          %{},
-          %{realm: state.realm}
-        )
+      {:ok, descriptor, state} ->
+        new_context = Map.put(context, :state, state)
+        {:ok, descriptor, state, new_context}
```
noaccOS: if the state is already in the context, this should only return the context (that way you enforce a single source of truth for the state)

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -253,365 +250,198 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
...
-        base64_payload = Base.encode64(payload)
+    Core.Error.handle_error(context, error)
+  end
 
-        error_metadata = %{
-          "interface" => inspect(interface),
-          "path" => inspect(path),
-          "base64_payload" => base64_payload
-        }
+  defp maybe_validate_value_type(context, interface_descriptor, mapping, value) do
```
noaccOS: I'm not sure this counts as a `maybe_` as this effectively validates the value

OmarBrbutovic (approved):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
lusergit: Actually the function returns `{value, timestamp, metadata} | state`, it's not a standard result tuple

noaccOS (approved):

PR #1324, opened 2025-06-10
refactor(DUP): move `prune_device_properties` to dedicated device module

OmarBrbutovic:

Move `prune_device_properties/3` away from `impl.ex` and test separately
* [ ] Yes
* [x] No

```docs

```

### DIFF
--- a/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
+++ b/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
@@ -19,19 +19,26 @@
 defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
   use Astarte.Cases.Data, async: true
   use Astarte.Cases.Device
+  use ExUnitProperties
   use Mimic
 
   import Astarte.Helpers.DataUpdater
+  import Astarte.InterfaceUpdateGenerators
+  import Ecto.Query
 
   alias Astarte.DataUpdaterPlant.DataUpdater
   alias Astarte.DataUpdaterPlant.DataUpdater.Core
+  alias Astarte.DataUpdaterPlant.DataUpdater.Queries
   alias Astarte.DataUpdaterPlant.RPC.VMQPlugin
   alias Astarte.DataAccess.Devices.Device, as: DatabaseDevice
+  alias Astarte.DataAccess.Interface, as: InterfaceQueries
   alias Astarte.DataAccess.Realms.Realm
   alias Astarte.DataAccess.Repo
   alias Astarte.Helpers.Database
 
   @timestamp_us_x_10 Database.make_timestamp("2025-05-14T14:00:32+00:00")
+
+  setup_all :populate_interfaces
 
   setup_all %{realm_name: realm_name, device: device} do
     setup_data_updater(realm_name, device.encoded_id)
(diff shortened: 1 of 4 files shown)


apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
```diff
@@ -85,6 +88,87 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
...
+        }
+
+        expected_pruned = %{
+          path: mapping_update2.path,
+          device_id: device.device_id,
+          interface_id: interface.interface_id
+        }
+
+        assert expected_kept in properties
+        assert expected_pruned not in properties
```
noaccOS: let's try to minimize the database access everywhere except for `Queries` modules and their tests, as eventually :tm: we'll want the services not to depend on the database themselves so this would become technical debt. Also, technically individual properties are not guaranteed to always be in "individual_properties" (although as of now they do)

instead, you could use other available methods from the queries module, such as

```suggestion
        mapping1 = Core.Interface.resolve_path(mapping_update1.path, interface_descriptor, interface.mappings)
        mapping2 = Core.Interface.resolve_path(mapping_update2.path, interface_descriptor, interface.mappings)
        refute Queries.retrieve_property_values(realm_name, device.device_id, interface_descriptor, mapping1) |> Enum.empty?()
        assert Queries.retrieve_property_values(realm_name, device.device_id, interface_descriptor, mapping2) |> Enum.empty?()
```

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
```diff
@@ -85,6 +88,87 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
...
+        realm_name: realm_name,
+        device: device
+      } = context
+
+      valid_interfaces =
+        interfaces |> Enum.filter(&(&1.type == :properties and &1.ownership == :device))
+
+      check all interface <- member_of(valid_interfaces),
+                [mapping_update1, mapping_update2] <-
+                  list_of(valid_mapping_update_for(interface), length: 2) do
```
noaccOS: I'm not sure _this_ is the property I would test
I would expect the property to be on the `decoded_payload`. this is kinda tricky but an (ugly) example of this would be

```elixir
    check all interfaces <- list_of(member_of(valid_interfaces), min_length: 1),
              interfaces = Enum.uniq_by(interfaces, &{&1.name, &1.major_version}),
              mapping_updates =
                Map.new(interfaces, fn interface ->
                  {interface, list_of(valid_mapping_update_for(interface))}
                end),
              mapping_updates <- fixed_map(mapping_updates),
              mapping_updates_list =
                Enum.flat_map(mapping_updates, fn {interface, updates} ->
                  updates
                  |> Enum.uniq_by(& &1.path)
                  |> Enum.map(&{interface, &1})
                end),
              split_size <- integer(0..Enum.count(mapping_updates_list)),
              mapping_updates_list <- shuffle(mapping_updates_list),
              {to_keep, to_delete} = Enum.split(mapping_updates_list, split_size),
              decoded_payload = to_delete |> Enum.map_join(";", fn {interface, mapping_update} -> interface.name <> mapping_update.path end) do

```

it's also important to make sure we have unique interfaces and mapping updates:
- there shouldn't be two interfaces with the same name and major version
- there shouldn't be two updates for the same interface with the same path, as all properties of the same interafce have unique paths

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
```diff
@@ -85,6 +88,87 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
...
+        # Insert two properties for the device-owned interface
+        Database.insert_values(realm_name, device, interface, [mapping_update1, mapping_update2])
```
noaccOS: whenever you `insert` in a test (and especially in properties where this is called thousands of times), you should also clean at the end to leave a clean state.

there is currently no helper to do that, but in the Queries module there is `delete_property_from_db/5` which fits nicely for this job, the wrapper should be very similar to `insert_values` but instead of calling `insert_value_into_db` it should call `delete_property_from_db`.

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
```diff
@@ -85,6 +88,87 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
...
+          device_id: device.device_id,
+          interface_id: interface.interface_id
+        }
+
+        assert expected_kept in properties
+        assert expected_pruned not in properties
+      end
+    end
+
+    property "does not remove properties of server owned interfaces", context do
```
noaccOS: in this test, nothing happens because `prune_device_properties` does not delete any value, so

1. this could be a `test` instead of a `property`, checking one instance is probably enough
2. you can use the setup for pre-generating interfaces with values. this wouldn't work in the other test as we would be polluting the state by deleting values, but this isn't the case for this test. To do so:
    1. Add `setup_all :populate_interfaces`
    2. Instead of `:interfaces`, you can grab from the context `:interfaces_with_data` for the whole list, or just `individual_properties_server_interface` to have one interface of the type you need with data in it
    3. From the context, grab the map `registered_paths` and access your interface like so `paths = registered_paths |> Map.fetch!({interface.name, interface.major_version})`, now `paths` is a list of all the registered paths for the interface you need

noaccOS (changes requested):

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
```diff
@@ -85,6 +91,100 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
...
+
+      for path <- paths do
+        expected_property = %{
+          path: path,
+          device_id: device.device_id,
+          interface_id: interface.interface_id
+        }
+
+        assert expected_property in properties_after
+      end
```
noaccOS: like in the other query, you can remove the database level here as well

```suggestion
    expected_paths = Enum.sort(paths)

    actual_paths =
      interface.mappings
      |> Enum.map(& &1.endpoint_id)
      |> Enum.flat_map(&Queries.all_device_owned_property_endpoint_paths!(realm_name, device.device_id, interface_descriptor, &1))
      |> Enum.sort()
      
    assert actual_paths == expected_paths
```

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
```diff
@@ -85,6 +91,100 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
...
+      valid_interfaces =
+        interfaces
+        |> Enum.filter(&(&1.type == :properties and &1.ownership == :device))
+        |> Enum.uniq_by(&{&1.name, &1.major_version})
+
+      check all interface <- member_of(valid_interfaces),
+                [mapping_update1, mapping_update2] <-
+                  list_of(valid_mapping_update_for(interface), length: 2) do
+        # Only proceed if mapping_update1 and mapping_update2 have different paths
+        if mapping_update1.path != mapping_update2.path do
```
noaccOS: this is not great as, worst case scenario, we may see the test passing successfully even though all tests were actually skipped

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
noaccOS: I'd still suggest going this route, some private functions to abstract concepts might help making this more manageable

noaccOS (changes requested):

PR #1326, opened 2025-06-11
refactor(dup): move handle_consume to its own module

noaccOS:
- helps testing the function in isolation
- rework `/emptyCache` to avoid a giant `else` block in its `with`
- add tests

one function from impl was made public to allow access from this module,
but it should be ported to its own module soon ™️ so it's not an issue

based on #1324

Fixes #

* [ ] Yes
* [ ] No

```docs

```


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    Core.Trigger.execute_device_error_triggers(
+      new_state,
+      "unexpected_control_message",
+      error_metadata,
+      timestamp
+    )
+
+    Core.DataHandler.update_stats(new_state, "", nil, path, payload)
```
lusergit, line 145: nit: this could be done trough `Core.Error`, but is fine as it is

PR #1341, opened 2025-06-17
feat(housekeeping)!: remove update_realm rpc

noaccOS:
moves validation to api level, which means errors in the api change a little. we now also crash instead of soft handling database errors.

BREAKING CHANGE: previously it was possible to specify the realm to update by setting `data.realm_name`, as it would be applied by the changeset and passed on to the backend. Now, the update can only refer to the realm from the path.

depends on #1337

* [X] Yes
* [ ] No

```docs

```


apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/realms/queries.ex
```diff
@@ -88,6 +89,79 @@ defmodule Astarte.Housekeeping.API.Realms.Queries do
...
+
+    kv_store = %{
+      group: "realm_config",
+      key: "datastream_maximum_storage_retention",
+      value: retention,
+      value_type: :integer
+    }
+
+    with {:error, xandra_error} <- KvStore.insert(kv_store, opts) do
+      raise xandra_error
```
davidebriani, line 130: Why raising here but not in other functions that update realm fields (`set_device_registration_limit`)?

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/realms/queries.ex
```diff
@@ -24,6 +24,7 @@ defmodule Astarte.Housekeeping.API.Realms.Queries do
   alias Astarte.DataAccess.Realms.Realm
   alias Astarte.DataAccess.Repo
   alias Astarte.Housekeeping.API.Realms.Realm, as: HKRealm
+  require IEx
```
davidebriani: ?

davidebriani (changes requested):

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/realms/queries.ex
eddbbt: ??

eddbbt (approved):
lgtm

apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/realms/queries.ex
noaccOS: `update_all` and `delete_all` raise on error, the odd one out is `KvStore.insert` which is closer to a `safe_insert` and returns `{:error, reason}` in case of database error

PR #1346, opened 2025-06-18
feat(RM): remove delete_device RPC

emirh1:

Perform `delete_device` directly on the api service.
Update tests to work with new database helpers.

Fixes #

* [ ] Yes
* [x] No

```docs

```

### DIFF
--- a/apps/astarte_realm_management_api/lib/astarte_realm_management_api/devices/devices.ex
+++ b/apps/astarte_realm_management_api/lib/astarte_realm_management_api/devices/devices.ex
@@ -1,7 +1,7 @@
 #
 # This file is part of Astarte.
 #
-# Copyright 2023 SECO Mind Srl
+# Copyright 2023 - 2025 SECO Mind Srl
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -17,9 +17,21 @@
 #
 
 defmodule Astarte.RealmManagement.API.Devices do
-  alias Astarte.RealmManagement.API.RPC.RealmManagement
+  alias Astarte.RealmManagement.API.Devices.Queries
+  alias Astarte.Core.Device
 
   def delete_device(realm_name, device_id) do
-    RealmManagement.delete_device(realm_name, device_id)
+    with {:ok, decoded_id} <- Device.decode_device_id(device_id),
+         :ok <- ensure_device_exists(realm_name, decoded_id) do
+      Queries.insert_device_into_deletion_in_progress(realm_name, decoded_id)
+      :ok
+    end
+  end
+
+  def ensure_device_exists(realm_name, device_id) do
+    case Queries.check_device_exists(realm_name, device_id) do
+      {:ok, true} -> :ok
+      {:ok, false} -> {:error, :device_not_found}
+    end
   end
 end
--- a/apps/astarte_realm_management_api/lib/astarte_realm_management_api/devices/queries.ex
+++ b/apps/astarte_realm_management_api/lib/astarte_realm_management_api/devices/queries.ex
@@ -1,0 +1,62 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.RealmManagement.API.Devices.Queries do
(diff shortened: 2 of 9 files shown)


apps/astarte_realm_management_api/lib/astarte_realm_management_api/devices/queries.ex
```diff
@@ -0,0 +1,61 @@
...
+    # TODO check
+    opts = [
+      prefix: keyspace_name,
+      consistency: Consistency.device_info(:write)
+    ]
+
+    _ = Repo.delete_all(query, opts)
+
+    :ok
+  end
```
lusergit: Device deletion does not work like this, for reference see [the old implementation in `engine.ex`](https://github.com/astarte-platform/astarte/blob/de4af683761fa5c4fb7d05015dbf66985bacaa1b/apps/astarte_realm_management/lib/astarte_realm_management/engine.ex#L924-L940)
Essentially it is carried out as follows
1. Creates a `DeletionInProgres` and inserts that into the DB (this is done by realm management)
2. `vqm`, and `dup` have to send their acks, they'll update the struct in the DB autonomously
3. the `DeviceRemover` every 5 minutes fetches all devices to delete (all those devices where all acks are in place) and deletes them by deleting the corresponding entries in several different tables (see [these files](https://github.com/astarte-platform/astarte/tree/release-1.3/apps/astarte_realm_management/lib/astarte_realm_management/device_removal) for reference).

I think that in this PR you can focus on the first point of this list, as the device removal is a big change that can be carried out in a separate PR

apps/astarte_realm_management_api/lib/astarte_realm_management_api/devices/queries.ex
```diff
@@ -0,0 +1,61 @@
...
+        select: device.device_id,
+        where: [device_id: ^device_id]
+
+    opts = [
+      prefix: keyspace,
+      consistency: Consistency.device_info(:read)
+    ]
+
+    Repo.some?(query, opts)
+  end
```
noaccOS: nit: not the scope of this pr, but this could be a single query

```suggestion
     case Repo.delete_all(query, opts) do
       {0, _} -> {:error, :device_not_found}
       {1, _} -> :ok
     end
  end
```

apps/astarte_realm_management_api/test/astarte_realm_management_api/devices/devices_test.exs
lusergit, line 1: I think what you want to do here is to setup the tests to create a device for you. luckily we already took care of that, you can copy from [`data_updater_plant` `device.ex` case](https://github.com/astarte-platform/astarte/blob/release-1.3/apps/astarte_data_updater_plant/test/support/cases/device.ex) (you'll also need the [`device.ex` helper](https://github.com/astarte-platform/astarte/blob/release-1.3/apps/astarte_data_updater_plant/test/support/helpers/device.ex)

lusergit (changes requested):
Overall it's ok, you just need to copy the logic and test utilities and it's good to go :+1:

apps/astarte_realm_management_api/lib/astarte_realm_management_api/devices/devices.ex
```diff
@@ -17,9 +17,16 @@
...
 
   def delete_device(realm_name, device_id) do
-    RealmManagement.delete_device(realm_name, device_id)
+    with {:ok, decoded_id} <- Device.decode_device_id(device_id),
+         {:ok, true} <- Queries.check_device_exists(realm_name, decoded_id) do
+      Queries.delete_device!(realm_name, decoded_id)
+    else
+      {:ok, false} ->
+        {:error, :device_not_found}
+    end
```
noaccOS: as it is written, this wouldn't work: when there is an `else` block in a with, if all patterns fail to match the return value then an exception is thrown.
in practice, this means that if you fail to decode the device_id, you crash
you can
~1. change the query with the suggestion from my previous comment and delete the else block~
2. have a wrapper function for check_device_exists which returns `:ok | {:error, :device_not_found}` so you don't need an else block
3. add a generic `{:error, reason} -> {:error, reason}` branch to the else block

~if you don't want to refactor the query,~ I suggest going with 2.

apps/astarte_realm_management_api/lib/astarte_realm_management_api/devices/queries.ex
noaccOS: actually, Luca is right and this query is different from the one in the backend service, so this can be ignored

apps/astarte_realm_management_api/test/astarte_realm_management_api/devices/devices_test.exs
```diff
@@ -19,23 +19,36 @@
...
+      realm: realm,
+      device_id: device_id
+    } do
+      keyspace = Realm.keyspace_name(realm)
+      {:ok, decoded_id} = Device.decode_device_id(device_id)
+
+      assert :ok = Devices.delete_device(realm, device_id)
+
+      [deletion] = Repo.all(DeletionInProgress, prefix: keyspace)
+      _ = Repo.delete!(deletion)
```
noaccOS, line 43: nit: as a rule of thumb, try to avoid calling database functions directly in non-query module tests, as _eventually_ the plan is to move all database functionality (query modules and _their_ tests included) in data access

noaccOS (approved):

PR #1357, opened 2025-06-23
refactor(RM): move `fetch_interface` to the RM frontend

lusergit:
Moves `fetch_interface` from RealmManagement to  the `API` frontend service. This is part of a series of PRs trying to refactor our services and rationalise our infrastructure.

- `RealmManagement.API` service now has the `RealmManagement` test and dev dependency
- The `fetch_interface` query gets moved to the frontend service
- Old tests are skipped, I found this approach simpler instead of trying to make everything work for tests that should disappear in a couple of weeks
- New tests have been put in place, adding support for astarte test cases (database and conn)

### DIFF
--- a/apps/astarte_realm_management_api/test/astarte_realm_management_api_web/controllers/interface_controller_test.exs
+++ b/apps/astarte_realm_management_api/test/astarte_realm_management_api_web/controllers/interface_controller_test.exs
@@ -17,13 +17,15 @@
 #
 
 defmodule Astarte.RealmManagement.APIWeb.InterfaceControllerTest do
-  use Astarte.RealmManagement.APIWeb.ConnCase, async: true
+  use Astarte.RealmManagement.API.DataCase, async: true
+  use Astarte.RealmManagement.APIWeb.ConnCase
 
   @moduletag :interfaces
 
   alias Astarte.Core.Interface
   alias Astarte.Helpers.Database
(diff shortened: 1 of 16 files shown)


apps/astarte_realm_management_api/test/astarte_realm_management_api_web/controllers/interface_controller_test.exs
```diff
@@ -20,6 +20,9 @@ defmodule Astarte.RealmManagement.APIWeb.InterfaceControllerTest do
   use Astarte.RealmManagement.APIWeb.ConnCase, async: true
 
   @moduletag :interfaces
+  # TODO: Remove this module and make `new_interface_controller_test.exs` the
+  # only one once all interface functions have been migrated to the API service.
+  @moduletag :skip
```
noaccOS: this would be better used on each single test so that they can be removed individually once all needed rpcs have been removed

apps/astarte_realm_management_api/test/support/cases/conn.ex
```diff
@@ -0,0 +1,55 @@
+#
```
noaccOS: I'd say let's not rewrite the same case twice, we can instead adapt the old case to our new use case

noaccOS (changes requested):

emirh1 (approved):

PR #1358, opened 2025-06-23
feat(RM): remove `get_interfaces_list` RPC

emirh1:

- `get_interfaces_list` query is moved to the frontend service
- New tests are added to new test modules
- install_interface is temporarily used from `Engine` module, until that RPC also gets removed

Fixes #

* [ ] Yes
* [x] No


apps/astarte_realm_management_api/test/astarte_realm_management_api/interfaces/new_interfaces_test.exs
```diff
@@ -44,6 +44,29 @@ defmodule Astarte.RealmManagement.API.Interfaces.NewInterfacesTest do
     end
   end
 
+  describe "list_interfaces/1" do
+    setup %{realm_name: realm_name} do
+      interfaces = Generators.Interface.interface() |> Enum.take(3)
```
lusergit: Nit, use `Enum.take_random/1` instead of taking a fixed number of elelments

apps/astarte_realm_management_api/test/astarte_realm_management_api_web/controllers/new_interface_controller_test.exs
```diff
@@ -46,4 +46,26 @@ defmodule AstarteRealmManagementApiWeb.NewInterfaceControllerTest do
       assert json_response(show_conn, 200)["data"]["interface_name"] == interface.name
     end
   end
+
+  describe "index endpoint" do
+    setup %{realm_name: realm_name} do
+      interfaces = Generators.Interface.interface() |> Enum.take(3)
```
lusergit: Same as before

apps/astarte_realm_management_api/test/astarte_realm_management_api/interfaces/new_interfaces_test.exs
lusergit: My bad, `take_random` traverses the whole list to get a random list, so this is not feasible with `StreamData`, instead you can do
```elixir
Generators.Interface.interface() |> list_of() |> Enum.at(0)
```
this way you have a random list of interfaces

apps/astarte_realm_management_api/mix.exs
```diff
@@ -87,8 +87,6 @@ defmodule Astarte.RealmManagement.API.Mixfile do
 
   defp deps do
     [
-      {:astarte_realm_management,
-       path: "../astarte_realm_management", only: [:dev, :test], runtime: false},
       {:phoenix, "~> 1.7"},
```
eddbbt, line 90: inside `apps/astarte_realm_management_api/test/astarte_realm_management_api_web/controllers/new_interface_controller_test.exs:27` it's still used, are we 100% sure that removing it does not cause any problem?

eddbbt: inside `apps/astarte_realm_management_api/test/astarte_realm_management_api_web/controllers/new_interface_controller_test.exs:27` and `apps/astarte_realm_management_api/test/astarte_realm_management_api/interfaces/new_interfaces_test.exs:27` it's still used, are we 100% sure that removing it does not cause any problem?

Also, inside `astarte_required_modules(_)` it's still present

eddbbt (commented):
Locally tests are giving a, somewhat, strange warning about exandra connection, before approving I'd wait to see if the same happens in CI (after a rebase, clearly)

apps/astarte_realm_management_api/mix.exs
emirh1: It was removed from `deps` because it was already present in `astarte_required_modules` and that triggered a warning

noaccOS: it should be removed from `astarte_required_modules` as it's only a test dependency and not a core astarte component

eddbbt (approved):

noaccOS (approved):

PR #1360, opened 2025-06-23
refactor(RM): move `install_interface` to the frontend

lusergit:
Based on #1357 

Moves `install_interface` to RealmManagementAPI, adding new tests for the ported code

### DIFF
--- a/apps/astarte_realm_management_api/lib/astarte_realm_management_api/rpc/realm_management.ex
+++ b/apps/astarte_realm_management_api/lib/astarte_realm_management_api/rpc/realm_management.ex
@@ -27,7 +27,6 @@ defmodule Astarte.RealmManagement.API.RPC.RealmManagement do
     GetInterfaceVersionsList,
     GetInterfaceVersionsListReply,
     GetInterfaceVersionsListReplyVersionTuple,
-    InstallInterface,
     Reply,
     InstallTriggerPolicy,
     GetTriggerPoliciesList,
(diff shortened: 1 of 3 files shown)


apps/astarte_realm_management_api/test/astarte_realm_management_api/interfaces/new_interfaces_test.exs
```diff
@@ -44,6 +58,33 @@ defmodule Astarte.RealmManagement.API.Interfaces.NewInterfacesTest do
...
+      type: mapping.value_type,
+      allow_unset: mapping.allow_unset,
+      retention: mapping.retention,
+      expiry: mapping.expiry,
+      database_retention_policy: mapping.database_retention_policy,
+      database_retention_ttl: mapping.database_retention_ttl,
+      explicit_timestamp: mapping.explicit_timestamp
+    }
+  end
+
```
OmarBrbutovic: You can move this directly to the database helper now, since it's being moved there in the next commit anyway

OmarBrbutovic (changes requested):
The the dependency of PRs could be changed, so that #1357 depends on #1360 and not the other way around, reducing the need for later changes

emirh1 (approved):

apps/astarte_realm_management_api/lib/astarte_realm_management_api/interfaces/interfaces.ex
```diff
@@ -41,59 +40,100 @@ defmodule Astarte.RealmManagement.API.Interfaces do
...
+      normalized_names = Enum.map(names, &normalize_interface_name/1)
+
+      collision? = Enum.any?(normalized_names, fn name -> name == normalized_interface_name end)
+
+      case collision? do
+        true -> {:error, :interface_name_collision}
+        false -> :ok
+      end
+    end
+  end
```
OmarBrbutovic: The updated version of this function does not allow reinstallation of a interface with the same name, while it is possible to reinstall a interface on Astarte if the major version is different.

OmarBrbutovic (changes requested):

OmarBrbutovic (approved):

noaccOS (approved):

PR #1363, opened 2025-06-24
feat(rm): remove get_trigger_policy_source and related functions

eddbbt:
remove get_trigger_policy_source, related functions and tests

BREAKING CHANGE: get_trigger_policy cannot be called from rpc anymore

### DIFF
--- a/apps/astarte_realm_management_api/test/astarte_realm_management_api/triggers/policies/policy_test.exs
+++ b/apps/astarte_realm_management_api/test/astarte_realm_management_api/triggers/policies/policy_test.exs
@@ -57,6 +57,8 @@ defmodule Astarte.RealmManagement.API.Triggers.Policies.PolicyTest do
         name = policy_map.name
 
         assert {:ok, %Policy{name: ^name}} = Policies.create_trigger_policy(realm, policy_map)
+
+        RealmManagement.Engine.delete_trigger_policy(realm, name)
       end
     end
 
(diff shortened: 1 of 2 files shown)


lusergit (approved):
:fire:

apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/policies/policies.ex
```diff
@@ -42,7 +45,12 @@ defmodule Astarte.RealmManagement.API.Triggers.Policies do
 
   """
   def get_trigger_policy_source(realm_name, policy_name) do
-    RealmManagement.get_trigger_policy_source(realm_name, policy_name)
+    with {:ok, policy_proto} <- Queries.fetch_trigger_policy(realm_name, policy_name) do
+      policy_proto
+      |> PolicyProto.decode()
+      |> Policy.from_policy_proto!()
+      |> Jason.encode()
+    end
```
davidebriani: `policy_proto` is fetched directly from the database, are we sure we need to decode it using protobuf, create a Policy struct, and encode it in JSON?

apps/astarte_realm_management_api/test/support/helpers/astarte_realm_management_mock.ex
```diff
@@ -42,6 +42,7 @@ defmodule Astarte.RealmManagement.API.Helpers.RPCMock do
     DeleteTriggerPolicy,
     GetTriggerPolicySource,
     GetTriggerPolicySourceReply,
+    DeleteDevice,
```
davidebriani: ?

apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/policies/policies.ex
davidebriani: Nevermind, it seems that the previous code was already doing this decoding after fetching from database.
Indeed, policies are inserted as protobuf-encoded binaries into the database kv-store table.

davidebriani (approved):

PR #1367, opened 2025-06-24
chore(housekeeping): squash housekeeping api and backend service

noaccOS:
depends on #1366

Fixes #

* [ ] Yes
* [ ] No

```docs

```


apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -16,33 +16,33 @@
...
-  alias Astarte.Housekeeping.API.Config.JWTPublicKeyPEMType
+defmodule Astarte.Housekeeping.Config do
+  alias Astarte.Housekeeping.Config.JWTPublicKeyPEMType
   alias Astarte.DataAccess.Config, as: DataAccessConfig
 
   use Skogsra
 
   @envdoc "The bind address for the Phoenix server."
-  app_env :bind_address, :astarte_housekeeping_api, :bind_address,
+  app_env :bind_address, :astarte_housekeeping, :bind_address,
```
OmarBrbutovic, line 26: Since we’re rebranding from Housekeeping API to just Housekeeping, would it make sense to update the related OS environment variables as well (e.g., remove the `_API` suffix from `HOUSEKEEPING_API_BIND_ADDRESS` and others) for consistency?
I realize this would be a breaking change, so it might need to be planned and communicated. Just wanted to flag it for discussion. Otherwise, keeping the current names makes sense.

OmarBrbutovic (approved):
One minor nitpick, now that the app name has changed, we should probably update the copyright notices in the affected files.
I realize this is tedious and not urgent, so just flagging it as something to maybe address.

PR #1368, opened 2025-06-24
Merge Pairing and Pairing API

davidebriani:

The PR merges the Pairing service into the Pairing API service, keeping only one service.
With this bold move we are able to avoid all existing RPC calls over AMQP between the two services: now resolving API requests just involves usual function calls, and perhaps database calls.

The PR first focuses on just copying over relevant files and code from Pairing into Pairing API, without modification, and adding test helpers that will be used for refactoring unit tests.

Then, the implementation of each API route is changed one by one to directly call the imported business logic instead of making an RPC request over AMQP.
Tests are updated accordingly to avoid mocking RPC requests and instead just expecting the desired behavior.
The major changes in tests are to setup the keyspaces with a realm, and a device, so that tests actually run against actual resources: the business logic is now tested end-to-end since there are no in-the-middle mocked RPC calls.


apps/astarte_pairing_api/test/astarte_pairing_api/agent/agent_test.exs
```diff
@@ -17,235 +17,146 @@
...
-
-        {:ok, @encoded_realm_not_found_response}
-      end)
-
-      assert {:error, :forbidden} = Agent.unregister_device(@test_realm, @test_device_id)
+    test "realm not found", %{device: device} do
+      # TODO: This should not raise an error, but return an error tuple
+      assert_raise Xandra.Error, ~r"Keyspace .*nonexistingrealm does not exist", fn ->
+        Agent.unregister_device("nonexistingrealm", device.encoded_id)
+      end
```
noaccOS, line 154: this is giving the api user a different meaning regarding the error, as it changes from a user error to a server error

apps/astarte_pairing_api/test/astarte_pairing_api/credentials/credentials_test.exs
```diff
@@ -99,105 +130,83 @@ defmodule Astarte.Pairing.API.CredentialsTest do
...
-               Credentials.get_astarte_mqtt_v1(
-                 @unexisting_realm,
-                 @hw_id,
-                 @secret,
-                 @device_ip,
-                 @valid_attrs
-               )
+    test "returns forbidden with realm not found", ctx do
+      # TODO: This should not raise an error, but return an error tuple
+      assert_raise Xandra.Error, ~r"Keyspace .*#{@unexisting_realm} does not exist", fn ->
```
noaccOS, line 192: same as above

apps/astarte_pairing_api/test/astarte_pairing_api/credentials/credentials_test.exs
```diff
@@ -213,171 +222,115 @@ defmodule Astarte.Pairing.API.CredentialsTest do
...
     end
 
-    test "returns device_not_found with device not found" do
-      MockRPCClient
-      |> expect(:rpc_call, fn _serialized_call, @rpc_destination, @timeout ->
-        {:ok, @encoded_device_not_found_response}
-      end)
+    test "raises with realm not found", ctx do
+      # TODO: This should not raise an error, but return an error tuple
+      assert_raise Xandra.Error, ~r"Keyspace .*#{@unexisting_realm} does not exist", fn ->
```
noaccOS, line 320: ^

apps/astarte_pairing_api/test/astarte_pairing_api_web/controllers/device_controller_test.exs
```diff
@@ -191,51 +224,68 @@ defmodule Astarte.Pairing.APIWeb.DeviceControllerTest do
...
+          ),
+          data: @invalid_crt_attrs
         )
 
       assert %{
                "valid" => false,
-               "timestamp" => @now_string,
-               "cause" => "EXPIRED",
+               "timestamp" => _timestamp,
+               "cause" => "INVALID",
```
noaccOS, line 291: Does not need to happen this pr, but it would be great to also test for expired certs

apps/astarte_pairing_api/test/support/cases/device.ex
```diff
@@ -158,4 +159,22 @@ defmodule Astarte.Cases.Device do
...
+    |> Repo.update!(prefix: keyspace)
+  end
+
+  def update_device!(realm_name, device_id, params \\ []) do
+    keyspace = Realm.keyspace_name(realm_name)
+
+    Repo.get!(Device, device_id, prefix: keyspace)
+    |> Ecto.Changeset.change(params)
+    |> Repo.update!(prefix: keyspace)
+  end
```
noaccOS: nit: should have been in helper

apps/astarte_pairing_api/test/astarte_pairing_api/agent/agent_test.exs
davidebriani: I added a couple of commits to improve this part:
- Ensure we use only use Repo from DataAccess for database queries
- Add a Plug for API routes that accepts a `realm` parameter to ensure that the realm exists before running any business logic. This ensures that the API user receives a well-formatted error, while we don't have to handle keyspace errors in the low-level database queries.

apps/astarte_pairing_api/lib/astarte_pairing_api_web/plug/verify_realm_exists.ex
```diff
@@ -0,0 +1,67 @@
...
+        conn
+
+      {:ok, false} ->
+        Logger.warning("Realm #{realm_name} does not exist.",
+          tag: "realm_not_found",
+          realm: realm_name
+        )
+
+        conn
+        |> put_status(:not_found)
```
noaccOS: Before, we were returning 403 in case of realm not found.
Now we're returning 404 if the realm is not found and 401/403 if the user does not have valid permissions, potentially leaking the information that a realm with a certain name exists.
For now, I think it would be best maintaining the old behavior of returning 403, at a later date we can discuss using other return codes, but they need to be consistent throughout the service

apps/astarte_pairing_api/test/astarte_pairing_api_web/controllers/device_controller_test.exs
davidebriani: Sounds fair, let's do that on another PR

apps/astarte_pairing_api/test/support/cases/device.ex
davidebriani: Yes, moved

apps/astarte_pairing_api/lib/astarte_pairing_api_web/plug/verify_realm_exists.ex
davidebriani: I agree, thanks for pointing that out.
Given that we also log when the realm is not found, we can both avoid disclosure of sensitive details on the outside and still have informative clues internally to see requests against a non-existing realm.
I updated the code with a 403 error.

eddbbt (approved):

PR #1371, opened 2025-06-25
feat(RM): remove install_trigger_policy RPC

emirh1:

`install_trigger_policy` query is moved to the frontend service
New tests are added to new test modules
Engine functions are transferred to new `Core` module

Fixes #

* [ ] Yes
* [x] No

```docs

```

### DIFF
--- a/apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/policies/policies.ex
+++ b/apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/policies/policies.ex
@@ -23,6 +23,9 @@ defmodule Astarte.RealmManagement.API.Triggers.Policies do
   alias Astarte.RealmManagement.API.Triggers.Queries, as: TriggerQueries
   alias Astarte.Core.Triggers.Policy
   alias Astarte.Core.Triggers.PolicyProtobuf.Policy, as: PolicyProto
+  alias Astarte.RealmManagement.API.Triggers.Policies.Core
+
+  require Logger
 
   @doc """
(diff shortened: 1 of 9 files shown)


apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/policies/policies.ex
```diff
@@ -67,11 +70,24 @@ defmodule Astarte.RealmManagement.API.Triggers.Policies do
...
+          tag: "install_policy_started",
+          policy_name: policy_name
+        )
+
+      policy_proto =
+        policy
+        |> Policy.to_policy_proto()
+        |> PolicyProto.encode()
+
+      Queries.install_trigger_policy(realm_name, policy_name, policy_proto)
```
ArnelaL, line 90: In the previous implamentation installation of trigger policy was always async, while the current one does it always syncrononously. You should explore expanding the code to support both options.

eddbbt (approved):

apps/astarte_realm_management_api/lib/astarte_realm_management_api/triggers/policies/policies.ex
eddbbt: it was async due to the fact that the rpc channel is not syncronous, imho there's no need to make this db write async. Also, the previous return value `{:ok, :started}` was not a confirm of a successful install, with this rewrite it is

less flakiness, less pieces that can fail, if there is no business decision to enforce an async installation of trigger policies there's also no problem

ArnelaL (approved):

noaccOS (approved):

PR #1380, opened 2025-06-30
fix(HK): Dialyzer warnings and type issues across modules

OmarBrbutovic:

Resolves several Dialyzer warnings and type mismatches across the codebase of housekeeping:
- `BackendSupervisor`: Fixed `init/1` has no local return by ensuring correct supervisor spec structure.
- `Config`: Changed :disable_authentication type from `:binary` to `:boolean` to match expected usage.
- `Core`: Removed unreachable error clause in `update_realm/2`.
- `Realms`: Replaced unresolved `Realm.t()` type with explicit `%Realm{}`.
- `VerifyHeader`: Fixed unreachable pattern in `get_secret/0` by correcting error handling.

These are non-functional changes that clean up static analysis and improve type correctness.

* [ ] Yes
* [x] No

```docs

```


apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realms.ex
```diff
@@ -85,7 +85,7 @@ defmodule Astarte.Housekeeping.Realms do
   Updates a realm with the provided list of attributes.
   Returns either {:ok, %Realm{}} or {:error, error}
   """
-  @spec update_realm(binary(), map()) :: {:ok, Realm.t()} | {:error, any()}
+  @spec update_realm(binary(), map()) :: {:ok, %Realm{}} | {:error, any()}
```
noaccOS: I believe credo screams at you if you do this
Was dialyzer not working with `Realm.t()`?

OmarBrbutovic: The type `Realm.t()` isn't actually defined, it's an `embedded_schema`, so there's no explicit `@type` for it, which is why Dialyzer complains with:

```bash
Unknown type: Astarte.Housekeeping.Realms.Realm.t/0. ElixirLS Dialyzer
```

noaccOS: then it should be changed to a `typed_embedded_schema` from the typedstruct library instead. we don't have credo enabled (yet), but: https://hexdocs.pm/credo/Credo.Check.Warning.SpecWithStruct.html

apps/astarte_housekeeping/lib/astarte_housekeeping/backend_supervisor.ex
```diff
@@ -32,16 +32,25 @@ defmodule Astarte.Housekeeping.BackendSupervisor do
...
-    Supervisor.init(children, opts)
+      opts = [strategy: :rest_for_one, name: __MODULE__]
+      Supervisor.init(children, opts)
+    rescue
+      exception ->
+        Logger.error("Failed to initialize BackendSupervisor: #{inspect(exception)}",
+          tag: "housekeeping_backend_sup_init"
+        )
+
+        :ignore
```
noaccOS: returning `:ignore` means this supervisor won't start but the application will be shown as running correctly (which it won't), meanwhile by throwing we make the supervisor restart and try again a few times, before giving up with an error.

If you don't like crashing, I [think](https://hexdocs.pm/elixir/1.12/Supervisor.html#start_link/3) you can just return an error:
```elixir
with {:ok, xandra_options} <- Config.xandra_options() do
  # ...
end
```

OmarBrbutovic: There is no `xandra_option()`, only `xandra_options!()`

noaccOS: If you find we have a problem with dialyer as it is handled right now, then it would be a good idea adding it.
For this pr it would be enough adding it to the HK Config module and adding a TODO, like
```elixir
  def xandra_options do
    try do
      {:ok, xandra_options!()}
    rescue
      _ -> {:error, :invalid_xandra_config}
    end
  end
```

But long term it should be handled in data access, possibly without using try/rescue

apps/astarte_housekeeping/lib/astarte_housekeeping/backend_supervisor.ex
```diff
@@ -32,16 +32,16 @@ defmodule Astarte.Housekeeping.BackendSupervisor do
...
-      {Astarte.DataAccess, data_access_opts}
-    ]
+      children = [
+        {Xandra.Cluster, hk_xandra_opts},
+        {Astarte.DataAccess, data_access_opts}
+      ]
 
-    opts = [strategy: :rest_for_one, name: __MODULE__]
-    Supervisor.init(children, opts)
+      Supervisor.init(children, strategy: :rest_for_one)
```
noaccOS: why was the name removed?

OmarBrbutovic: it is not a valid [option](https://hexdocs.pm/elixir/1.15.7/Supervisor.html#init/2-options) for `Supervisor.init`, and throws a dialyzer warning in this configuration

noaccOS: You're right, it should be moved to [`start_link`](https://hexdocs.pm/elixir/1.15.7/Supervisor.html#start_link/2)

OmarBrbutovic: it already is there, so nothing changed

apps/astarte_housekeeping/lib/astarte_housekeeping_web/plug/verify_header.ex
```diff
@@ -43,9 +43,8 @@ defmodule Astarte.HousekeepingWeb.Plug.VerifyHeader do
   end
 
   defp get_secret do
-    with {:ok, public_key_pem} <- Auth.fetch_public_key(),
-         %JWK{} = jwk <- JWK.from_pem(public_key_pem) do
-      jwk
+    with {:ok, public_key_pem} <- Auth.fetch_public_key() do
+      JWK.from_pem(public_key_pem)
```
noaccOS: nit: as this is now a `with` block with one match and an `else` block, what do you think about changing it to a `case` instead?

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/core.ex
```diff
@@ -30,7 +30,6 @@ defmodule Astarte.Housekeeping.Realms.Core do
         case update_realm_field(realm, field, changed_value) do
           :ok -> {:cont, {:ok, %{realm | field => changed_value}}}
           {:ok, realm} -> {:cont, {:ok, realm}}
-          error -> {:halt, error}
```
noaccOS, line 33: it's true that these functions can't return an error, but then I'd say this should be changed to a simple reduce and not a reduce while. that also means it should now return only the realm (and not `{:ok, realm}`) and so its usage should be updated in `realms.ex`

lusergit (approved):

PR #1383, opened 2025-06-30
refactor(DUP): use exandra for all queries

ArnelaL:

- use Exandra for all queries in DUP

* [ ] Yes
* [x] No

```docs

```

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
@@ -869,84 +869,33 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
   end
 
   def check_device_deletion_in_progress(realm_name, device_id) do
-    keyspace_name = Realm.keyspace_name(realm_name)
+    keyspace = Realm.keyspace_name(realm_name)
 
-    Xandra.Cluster.run(
-      :xandra,
-      &do_check_device_deletion_in_progress(&1, keyspace_name, device_id)
-    )
-  end
+    query =
+      from d in DeletionInProgress,
+        prefix: ^keyspace,
+        where: d.device_id == ^device_id
 
-  defp do_check_device_deletion_in_progress(conn, realm_name, device_id) do
-    statement = """
-    SELECT *
-    FROM #{realm_name}.deletion_in_progress
-    WHERE device_id = :device_id
-    """
+    consistency = Consistency.device_info(:read)
 
-    opts = [
-      consistency: Consistency.device_info(:read),
-      uuid_format: :binary
-    ]
-
-    with {:ok, prepared} <- Xandra.prepare(conn, statement),
-         {:ok, %Xandra.Page{} = page} <-
-           Xandra.execute(conn, prepared, %{"device_id" => device_id}, opts) do
-      result_not_empty? = not Enum.empty?(page)
-      {:ok, result_not_empty?}
-    else
-      {:error, %Xandra.Error{} = error} ->
-        _ =
-          Logger.warning(
-            "Database error while checking device deletion in progress: #{Exception.message(error)}"
-          )
-
-        {:error, :database_error}
-
-      {:error, %Xandra.ConnectionError{} = error} ->
-        _ =
-          Logger.warning(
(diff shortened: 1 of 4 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -869,84 +869,45 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
+    consistency = Consistency.domain_model(:read)
 
-    Xandra.execute!(conn, statement, %{},
-      consistency: Consistency.device_info(:read),
-      uuid_format: :binary
-    )
-    |> Enum.to_list()
+    with {:ok, devices} <- Repo.fetch_all(query, consistency: consistency) do
+      devices
+    end
```
noaccOS: This was previously a bang

```suggestion
    Repo.all(query, consistency: consistency)
```

nit: also tbf you don't need to specify a custom query for each repo operation, especially `select *` are quite ergonomic

```suggestion
    Repo.all(DeletionInProgress, prefix: keyspace_name, consistency: consistency)
```

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/dup_device_delete_test.exs
```diff
@@ -85,24 +81,15 @@ defmodule Astarte.DataUpdaterPlant.DeviceDeleteTest do
...
-          Xandra.execute!(conn, prepared, %{"device_id" => device_id}, uuid_format: :binary)
+    dup_start_ack_query =
+      from d in DeletionInProgress,
+        prefix: ^keyspace_name,
+        where: d.device_id == ^device_id,
+        select: d.dup_start_ack
 
-        Enum.to_list(page)
-      end)
+    {:ok, dup_start_ack_result} = Repo.fetch_all(dup_start_ack_query)
```
noaccOS: nit: this is quite alright as it is but don't be afraid of being assertive in tests

```suggestion
    dup_start_ack_result = Repo.all(dup_start_ack_query)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -869,84 +869,45 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
+    query =
+      from Realm,
+        prefix: ^keyspace_name
+
+    consistency = Consistency.domain_model(:read)
 
-    Enum.to_list(realms)
+    with {:ok, realms} <- Repo.fetch_all(query, consistency: consistency) do
+      realms
+    end
```
noaccOS: this was also a bang

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -869,84 +869,45 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
-        _ =
-          Logger.warning(
-            "Database connection error while checking device deletion in progress: #{Exception.message(error)}"
-          )
+    consistency = Consistency.device_info(:read)
 
-        {:error, :database_connection_error}
+    with {:ok, result} <- Repo.fetch_all(query, consistency: consistency) do
+      {:ok, Enum.any?(result)}
     end
```
noaccOS: while in practice this works, as any returned struct is truthy and makes this check pass, the semantic is not quite the same as a result like `[false]` would return `false` with `Enum.any?` but not with the previous logic

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
lusergit: I agree, what you probably want to do here is
```elixir
Repo.some?(query, consistency: consistency)
```

noaccOS: `Repo.some?` does not return database errors and instead masks them with an `{:ok, false}`. Until that is fixed, it should not be used here

PR #1384, opened 2025-06-30
chore(rm): address (the majority of) `dialyzer` errors

lusergit:
Based on #1378 

Fixes the majority of `mix dialyzer` errors, leaving one error in `KvStore`  insertions

This also removes the unused `rpc` modules to interact with the backend service as nobody is using it

### DIFF
--- a/apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/plug/verify_header.ex
+++ b/apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/plug/verify_header.ex
@@ -33,7 +33,11 @@ defmodule Astarte.RealmManagement.APIWeb.Plug.VerifyHeader do
   end
 
   def call(conn, opts) do
-    secret = get_secret(conn)
+    secret =
+      case fetch_secret(conn) do
+        {:ok, secret} -> secret
+        {:error, _reason} -> nil
+      end
 
     merged_opts =
       opts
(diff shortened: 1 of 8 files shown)


ArnelaL (approved):

apps/astarte_realm_management_api/lib/astarte_realm_management_api_web/plug/verify_header.ex
```diff
@@ -42,19 +46,22 @@ defmodule Astarte.RealmManagement.APIWeb.Plug.VerifyHeader do
...
+          {:ok, secret}
+
+        _secrets ->
+          _ =
+            Logger.warning(
+              "Multiple JWKs found in pem file. This configuration is not supported.",
+              tag: :multiple_jwks
+            )
+
+          {:error, :multiple_jwks}
```
noaccOS: I'm not sure this change is a good idea, as this situation can't really happen (jwk returns a list only if you give it a list)

I propose a couple alternatives
1. add a typespec to `Auth.fetch_public_key` to signal dialyzer it should always return a string
2. leave the code as it is but raise instead of returning an error, as it is an unexpected event

lusergit: I went with the first option as it required less changes to the existing behavior :+1:

PR #1387, opened 2025-07-01
ci: fail compilation with warnings 

lusergit:
`mix compile` did not accept the parameter `--warning-as-errors`, while it accepts the
parameter `--warnings-as-errors`.

This also address the compilation warnings in all the necessary astarte services


apps/astarte_appengine_api/mix.exs
```diff
@@ -106,7 +106,12 @@ defmodule Astarte.AppEngine.API.Mixfile do
       {:skogsra, "~> 2.2"},
       {:castore, "~> 0.1.0"},
       {:observer_cli, "~> 1.5"},
-      {:dialyxir, "~> 1.0", only: [:dev, :ci], runtime: false}
+      {:dialyxir, "~> 1.0", only: [:dev, :ci], runtime: false},
+      # Required to update guardian and gettext
+      {:plug_crypto, "~> 1.2",
+       [env: :prod, hex: "plug_crypto", repo: "hexpm", optional: false, override: true]},
+      {:decimal, "~> 1.7",
+       [env: :prod, hex: "decimal", repo: "hexpm", optional: true, override: true]}
```
noaccOS: why is it required? did they conflict with other dependencies?

lusergit: yas

lusergit: actually no, fixed

noaccOS (approved):

PR #1389, opened 2025-07-01
chore(rm): move to `Astarte.Cases.Data` and remove old test setup

lusergit:
Moves all `DataCase` instances to `Astarte.Cases.Data`, removing the old test setup and the dependency to realm_management in the test environment


lusergit:
Superseded (ante litteram) by #1379

PR #1415, opened 2025-07-09
fix: ensure rpc server always starts

eddbbt:
previously, we avoided adding the server as a child to the supervisor if a server existed already, leaving the job to horde to eventually create a consistent status between the supervisors.

now, we do the same while making sure the rpc server is always added as a child to the supervisor, so that it is not possible for it to just not start

Fixes #

* [ ] Yes
* [ ] No

```docs

```


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/rpc/server.ex
```diff
@@ -33,7 +33,12 @@ defmodule Astarte.DataUpdaterPlant.RPC.Server do
     name = {:via, Horde.Registry, {Registry.DataUpdaterRPC, :server}}
     opts = Keyword.put(opts, :name, name)
 
-    GenServer.start_link(__MODULE__, args, opts)
+    case GenServer.start_link(__MODULE__, args, opts) do
+      {:ok, pid} -> {:ok, pid}
+      {:error, {:already_started, pid}} -> {:ok, pid}
+      {:error, reason} -> {:error, reason}
+      :ignore -> :ignore
+    end
```
lusergit: nit:
```suggestion
    with {:error, {:already_started, pid}} <- GenServer.start_link(__MODULE__, args, opts) do
      {:ok, pid}
    end
```

lusergit (approved):
lgtm, only a small nit

noaccOS (approved):

PR #1416, opened 2025-07-09
refactor(e2e): install data triggers

emirh1:

Install data triggers in e2e tests through realm management

Fixes #

* [ ] Yes
* [x] No

```docs

```

### DIFF
--- a/tools/astarte_e2e/lib/astarte_e2e/data_trigger.ex
+++ b/tools/astarte_e2e/lib/astarte_e2e/data_trigger.ex
@@ -1,0 +1,86 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule AstarteE2E.DataTrigger do
+  require Logger
+
+  use Task
+
+  alias AstarteE2E.Config
+
+  def start_link(opts) do
+    Task.start_link(__MODULE__, :install_data_trigger!, [opts])
+  end
+
+  def install_data_trigger!(opts) do
+    device_id = Keyword.fetch!(opts, :device_id)
+    base_url = Config.realm_management_url!()
+    realm = Config.realm!()
+    astarte_jwt = Config.jwt!()
+
+    url = Path.join([base_url, "v1", realm, "triggers"])
+
+    headers = [
+      {"Accept", "application/json"},
+      {"Content-Type", "application/json"},
+      {"Authorization", "Bearer #{astarte_jwt}"}
+    ]
+
+    triggers = [
+      %{
+        name: "valuetrigger-datastream",
+        device_id: device_id,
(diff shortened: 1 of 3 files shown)


tools/astarte_e2e/lib/astarte_e2e/data_trigger.ex
```diff
@@ -0,0 +1,100 @@
...
+
+    child = %{
+      id: :install_data_trigger,
+      start: {Task, :start_link, [fn -> install_data_trigger!(device_id) end]},
+      type: :worker,
+      restart: :transient
+    }
+
+    Supervisor.init([child], strategy: :one_for_one)
+  end
```
noaccOS: you don't need to resort to a one-child supervisor, take a look at the [doc](https://hexdocs.pm/elixir/1.15.7/Task.html#module-statically-supervised-tasks)

```suggestion
  def start_link(opts) do
    Task.start_link(__MODULE__, :install_data_trigger!, [opts])
  end
```

of course this means having `install_data_trigger!` parse `opts` on its own, as it's better to crash inside the new process

tools/astarte_e2e/lib/astarte_e2e/data_trigger.ex
```diff
@@ -0,0 +1,86 @@
...
+
+defmodule AstarteE2E.DataTrigger do
+  require Logger
+
+  use Task
+
+  alias AstarteE2E.Config
+
+  def start_link(opts) do
+    Task.start_link(__MODULE__, :install_data_trigger!, opts)
```
noaccOS: I believe this should be a list of arguments, so I would expect this to be

 ```suggestion
     Task.start_link(__MODULE__, :install_data_trigger!, [opts])
```

Honestly I'm baffled this does not generate an error?

Comment (noaccOS) [reply tools/astarte_e2e/lib/astarte_e2e/data_trigger.ex]: Yeah in the logs there is in fact a crash from this process

File: tools/astarte_e2e/lib/astarte_e2e/data_trigger.ex [hunk 3]
```diff
@@ -0,0 +1,86 @@
...
+    Task.start_link(__MODULE__, :install_data_trigger!, [opts])
+  end
+
+  def install_data_trigger!(opts) do
+    device_id = Keyword.fetch!(opts, :device_id)
+    base_url = Config.realm_management_url!()
+    realm = Config.realm!()
+    astarte_jwt = Config.jwt!()
+
+    url = "#{base_url}/v1/#{realm}/triggers"
```
noaccOS: Let's use [`Path.join`](https://hexdocs.pm/elixir/1.12/Path.html#join/1), which also handles trailing slashes in path tokens

```suggestion
    url = Path.join([base_url, "v1", realm, "triggers"])
```

tools/astarte_e2e/lib/astarte_e2e/data_trigger.ex
```diff
@@ -0,0 +1,86 @@
...
+        }
+      }
+    ]
+
+    Enum.each(triggers, fn trigger ->
+      body = Jason.encode!(%{"data" => trigger})
+
+      case HTTPoison.post(url, body, headers) do
+        {:ok, %HTTPoison.Response{status_code: 201}} ->
+          {:cont, :ok}
```
noaccOS: `Enum.each` just runs a function for each element, and doesn't care about the results (`:cont`/`:halt`) of each function

You probably meant to use [`Enume.reduce_while`](https://hexdocs.pm/elixir/Enum.html#reduce_while/3)

```suggestion
    Enum.reduce_while(triggers, fn trigger, :ok ->
      body = Jason.encode!(%{"data" => trigger})

      case HTTPoison.post(url, body, headers) do
        {:ok, %HTTPoison.Response{status_code: 201}} ->
          {:cont, :ok}
```

noaccOS (changes requested):

PR #1429, opened 2025-07-14
refactor(hk): add repo migration status to the health check

Dam-99:

Housekeeping relies on the database in order to provide its service. accordingly, the health check was making sure the connection to the database is up to report the service health.

However, it was possible for the connection to the database to be up, but it still having pending migrations, which would cause any calls to the service to fail.

For this reason, a check for the status of migrations is added to the health check'

Fixes #

* [ ] Yes
* [ ] No

```docs

```


apps/astarte_housekeeping/lib/astarte_housekeeping_web/plug/health_plug.ex
```diff
@@ -66,4 +67,15 @@ defmodule Astarte.HousekeepingWeb.HealthPlug do
...
+  defp repo_is_migrated() do
+    Repo
+    |> Ecto.Migrator.migrations()
+    |> Enum.reduce_while(:ok, fn migr_status, acc ->
+      case migr_status do
+        {:up, _id, _name} -> {:cont, acc}
+        {:down, id, name} -> {:halt, {:error, {id, name}}}
+      end
+    end)
+  end
```
lusergit: This does not work. Housekeeping migrations are not actual ecto migrations, but rather hand-craffted migrations meant to be run before the start of the application.

apps/astarte_housekeeping/lib/astarte_housekeeping_web/plug/health_plug.ex
```diff
@@ -21,25 +21,26 @@ defmodule Astarte.HousekeepingWeb.HealthPlug do
...
+        :telemetry.execute(
+          [:astarte, :housekeeping, :service],
+          %{health: 1},
+          %{status: status}
+        )
 
+        conn
+        |> send_resp(:ok, "")
+        |> halt()
+      else
```
lusergit: This does not work, as `get_health/0` might also return `{:ok, %{status: :bad}}` which is not a valid connection, messing up the return type of `call`

lusergit (approved):
🚀

emirh1 (approved):

PR #1431, opened 2025-07-14
refactor(DUP): Expand health check to include VerneMQ status 

OmarBrbutovic:

Previously, the health check only verified database connectivity via DataAccess. This update adds a new check for VerneMQ availability using Horde.Registry lookup.

Fix test failure due to dependency changes causing `{:error, :interface_not_found}`.

Replace deprecated `use Plug.Test` with `import Plug.Conn` to resolve warnings.

depends on https://github.com/astarte-platform/astarte/pull/1428

* [ ] Yes
* [x] No

```docs

```

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/health.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/health.ex
@@ -1,0 +1,75 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.DataUpdaterPlant.Health.Health do
+  @moduledoc """
+  Health check module for Data Updater Plant service
+  """
+
+  alias Astarte.DataAccess.Health.Health, as: DatabaseHealth
+
+  require Logger
+
+  @doc """
+  Gets the backend health.
+  """
+  def get_health do
+    DatabaseHealth.get_health()
+    |> from_database_health()
+    |> cast_health_check(&vernemq_health/0)
+    |> tap(&emit_telemetry/1)
+  end
+
+  # Use the existing DataAccess health check for database connectivity
+  defp from_database_health(database_health) do
+    # When degraded, some database nodes are available so it's still ok
+    case database_health do
+      :ready -> :ready
+      :degraded -> :ready
+      :bad -> :bad
+      :error -> :bad
+    end
+  end
+
+  # Check if VerneMQ service is available via Horde registry lookup
+  defp vernemq_health do
+    case Horde.Registry.lookup(Registry.VMQPluginRPC, :server) do
+      [] ->
+        Logger.warning("VerneMQ RPC server not found in registry", tag: "vernemq_health_check")
+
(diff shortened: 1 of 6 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/health.ex
```diff
@@ -0,0 +1,78 @@
...
+
+      [{_pid, _value}] ->
+        :ready
+
+      _multiple_entries ->
+        Logger.warning("Multiple VerneMQ RPC servers found in registry",
+          tag: "vernemq_health_check"
+        )
+
+        :degraded
```
noaccOS: this should not be needed, even internally in the horde code there is no check for multiple results

```suggestion
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/health.ex
```diff
@@ -0,0 +1,78 @@
...
+    end
+  rescue
+    error ->
+      Logger.error("Error checking VerneMQ health: #{inspect(error)}",
+        tag: "vernemq_health_check"
+      )
+
+      :error
```
noaccOS: I don't think this function can raise?

```suggestion
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/health.ex
```diff
@@ -0,0 +1,78 @@
...
+
+  alias Astarte.DataAccess.Health.Health
+
+  require Logger
+
+  def get_health do
+    checks = %{
+      database: check_database_health(),
+      vernemq: check_vernemq_health()
+    }
```
noaccOS: one nit with this approach is that it always runs all checks. it would be better if the first failed check made all subsequent checks not run at all, but this is an optimization that can be applied later

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/health.ex
```diff
@@ -0,0 +1,78 @@
...
+    Health.get_health()
+  end
+
+  # Check if VerneMQ service is available via Horde registry lookup
+  defp check_vernemq_health() do
+    case Horde.Registry.lookup(Registry.VMQPluginRPC, :server) do
+      [] ->
+        Logger.warning("VerneMQ RPC server not found in registry", tag: "vernemq_health_check")
+
+        :error
```
noaccOS: I would use `:bad` instead of `:error` for this result but it's arguable, data_access returns `:error` for connection errors and `:bad` for xandra errors

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/health.ex
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  defp determine_overall_status(checks) do
+    statuses = Map.values(checks)
+
+    cond do
+      Enum.any?(statuses, &(&1 == :error)) -> :error
+      Enum.any?(statuses, &(&1 == :degraded)) -> :degraded
+      true -> :ready
+    end
```
noaccOS: this doesn't account for `:bad` health checks results

ArnelaL (approved):

apps/astarte_data_updater_plant/test/astarte_data_updater_plant_web/astarte_data_updater_plant_web_test.exs
```diff
@@ -23,7 +23,7 @@ defmodule Astarte.DataUpdaterPlantWeb.AstarteDataUpdaterPlantWebTest do
   import Plug.Test
 
   alias Astarte.DataUpdaterPlantWeb.Router
-  alias Astarte.DataAccess.Health.Health
```
noaccOS, line 25: this is fine but it would be even better if it was left as is and an addidional expectation was set for `Horde.Registry.lookup`, so that you can test your health logic

noaccOS (approved):

davidebriani (approved):

davidebriani (approved):

noaccOS (approved):

PR #1441, opened 2025-07-17
refactor(e2e): move realm creation to astarte_e2e

emirh1:
Create Realm directly in astarte_e2e instead of a separate CI step.

Create realm inside astarte_e2e tool, instead of CI.

Fixes #

* [ ] Yes
* [x] No

### DIFF
--- a/.github/workflows/astarte-end-to-end-test-workflow.yaml
+++ b/.github/workflows/astarte-end-to-end-test-workflow.yaml
@@ -63,15 +63,11 @@ jobs:
         wget https://github.com/astarte-platform/astartectl/releases/download/v22.11.02/astartectl_22.11.02_linux_x86_64.tar.gz
         tar xf astartectl_22.11.02_linux_x86_64.tar.gz
         chmod +x astartectl
-    - name: Create realm
+    - name: Generate realm keypair and JWT
       run: |
         ./astartectl utils gen-keypair test
-        ./astartectl housekeeping realms create test --astarte-url http://api.astarte.localhost --realm-public-key test_public.pem -k compose/astarte-keys/housekeeping_private.pem -y
-        echo "E2E_REALM=test" >> $GITHUB_ENV
-        sleep 5
-    - name: Ensure Trigger Engine started the consumer
-      # TODO: use a proper way to do this
-      run: sleep 30
+        E2E_HOUSEKEEPING_API_JWT=$(./astartectl utils gen-jwt housekeeping -k compose/astarte-keys/housekeeping_private.pem)
+        echo "E2E_HOUSEKEEPING_API_JWT=$E2E_HOUSEKEEPING_API_JWT" >> $GITHUB_ENV
     - name: Generate JWT
       run: |
         JWT=$(./astartectl utils gen-jwt appengine channels pairing realm-management -k test_private.pem)
@@ -82,6 +78,7 @@ jobs:
         E2E_HOST: "astarte-e2e"
         E2E_PAIRING_URL: http://api.astarte.localhost/pairing
         E2E_APPENGINE_URL: http://api.astarte.localhost/appengine
+        E2E_HOUSEKEEPING_URL: http://api.astarte.localhost/housekeeping
         E2E_REALM_MANAGEMENT_URL: http://api.astarte.localhost/realmmanagement
         E2E_AMQP_CONSUMER_HOST: rabbitmq
         E2E_IGNORE_SSL_ERRORS: true
@@ -89,7 +86,13 @@ jobs:
         E2E_CHECK_REPETITIONS: 5
         E2E_MAILER_TO_ADDRESS: mail@example.com
         E2E_MAIL_SUBJECT: "Subj: Astarte Notification"
-      run: docker run --env-file=<(env | grep '^E2E_') --network astarte --hostname "$E2E_HOST" astarte-e2e
+      run: |
+       docker run \
+        -e E2E_REALM_PUBLIC_KEY_PEM="$(cat ../../test_public.pem)" \
+        --env-file=<(env | grep '^E2E_' | grep -v '^E2E_REALM_PUBLIC_KEY_PEM=') \
+        --network astarte \
+        --hostname "$E2E_HOST" \
+        astarte-e2e
     - name: Check Docker
       if: ${{ failure() }}
       run: docker compose logs
--- a/tools/astarte_e2e/lib/astarte_e2e/application.ex
+++ b/tools/astarte_e2e/lib/astarte_e2e/application.ex
@@ -20,15 +20,22 @@ defmodule AstarteE2E.Application do
   use Application
 
   alias AstarteE2E.Config
+  alias AstarteE2E.Realm
   alias AstarteE2E.ServiceNotifier
 
   require Logger
+
+  @trigger_engine_consumer_tracker_cycle_duration :timer.seconds(30)
 
   @impl true
   def start(_type, _args) do
     Logger.info("Starting AstarteE2E application.", tag: "application_start")
 
-    with :ok <- Config.validate() do
+    with :ok <- Config.validate(),
+         :ok <- Realm.create_realm!() do
+      # ensure trigger engine started the consumer for the new realm
+      :timer.sleep(@trigger_engine_consumer_tracker_cycle_duration)
+
       children = [
         {Registry, keys: :unique, name: Registry.AstarteE2E},
         AstarteE2EWeb.Telemetry,
--- a/tools/astarte_e2e/lib/astarte_e2e/config.ex
+++ b/tools/astarte_e2e/lib/astarte_e2e/config.ex
@@ -24,6 +24,7 @@ defmodule AstarteE2E.Config do
   alias AstarteE2E.Config.ListOfStrings
   alias AstarteE2E.Config.BambooMailAdapter
   alias AstarteE2E.Config.NormalizedMailAddress
+  alias AstarteE2E.Config.JWTPublicKeyPEMType
 
   @type client_option ::
           {:url, String.t()}
@@ -51,6 +52,12 @@ defmodule AstarteE2E.Config do
     type: :binary,
     required: true
 
+  @envdoc "Astarte Housekeeping URL (e.g. https://api.astarte.example.com/housekeeping)."
+  app_env :housekeeping_url, :astarte_e2e, :housekeeping_url,
+    os_env: "E2E_HOUSEKEEPING_URL",
+    type: :binary,
+    required: true
+
   @envdoc "Ignore SSL errors. Defaults to false. Changing the value to true is not advised for production environments unless you're aware of what you're doing."
   app_env :ignore_ssl_errors, :astarte_e2e, :ignore_ssl_errors,
     os_env: "E2E_IGNORE_SSL_ERRORS",
@@ -73,7 +80,8 @@ defmodule AstarteE2E.Config do
   app_env :realm, :astarte_e2e, :realm,
     os_env: "E2E_REALM",
     type: :binary,
-    required: true
+    default: "test",
+    required: false
 
   @envdoc "The Astarte JWT employed to access Astarte APIs. The token can be generated with: `$ astartectl utils gen-jwt <service> -k <your-private-key>.pem`."
   app_env :jwt, :astarte_e2e, :jwt,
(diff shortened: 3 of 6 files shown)


ArnelaL (commented):
- The command for creating a realm should be removed from the astarte-end-to-end-test-workflow file, as it is no longer necessary.

- Although the CI passes, the request to create a realm returns forbidden, causing Match Error to happen. This happens because the token used for authentication is scoped only for AppEngine and Channels. To resolve this, the token generation should be updated to include permissions for Housekeeping as well.

noaccOS:
> Given this conflict, I believe this PR should be postponed until the interface installation is fully moved to astarte_e2e, which will make the realm creation in the workflow unnecessary

Yeah, I see 2 options here
- wait for the interface installation pr to be merged first
- change this pr to create _another_ realm, just to prove this works, then we can revert it back to create the correct realm once the interface installation is in

.github/workflows/astarte-end-to-end-test-workflow.yaml
```diff
@@ -63,11 +63,13 @@ jobs:
...
-    - name: Create realm
+    - name: Generate realm keypair and JWT
       run: |
         ./astartectl utils gen-keypair test
-        ./astartectl housekeeping realms create test --astarte-url http://api.astarte.localhost --realm-public-key test_public.pem -k compose/astarte-keys/housekeeping_private.pem -y
-        echo "E2E_REALM=test" >> $GITHUB_ENV
+        E2E_HOUSEKEEPING_API_JWT=$(./astartectl utils gen-jwt housekeeping -k compose/astarte-keys/housekeeping_private.pem)
+        echo "E2E_HOUSEKEEPING_API_JWT=$E2E_HOUSEKEEPING_API_JWT" >> $GITHUB_ENV
+
+        echo "E2E_REALM_PUBLIC_KEY_PEM=$(pwd)/test_public.pem" >> $GITHUB_ENV
```
noaccOS: this won't work, because the e2e test is run inside docker and has no access to external files (unless you mount them)
I suggest you just use the file contents and call it a day

```suggestion
        echo "E2E_REALM_PUBLIC_KEY_PEM=$(cat ./test_public.pem)" >> $GITHUB_ENV
```

tools/astarte_e2e/lib/astarte_e2e/application.ex
```diff
@@ -33,6 +34,7 @@ defmodule AstarteE2E.Application do
         {Registry, keys: :unique, name: Registry.AstarteE2E},
         AstarteE2EWeb.Telemetry,
         {ServiceNotifier, Config.notifier_opts()},
+        {Realm, Config.realm_opts()},
```
noaccOS: placing this here means you have no guarantee on the order of operations between realm creation and task scheduler execution. realm creation should be performed synchronously before everything else, either here or in the task scheduler

.github/workflows/astarte-end-to-end-test-workflow.yaml
```diff
@@ -63,11 +63,17 @@ jobs:
...
         ./astartectl utils gen-keypair test
-        ./astartectl housekeeping realms create test --astarte-url http://api.astarte.localhost --realm-public-key test_public.pem -k compose/astarte-keys/housekeeping_private.pem -y
-        echo "E2E_REALM=test" >> $GITHUB_ENV
+        E2E_HOUSEKEEPING_API_JWT=$(./astartectl utils gen-jwt housekeeping -k compose/astarte-keys/housekeeping_private.pem)
+        echo "E2E_HOUSEKEEPING_API_JWT=$E2E_HOUSEKEEPING_API_JWT" >> $GITHUB_ENV
+
+        {
+        echo -n "E2E_REALM_PUBLIC_KEY_PEM="
+        awk '{printf "%s\\n", $0}' ./test_public.pem
+        } >> $GITHUB_ENV
```
noaccOS: Well not sure why you need to take out `awk`, I believe you just need to escape the multiline string when you assign it to the variable. newlines should not be a problem inside variables as long as bash knows it's a string

```suggestion
        echo "E2E_REALM_PUBLIC_KEY_PEM='$(cat ./test_public.pem)'" >> $GITHUB_ENV
```

example: 

```bash
$ echo 'test
  second
  line' > testfile

$ MYVAR="$(cat testfile)"

$ echo "$MYVAR"
test
second
line
```

emirh1: Multiline string env variables cannot be defined on github like this: 
`echo "E2E_REALM_PUBLIC_KEY_PEM='$(cat ./test_public.pem)'" >> $GITHUB_ENV`

I get this error:
`Unable to process file command 'env' successfully.`

noaccOS: yeah, you're right. have you already tried the method used in the [documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-of-a-multiline-string)?

emirh1: yes, but it only saves the first line of public key in env variable like this: 
` E2E_REALM_PUBLIC_KEY_PEM: -----BEGIN PUBLIC KEY-----`.

noaccOS: I made a simple proof of concept and it looks like it should work: https://github.com/noaccOS/test-ci/actions/runs/16941819871/job/48012657697

emirh1: I think actually grep is causing the issue here. Multiline env variable is stored correctly, but grep matches only the first line, and truncates the variable in this command:
`run: docker run --env-file=<(env | grep '^E2E_') --network astarte --hostname "$E2E_HOST" astarte-e2e `

.github/workflows/astarte-end-to-end-test-workflow.yaml
```diff
@@ -63,11 +63,12 @@ jobs:
...
+    - name: Generate realm keypair and JWT
       run: |
         ./astartectl utils gen-keypair test
-        ./astartectl housekeeping realms create test --astarte-url http://api.astarte.localhost --realm-public-key test_public.pem -k compose/astarte-keys/housekeeping_private.pem -y
-        echo "E2E_REALM=test" >> $GITHUB_ENV
+        E2E_HOUSEKEEPING_API_JWT=$(./astartectl utils gen-jwt housekeeping -k compose/astarte-keys/housekeeping_private.pem)
+        echo "E2E_HOUSEKEEPING_API_JWT=$E2E_HOUSEKEEPING_API_JWT" >> $GITHUB_ENV
+        
         sleep 5
     - name: Ensure Trigger Engine started the consumer
```
noaccOS: this delay should be removed, as it was added to the e2e project instead

.github/workflows/astarte-end-to-end-test-workflow.yaml
```diff
@@ -82,14 +83,21 @@ jobs:
         E2E_HOST: "astarte-e2e"
         E2E_PAIRING_URL: http://api.astarte.localhost/pairing
         E2E_APPENGINE_URL: http://api.astarte.localhost/appengine
+        E2E_HOUSEKEEPING_URL: http://api.astarte.localhost/housekeeping
         E2E_REALM_MANAGEMENT_URL: http://api.astarte.localhost/realmmanagement
         E2E_AMQP_CONSUMER_HOST: rabbitmq
         E2E_IGNORE_SSL_ERRORS: true
         E2E_CHECK_INTERVAL_SECONDS: 5
-        E2E_CHECK_REPETITIONS: 5
+        E2E_CHECK_REPETITIONS: 20
```
noaccOS: why was this changed?

tools/astarte_e2e/lib/astarte_e2e/config.ex
```diff
@@ -69,12 +76,6 @@ defmodule AstarteE2E.Config do
     type: :binary,
     required: true
 
-  @envdoc "Realm name."
-  app_env :realm, :astarte_e2e, :realm,
-    os_env: "E2E_REALM",
-    type: :binary,
-    required: true
-
```
noaccOS: having the realm name be customizable might still be useful in case we need to run the e2e project on an already existing astarte instance, which may already have a `"test"` realm. I suggest we keep the old behavior here

tools/astarte_e2e/lib/astarte_e2e/task_scheduler.ex
```diff
@@ -35,10 +36,12 @@ defmodule AstarteE2E.TaskScheduler do
   @impl GenServer
   def init(_init_arg) do
     Process.flag(:trap_exit, true)
+
     interfaces = Interface.generate_interfaces!()
     :ok = Interface.install_interfaces!(interfaces)
 
     opts = [interfaces: interfaces]
+    :timer.sleep(5000)
```
noaccOS: why do we need this now?

tools/astarte_e2e/lib/astarte_e2e/volatile_trigger_roundtrip/executor.ex
```diff
@@ -56,6 +56,7 @@ defmodule AstarteE2E.VolatileTriggerRoundtrip.Executor do
       {:ok, client} = Client.start_link(client_opts)
       Client.wait_for_connection(realm, encoded_id)
 
+      :timer.sleep(5000)
```
noaccOS: same as above

tools/astarte_e2e/lib/astarte_e2e/application.ex
```diff
@@ -28,7 +29,10 @@ defmodule AstarteE2E.Application do
   def start(_type, _args) do
     Logger.info("Starting AstarteE2E application.", tag: "application_start")
 
-    with :ok <- Config.validate() do
+    with :ok <- Config.validate(),
+         :ok <- Realm.create_realm!() do
+      :timer.sleep(30000)
```
noaccOS: let's have a module attribute to describe why we need this

```elixir
  @trigger_engine_consumer_tracker_cycle_duration :timer.seconds(30)
```

then in the usage

```elixir
      # ensure trigger engine started the consumer for the new realm
      :timer.sleep(@trigger_engine_consumer_tracker_cycle_duration)
```

tools/astarte_e2e/lib/astarte_e2e/task_scheduler.ex
emirh1: I forgot to remove this

tools/astarte_e2e/lib/astarte_e2e/config.ex
```diff
@@ -177,6 +185,16 @@ defmodule AstarteE2E.Config do
     os_env: "E2E_MAIL_SERVICE",
     type: BambooMailAdapter
 
+  @envdoc "The JWT public key."
+  app_env :jwt_public_key_pem, :astarte_e2e, :jwt_public_key_pem,
+    os_env: "E2E_REALM_PUBLIC_KEY_PEM",
```
noaccOS, line 190: the end to end tests won't run if the jwt is not specified

```suggestion
    os_env: "E2E_REALM_PUBLIC_KEY_PEM",
    required: true,
```

tools/astarte_e2e/lib/astarte_e2e/config.ex
```diff
@@ -177,6 +185,16 @@ defmodule AstarteE2E.Config do
...
+    type: JWTPublicKeyPEMType
+
+  @envdoc "JWT for housekeeping authentication."
+  app_env :housekeeping_jwt, :astarte_e2e, :housekeeping_jwt,
+    os_env: "E2E_HOUSEKEEPING_API_JWT",
```
noaccOS, line 196:
```suggestion
    os_env: "E2E_HOUSEKEEPING_API_JWT",
    required: true
```

.github/workflows/astarte-end-to-end-test-workflow.yaml
```diff
@@ -63,15 +63,12 @@ jobs:
...
         chmod +x astartectl
-    - name: Create realm
+    - name: Generate realm keypair and JWT
       run: |
         ./astartectl utils gen-keypair test
-        ./astartectl housekeeping realms create test --astarte-url http://api.astarte.localhost --realm-public-key test_public.pem -k compose/astarte-keys/housekeeping_private.pem -y
-        echo "E2E_REALM=test" >> $GITHUB_ENV
+        E2E_HOUSEKEEPING_API_JWT=$(./astartectl utils gen-jwt housekeeping -k compose/astarte-keys/housekeeping_private.pem)
+        echo "E2E_HOUSEKEEPING_API_JWT=$E2E_HOUSEKEEPING_API_JWT" >> $GITHUB_ENV
         sleep 5
```
noaccOS: I think this sleep also serves no use anymore, as we're only creating the jwts before starting the project

tools/astarte_e2e/lib/astarte_e2e/task_scheduler.ex
```diff
@@ -35,6 +35,7 @@ defmodule AstarteE2E.TaskScheduler do
   @impl GenServer
   def init(_init_arg) do
     Process.flag(:trap_exit, true)
+
```
eddbbt, line 38: extra newline

PR #1447, opened 2025-07-18
fix(AE): Return correct 401 error for non-existent realms 

OmarBrbutovic:

Fix:
- When calling any AppEngine API with a non-existent realm, Astarte was returning an error that differed from the one specified in the [official documentation](https://docs.astarte-platform.org/astarte/latest/api/index.html#/device/getInterfaces)
- Updated the `fallback_controller` to correctly return a 401, ensuring  behavior aligns with documented expectations
- Fix existing and add a new test for this case

Refactor:
- Updated `fetch_public_key` to use `Repo.safe_fetch_one` from `DataAccess` instead of reimplementing already existing functionality
- Move `queries` module to `auth` namespace since it is only used there
- Added tests

* [ ] Yes
* [x] No

```docs

```


apps/astarte_appengine_api/lib/astarte_appengine_api/auth/queries.ex
```diff
@@ -0,0 +1,53 @@
...
+
+      {:error, :not_found} ->
+        {:error, :public_key_not_found}
+
+      {:error, :realm_not_found} ->
+        {:error, :not_existing_realm}
+
+      {:error, error} ->
+        {:error, error}
+    end
```
lusergit: `error` is a valid key in the `fetch_*` functions, so that you can

```suggestion
    opts = [
      uuid_format: :binary, 
      consistency: Consistency.domain_model(:read), 
      error: :public_key_not_found
    ]

    with {:error, :realm_not_found} <- Repo.safe_fetch_one(schema_query, opts) do
        {:error, :not_existing_realm}
    end
```

davidebriani: Given also that DataAccess supports a custom `error` option (https://github.com/astarte-platform/astarte_data_access/blob/81800c1627c3ef2b33ad6254532b0a2e161054fa/lib/astarte_data_access/repo.ex#L123), we can do just:

```suggestion
    opts = [uuid_format: :binary, consistency: Consistency.domain_model(:read), error: :public_key_not_found]

    Repo.safe_fetch_one(schema_query, opts)
```

and we can catch `:realm_not_found` in the fallback controller.

davidebriani (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api_web/controllers/fallback_controller.ex
```diff
@@ -212,22 +214,52 @@ defmodule Astarte.AppEngine.APIWeb.FallbackController do
...
-  # In all other cases, we reply with 403
-  def auth_error(conn, _reason, _opts) do
-    conn
-    |> put_status(:forbidden)
-    |> put_view(Astarte.AppEngine.APIWeb.ErrorView)
-    |> render(:"403")
+  # This is called when the JWT token is present but the user is not authorized
+  def auth_error(conn, reason, _opts) do
+    # Check if this is due to a non-existing realm and respond with 401
+    if is_non_existing_realm_error?(conn) do
```
noaccOS: Is this needed? Isn't this only called by guardian when the jwt is invalid (and not for actual realms not existing)?

noaccOS (commented):
Do the changes to the fallback controller also handle the situation where a xandra error with `Keyspace realmname does not exist` is thrown? It should be handled just like in [data access](https://github.com/astarte-platform/astarte_data_access/blob/81800c1627c3ef2b33ad6254532b0a2e161054fa/lib/astarte_data_access/repo.ex#L171)

apps/astarte_appengine_api/lib/astarte_appengine_api_web/plug/verify_header.ex
```diff
@@ -14,13 +17,20 @@ defmodule Astarte.AppEngine.APIWeb.Plug.VerifyHeader do
...
+        |> Keyword.merge(secret: secret)
 
-    GuardianVerifyHeader.call(conn, merged_opts)
+      GuardianVerifyHeader.call(conn, merged_opts)
+    catch
+      {:halt_with_error, error} ->
+        conn
+        |> FallbackController.call(error)
+        |> halt()
+    end
```
noaccOS: nit: prefer matching on errors instead of throwing

```elixir
def call(conn, opts) do
  with {:ok, secret} <- fetch_secret(conn) do
    # ...
    GuardianVerifyHeader.call()
  end
end

def fetch_secret(conn) do
  %{"realm_name" => realm} = conn.path_params

  with {:ok, public_key} <- Auth.fetch_public_key(realm) do
    JWK.from_pem(public_key)
  end
end

noaccOS (approved):

PR #1459, opened 2025-07-24
chore(PA): Install credo dependency 

OmarBrbutovic:

Add credo to the project. Credo is a static code analysis tool that can suggest refactoring opportunities for complex code, inconsistencies and common mistakes. The aim is to enforce a better coding style.

* [ ] Yes
* [x] No

```docs

```


apps/astarte_pairing/.credo.ex
lusergit, line 1: I missed this change in the last PR adding `credo` to housekeeping, but this is not the credo default, we should comply with the default and then use the proper command in CI to check what we want.

apps/astarte_pairing/lib/astarte_pairing/engine.ex
```diff
@@ -170,10 +170,10 @@ defmodule Astarte.Pairing.Engine do
          {:ok, registered_devices_count} <- Queries.fetch_registered_devices_count(realm_name) do
       if registration_limit != nil and registered_devices_count >= registration_limit do
         _ =
-          Logger.warning("Cannot register device: reached device registration limit",
+          Logger.warning("Cannot register device: reached device registration limit", %{
             realm_name: realm_name,
             tag: "device_registration_limit_reached"
-          )
+          })
```
lusergit: I think that rather than changing the keyword list to a map here we should change the logger's config in `config/config.ex`:

```elixir
# Configures Elixir's Logger
config :logger, :default_formatter,
  format: "$time $metadata[$level] $message\n",
  metadata: [:request_id, :tag, :realm]
```

and instead of `realm_name` in logs just `realm` is fine

lusergit (changes requested):

Dam-99 (approved):

PR #1468, opened 2025-07-29
feat(hk): add support for maps in HOUSEKEEPING_ASTARTE_KEYSPACE_REPLI…

eddbbt:


emirh1 (approved):

apps/astarte_housekeeping/lib/astarte_housekeeping/config/network_type.ex
```diff
@@ -0,0 +1,50 @@
...
+      {int, ""} ->
+        {:ok, int}
+
+      _ ->
+        case Jason.decode(value) do
+          {:ok, decoded} -> {:ok, decoded}
+          _ -> {:ok, value}
+        end
+    end
+  end
```
noaccOS: Instead of this dynamic typing approach, how about keeping the api consistent with the [realm creation](https://github.com/astarte-platform/astarte/blob/a1f9ffd0036cfefb11ae7804064c31e896f33dde/apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realm.ex#L101)?

- One env to choose between simple and network topology strategy
- the pre-existing `HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_FACTOR` remains an integer and is used in case of simple strategy
- One env used for the datacenter replication. It should not only be decoded from json, but also validated to be a map `%{non_empty_string => integer}`

eddbbt: I'll rework the realm creation instead, imho it's a lot cleaner to have a single env/param and differentiate later (also easier to support for the operator, instead of having 3 different variables)

map validation is out of scope for skosgra, and the function  `build_replication_map_str` already validate it (and crash with a meaningful message in case).

noaccOS: That would imply a breaking change. Let's keep the current behavior for now, as it also gives a cleaner data definition on the operator side of things

Also, while you're correct in that the right now the program still crashes with invalid data, it's always better to validate your input at the api level so that you can trust your data throughout the rest of the code

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -51,14 +51,29 @@ defmodule Astarte.Housekeeping.Config do
     type: :boolean,
     default: false
 
-  @envdoc "Replication factor for the astarte keyspace, defaults to 1"
+  @envdoc "Decide replication factor, between :simple and :network, defaults to :simple"
```
noaccOS:
```suggestion
  @envdoc "Replication strategy for the `astarte` keyspace, either `simple` or `network`. Defaults to `simple`"
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -51,14 +51,29 @@ defmodule Astarte.Housekeeping.Config do
...
+  app_env :astarte_keyspace_replication_mode,
+          :astarte_housekeeping,
+          :astarte_keyspace_replication_mode,
+          os_env: "HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_MODE",
+          type: Astarte.Housekeeping.Config.NetworkMode,
```
noaccOS:
```suggestion
  app_env :astarte_keyspace_replication_strategy,
          :astarte_housekeeping,
          :astarte_keyspace_replication_strategy,
          os_env: "HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_STRATEGY",
          type: Astarte.Housekeeping.Config.ReplicationStrategy,
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -51,14 +51,29 @@ defmodule Astarte.Housekeeping.Config do
...
+          default: :simple
+
+  @envdoc "Simple replication factor for the astarte keyspace, defaults to 1"
```
noaccOS:
```suggestion
  @envdoc "Replication factor for the astarte keyspace, used when simple strategy is used for the astarte keyspace. defaults to 1"
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -51,14 +51,29 @@ defmodule Astarte.Housekeeping.Config do
...
   app_env :astarte_keyspace_replication_factor,
           :astarte_housekeeping,
           :astarte_keyspace_replication_factor,
           os_env: "HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_FACTOR",
           type: :integer,
           default: 1
 
+  @envdoc "Network Topology replication factor for the astarte keyspace"
```
noaccOS:
```suggestion
  @envdoc "Replication map for the astarte keyspace, used when network topology strategy is used for the astarte keyspace."
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -51,14 +51,29 @@ defmodule Astarte.Housekeeping.Config do
...
+  app_env :astarte_keyspace_network_replication_factor,
+          :astarte_housekeeping,
+          :astarte_keyspace_network_replication_factor,
+          os_env: "HOUSEKEEPING_ASTARTE_KEYSPACE_NETWORK_REPLICATION_FACTOR",
+          type: Astarte.Housekeeping.Config.NetworkType
```
noaccOS:
```suggestion
  app_env :astarte_keyspace_network_replication_map,
          :astarte_housekeeping,
          :astarte_keyspace_network_replication_map,
          os_env: "HOUSEKEEPING_ASTARTE_KEYSPACE_NETWORK_REPLICATION_MAP",
          type: Astarte.Housekeeping.Config.NetworkReplicationMap
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config/network_mode.ex
```diff
@@ -0,0 +1,43 @@
...
+defmodule Astarte.Housekeeping.Config.NetworkMode do
+  @moduledoc """
+  The network topology to acccept, either an integer or a binary json
+  """
+  use Skogsra.Type
+
+  @impl Skogsra.Type
+  def cast(value)
+
+  def cast(value) when is_atom(value) do
```
noaccOS: In `Astarte.Housekeeping.Config.TelemetryType`, the `cast` function expects to receive a string.
I'm not sure which one is correct, but I would expect this module to be functionally identical to `TelemetryType`

apps/astarte_housekeeping/lib/astarte_housekeeping/config/network_mode.ex
```diff
@@ -0,0 +1,43 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Astarte.Housekeeping.Config.NetworkMode do
+  @moduledoc """
+  The network topology to acccept, either an integer or a binary json
+  """
```
noaccOS:
```suggestion
  @moduledoc """
  The replication strategy to use for the astarte keyspace
  """
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config/network_type.ex
```diff
@@ -0,0 +1,40 @@
...
+
+  @impl Skogsra.Type
+  def cast(value)
+
+  def cast(value) when is_binary(value) do
+    with {:ok, %{"" => int}} <- Jason.decode(value),
+         false <- is_integer(int) do
+      :error
+    end
+  end
```
noaccOS: aside from the fact that this code snippet can potentially return `true` (eg if `value` is `{"": "hello"}`) or `{:error, %Jason.DecodeError{}}` (and skogsra only accepts `:error`), this is quite hard to parse for my brain, I suggest using the `with` block like a monad (or a changeset in elixir terms). The best code is not always the one with the least number of lines of code

```suggestion
  def cast(value) when is_binary(value) do
    with {:ok, replication_map} <- parse_replication_map(value),
         :ok <- validate_replication_map(replication_map) do
      {:ok, replication_map}
    end
  end

  @spec parse_replication_map(String.t()) :: :ok | :error
  defp parse_replication_map(replication_map_str) do
    with {:ok, decoded_json} <- Jason.decode(replication_map_str),
         true <- is_map(decoded_json) do
      {:ok, decoded_json}
    else
      _ -> :error
    end
  end

  @spec validate_replication_map(map()) :: :ok | :error
  defp validate_replication_map(replication_map) do
    all_keys_nonempty_strings? = replication_map |> Map.keys() |> Enum.all?(&nonempty_string?/1)
    all_values_positive_integers? = replication_map |> Map.values() |> Enum.all?(&positive_integer?/1)

    valid_replication_map? = all_keys_nonempty_strings? and all_values_positive_integers?

    case valid_replication_map? do
      true -> :ok
      false -> :error
    end
  end

  defp nonempty_string?(string) when is_binary(string), do: String.length(string) > 0
  defp nonempty_string?(_other), do: false

  defp positive_integer?(integer) when is_integer(integer), do: integer > 0
  defp positive_integer?(_other), do: false
```

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
```diff
@@ -318,24 +317,40 @@ defmodule Astarte.Housekeeping.Realms.Queries do
...
+    astarte_keyspace_simple_replication = Config.astarte_keyspace_replication_factor!()
+    astarte_keyspace_network_replication = Config.astarte_keyspace_network_replication_factor!()
+    astarte_keyspace_replication_mode = Config.astarte_keyspace_replication_mode!()
+
+    case astarte_keyspace_replication_mode do
+      :simple -> build_replication_map_str(astarte_keyspace_simple_replication)
+      :network -> build_replication_map_str(astarte_keyspace_network_replication)
+      _ -> raise {:error, :invalid_replication_mode}
+    end
+  end
```
noaccOS: as this code should be generic, let's avoid adding parameters only used in the specific case of the astarte keyspace creation. this kind of logic should belong to `create_astarte_keyspace` (or a private helper function used by it)

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
```diff
@@ -318,24 +317,40 @@ defmodule Astarte.Housekeeping.Realms.Queries do
...
+  defp build_replication_map_str() do
+    astarte_keyspace_simple_replication = Config.astarte_keyspace_replication_factor!()
+    astarte_keyspace_network_replication = Config.astarte_keyspace_network_replication_factor!()
+    astarte_keyspace_replication_mode = Config.astarte_keyspace_replication_mode!()
+
+    case astarte_keyspace_replication_mode do
+      :simple -> build_replication_map_str(astarte_keyspace_simple_replication)
+      :network -> build_replication_map_str(astarte_keyspace_network_replication)
+      _ -> raise {:error, :invalid_replication_mode}
+    end
```
noaccOS: you're eagerly reading these variables, but the network replication map could not exist in case of simple replication.
instead, these values should be lazily read when we know which replication strategy we're using

```suggestion
    astarte_keyspace_replication_mode = Config.astarte_keyspace_replication_mode!()

    case astarte_keyspace_replication_mode do
      :simple -> Config.astarte_keyspace_replication_factor!() |> build_replication_map_str()
      :network -> Config.astarte_keyspace_network_replication_factor!() |> build_replication_map_str()
      _ -> raise {:error, :invalid_replication_mode}
    end
```

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
```diff
@@ -318,24 +317,40 @@ defmodule Astarte.Housekeeping.Realms.Queries do
...
 
+  defp build_replication_map_str() do
+    astarte_keyspace_simple_replication = Config.astarte_keyspace_replication_factor!()
+    astarte_keyspace_network_replication = Config.astarte_keyspace_network_replication_factor!()
+    astarte_keyspace_replication_mode = Config.astarte_keyspace_replication_mode!()
+
+    case astarte_keyspace_replication_mode do
+      :simple -> build_replication_map_str(astarte_keyspace_simple_replication)
+      :network -> build_replication_map_str(astarte_keyspace_network_replication)
+      _ -> raise {:error, :invalid_replication_mode}
```
noaccOS: you don't need this check as you validate the replication strategy at skogsra level

```suggestion
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config/network_type.ex
eddbbt: skosgra does not refuse to run or give any particular exception when an invalid value is provided as a return of a cast function(invalid values are always handled https://github.com/gmtprime/skogsra/blob/v2.5.0/lib/skogsra/type.ex) , so both solutions are feature equivalent.
Evidence: 
results of my cast
image (2)

results of your cast 

image



this is the env var used for the test
```
HOUSEKEEPING_ASTARTE_KEYSPACE_NETWORK_REPLICATION_MAP={"": "hello"}
```


Despite that, I agree that it can be written in a more readable way, and I'm open for suggestions (something more concise maybe?

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -94,6 +109,28 @@ defmodule Astarte.Housekeeping.Config do
...
+      {:ok, nil} ->
+        raise "No HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_FACTOR or HOUSEKEEPING_ASTARTE_KEYSPACE_NETWORK_REPLICATION_MAP have been found, set accordingly with HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_STRATEGY value"
+
+      {:ok, _key} ->
+        :ok
+
+      _ ->
+        raise "No HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_FACTOR or HOUSEKEEPING_ASTARTE_KEYSPACE_NETWORK_REPLICATION_MAP have been found, set accordingly with HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_STRATEGY value"
+    end
+  end
```
noaccOS: - I believe the variables can never error out, as they have a system env var associated with them (and so they default to `""`). Though, I accept this check as it's futureproof (but I'd be fine even if it wasn't there)
- you should be able to infer which variable has an invalid value
- the error messages are incorrect, as they say the value is unset when it really is invalid
- Technicaly, the replication strategy should also be validated

I suggest doing something like:
```elixir
def validate_astarte_replication! do
  case astarte_keyspace_replication_strategy!() do
    :simlpe -> validate_astarte_replication_factor!()
    :network -> validate_astarte_replication_map!()
    nil -> raise "Invalid replication strategy set for the astarte keyspace"
  end
end

defp validate_astarte_replication_factor! do
  case astarte_keyspace_replication_factor() do
    {:ok, nil} -> raise "Invalid replication factor for the astarte keyspace. Check the value of HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_FACTOR"
    {:ok, _} -> :ok
    {:error, _} -> raise "HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_FACTOR must be set when using simple replication strategy for the astarte keyspace"
  end
end

# similar for replication map
```

You may also abstract this behavior and do something more similar to your current solution (maybe returning a tuple `{env_var, replication}`?) but IMO it's more organized if we separate their checks

apps/astarte_housekeeping/lib/astarte_housekeeping/config/network_replication_map_type.ex
```diff
@@ -0,0 +1,40 @@
...
+
+  @impl Skogsra.Type
+  def cast(value)
+
+  def cast(value) when is_binary(value) do
+    with {:ok, %{"" => int}} <- Jason.decode(value),
+         false <- is_integer(int) do
+      :error
+    end
+  end
```
noaccOS: 1. The [spec](https://hexdocs.pm/skogsra/Skogsra.Type.html#cast/2) says: `:: {:ok, term()} | :error`. Please let's stick to it's expected return type to avoid unforseen issues
2. this only validates the value of the empty string key to be an integer and does not validate all keys to be strings

noaccOS (changes requested):

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
eddbbt: unfortunately ` {:ok, nil}` does not pass dialyzer, other than that, ack

noaccOS: It should though. I don't have the context but the problem might be that dialyzer always finds the value to be `{:ok, nil}` and so it assumes the function can never not crash. It should be inspected further

ArnelaL (approved):

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -94,6 +109,37 @@ defmodule Astarte.Housekeeping.Config do
...
+      :simple -> validate_astarte_replication_factor!()
+      :network -> validate_astarte_replication_map!()
+      nil -> raise "Invalid replication strategy set for the astarte keyspace"
+    end
+  end
+
+  defp validate_astarte_replication_factor! do
+    case astarte_keyspace_replication_factor() do
+      {:ok, _} ->
+        :ok
```
noaccOS: this still needs to validate the replication factor is not `nil` (ie: the cast was successful)

```suggestion
      {:ok, replication_factor} when replication_factor != nil ->
        :ok
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -94,6 +109,37 @@ defmodule Astarte.Housekeeping.Config do
...
+
+      _ ->
+        raise "Invalid or empty replication factor for the astarte keyspace. Check the value of HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_FACTOR"
```
noaccOS: technically the "empty" case is not a problem here as the default would be used. it is indeed a problem for the replication map though

Let's also make it clear that we're expecting simple replication to be used here

```suggestion
        raise "Invalid replication factor for the astarte keyspace with simple replication strategy. Check the values of HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_STRATEGY and HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_FACTOR"
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -94,6 +109,37 @@ defmodule Astarte.Housekeeping.Config do
...
+    end
+  end
+
+  defp validate_astarte_replication_map! do
+    case astarte_keyspace_network_replication_map() do
+      {:ok, _} ->
+        :ok
```
noaccOS:
```suggestion
      {:ok, replication_map} when replication_map != nil ->
        :ok
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -94,6 +109,37 @@ defmodule Astarte.Housekeeping.Config do
...
+
+      _ ->
+        raise "Invalid or empty replication map for the astarte keyspace. Check the value of HOUSEKEEPING_ASTARTE_KEYSPACE_NETWORK_REPLICATION_MAP"
```
noaccOS:
```suggestion
        raise "Invalid or empty replication map for the astarte keyspace with network topology replication strategy. Check the values of HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_STRATEGY and HOUSEKEEPING_ASTARTE_KEYSPACE_NETWORK_REPLICATION_MAP"
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config/network_replication_map_type.ex
```diff
@@ -0,0 +1,48 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Astarte.Housekeeping.Config.NetworkReplicationMap do
+  @moduledoc """
+  The network topology to acccept, either an integer or a binary json
+  """
```
noaccOS:
```suggestion
  @moduledoc """
  The network replication map for a scylladb keyspace, a map of non-empty strings to positive integers
  """
```

noaccOS (changes requested):
nit: would also be great to add unit tests for the new types, to make sure they work as expected

apps/astarte_housekeeping/test/astarte_housekeeping/config/network_replication_map_test.exs
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    test "empty value" do
+      assert :error = NetworkReplicationMap.cast("")
+    end
+
+    test "nil" do
+      assert :error = NetworkReplicationMap.cast(nil)
+      assert :error = NetworkReplicationMap.cast("nil")
+    end
```
noaccOS: let's add some context here

```suggestion
  describe "cast/1 does not cast" do
    test "negative integers as values" do
      assert :error = NetworkReplicationMap.cast("{\"datacenter1\": -1}")
    end

    test "empty keys" do
      assert :error = NetworkReplicationMap.cast("{\"\": 1}")
    end

    test "non-integer values" do
      assert :error = NetworkReplicationMap.cast("{\"datacenter1\": \"invalid value\"}")
    end

    test "non-objects as top-level elements" do
      assert :error = NetworkReplicationMap.cast("[{\"datacenter1\": \"\"}]")
    end

    test "empty value" do
      assert :error = NetworkReplicationMap.cast("")
    end

    test "nil" do
      assert :error = NetworkReplicationMap.cast(nil)
      assert :error = NetworkReplicationMap.cast("nil")
    end
  end
```

apps/astarte_housekeeping/test/astarte_housekeeping/config/network_replication_map_test.exs
```diff
@@ -0,0 +1,61 @@
...
+#
+
+defmodule Astarte.Housekeeping.Config.NetworkReplicationMapTest do
+  use ExUnit.Case
+  alias Astarte.Housekeeping.Config.NetworkReplicationMap
+
+  describe "network replication map test" do
+    test "a map of non-empty strings and positive integers" do
+      assert {:ok, %{"datacenter1" => 1}} = NetworkReplicationMap.cast("{\"datacenter1\": 1}")
+    end
```
noaccOS: maybe a more complex example (at least two keys) would be better

```suggestion
  describe "cast/1 casts" do
    test "a map of non-empty strings and positive integers" do
      assert {:ok, %{"dc1" => 1, dc2 => 2}} = NetworkReplicationMap.cast(~s[{"dc1": 1, "dc2": 2}])
    end
  end
```

noaccOS (changes requested):

noaccOS (approved):

PR #1475, opened 2025-08-05
feat(hk): make a new vhost for each new realm created

eddbbt:


apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -84,6 +84,30 @@ defmodule Astarte.Housekeeping.Config do
           type: Astarte.Housekeeping.Config.TelemetryType,
           default: :expose
 
+  @envdoc "The host for the AMQP connection."
+  app_env :amqp_host, :astarte_housekeeping, :amqp_host,
+    os_env: "RABBITMQ_HOST",
```
noaccOS: I'm not 100% satisfied with the naming but I think `HOUSEKEEPING_AMQP_HOST` would be an improvement
`RABBITMQ_HOST` is currently being used but only for tests, and we don't really mention "Rabbit" specifically in our environment variables elsewhere

Feel free to propose alternatives

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
```diff
@@ -194,7 +194,8 @@ defmodule Astarte.Housekeeping.Realms.Queries do
   def create_realm(realm_name, public_key_pem, replication, device_limit, max_retention, opts) do
     with :ok <- validate_realm_name(realm_name),
          keyspace_name = Realm.keyspace_name(realm_name),
-         {:ok, replication_map_str} <- build_replication_map_str(replication) do
+         {:ok, replication_map_str} <- build_replication_map_str(replication),
+         {:ok, _} <- Astarte.Housekeeping.Vhosts.create_vhost(realm_name) do
```
noaccOS: Let's think about how this could go bad.

- If it creates the vhost but the relam creation fails, we leave a dangling vhost entry. This is not ideal but we also don't rollback creation steps in `do_create_realm` if a step goes wrong, so it would at least be consistent with what we already have
- If the vhost already exists, this currently does not create the realm. In name-collision scenarios, this is fine, but if the realm creation process crashed because of a database connection error, now it's no longer possible to create the realm through the astarte api, as the vhost would already be present

IMO, this could be the best approach, but if we successfully created the vhost, then in case of database/connection error we should try to delete the newly created vhost. This does not guarantee a clean state (the pod may be shut down completely just after vhost creation) but should resolve 99% of the errors, leaving only 1% situations to be handled manually

apps/astarte_housekeeping/lib/astarte_housekeeping/vhosts/vhosts.ex
```diff
@@ -0,0 +1,19 @@
...
+
+  def create_vhost(realm) do
+    rabbit_client()
+    |> ExRabbitMQAdmin.Vhost.put_vhost(realm)
+  end
+
+  defp rabbit_client() do
+    ExRabbitMQAdmin.client(
+      base_url:
+        "http://" <> Config.amqp_host!() <> ":" <> (Config.amqp_port!() |> Integer.to_string())
```
noaccOS: nit: this could be exposed by the `Config` module as `amqp_base_url!()`

noaccOS: should the protocol be customizable? to also allow for `https` connections

apps/astarte_housekeeping/lib/astarte_housekeeping/vhosts/vhosts.ex
```diff
@@ -0,0 +1,19 @@
...
+    )
+    |> ExRabbitMQAdmin.add_basic_auth_middleware(
+      username: Config.amqp_username!(),
+      password: Config.amqp_password!()
```
noaccOS: nit: in the library's examples, I see these options can be set in the `config` instead of manually calling them each time. we could set them in `config/runtime.exs`

we don't really do this in astarte but we do it quite a lot in [edgehog](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/config/runtime.exs#L220)

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -84,6 +84,30 @@ defmodule Astarte.Housekeeping.Config do
...
+    type: :binary,
+    default: "localhost"
```
noaccOS: I don't think we should assume localhost to be the default value, imo we should just have it mandatory
This of course is a breaking change

apps/astarte_housekeeping/lib/astarte_housekeeping/vhosts/vhosts.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Astarte.Housekeeping.Vhosts do
+  alias Astarte.Housekeeping.Config
+
+  def create_vhost(realm) do
+    rabbit_client()
+    |> ExRabbitMQAdmin.Vhost.put_vhost(realm)
```
noaccOS: I'm not sure we should have the vhost be named exactly like the realm
- have we checked there is no reserved vhost name but allowed realm name? this wolud fail dramatically
- we can have multiple realms named the same but in different astarte instances. while this feature was thought for multiple astarte instances using the same underlying database, I'd rather have it be part of the name

how about

```elixir
def vhost_name(realm_name) do
  astarte_instance = Astarte.DataAccess.Config.astarte_instance_id!()
  "astarte_triggers_#{astarte_instance}_#{realm_name}"
end
```

?

This of course _should_ be added to some shared dependency as it also needs to be used by dup, but we can think about where to put it later

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
eddbbt: `localhost` works for tests (mix test, not e2e) without further cursing. In ever other scenario, it shall break

apps/astarte_housekeeping/lib/astarte_housekeeping/vhosts/vhosts.ex
lusergit: [RabbitMQ docs](https://www.rabbitmq.com/docs/vhosts) do not mention naming restrictions (a part for `/` i guess) however I personall like this solution as it solves the same realm different astarte_instance problem 👍

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -84,6 +84,45 @@ defmodule Astarte.Housekeeping.Config do
...
+
+  @envdoc "The protocol for the AMQP connection."
+  app_env :amqp_protocol, :astarte_housekeeping, [Astarte.DataAccess.Repo, :amqp_protocol],
+    os_env: "HOUSEKEEPING_AMQP_PROTOCOL",
+    type: :binary,
+    env_overrides: [
+      dev: [default: "http"],
+      test: [default: "http"]
+    ]
+
```
noaccOS: after irl discussions, let's just assume it's http for now and we'll implement all ssl related options later. this is to avoid making breaking changes twice

```suggestion
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -84,6 +84,45 @@ defmodule Astarte.Housekeeping.Config do
           type: Astarte.Housekeeping.Config.TelemetryType,
           default: :expose
 
+  @envdoc "The host for the AMQP connection."
+  app_env :amqp_host, :astarte_housekeeping, [Astarte.DataAccess.Repo, :amqp_host],
```
noaccOS: I just noticed, why are these under DataAccess?

```suggestion
  app_env :amqp_host, :astarte_housekeeping, :amqp_host,
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -84,6 +84,45 @@ defmodule Astarte.Housekeeping.Config do
...
+    os_env: "HOUSEKEEPING_AMQP_HOST",
+    type: :binary,
+    env_overrides: [
+      dev: [default: "localhost"],
+      test: [default: "localhost"]
+    ]
+
+  @envdoc "The port for the AMQP connection."
+  app_env :amqp_port, :astarte_housekeeping, [Astarte.DataAccess.Repo, :amqp_port],
```
noaccOS: same as above

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -92,6 +131,10 @@ defmodule Astarte.Housekeeping.Config do
     disable_authentication!()
   end
 
+  def amqp_base_url!() do
+    "#{amqp_protocol!()}://#{amqp_host!()}:#{amqp_port!()}"
```
noaccOS:
```suggestion
    "http://#{amqp_host!()}:#{amqp_port!()}"
```

apps/astarte_housekeeping/lib/astarte_housekeeping/vhosts/vhosts.ex
```diff
@@ -0,0 +1,32 @@
...
+
+    rabbit_client()
+    |> ExRabbitMQAdmin.Vhost.put_vhost(realm)
+  end
+
+  def delete_vhost(realm) do
+    realm = vhost_name(realm)
+
+    rabbit_client()
+    |> ExRabbitMQAdmin.Vhost.delete_vhost(realm)
```
noaccOS: same as above

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realms.ex
```diff
@@ -73,7 +73,6 @@ defmodule Astarte.Housekeeping.Realms do
            Ecto.Changeset.apply_action(changeset, :insert) do
       case Core.create_realm(realm, opts) do
         :ok -> {:ok, realm}
-        {:ok, :started} -> {:ok, realm}
```
noaccOS: let's not delete this line

noaccOS (changes requested):

apps/astarte_housekeeping/lib/astarte_housekeeping/vhosts/vhosts.ex
eddbbt: above, where?

noaccOS: guess I didn't confirm a comment, love github reviews. 
here and in the function for creation, you're using the name `realm` but it does not represent a realm

```suggestion
    vhost_name = vhost_name(realm)

    rabbit_client()
    |> ExRabbitMQAdmin.Vhost.delete_vhost(vhost_name)
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -84,6 +84,36 @@ defmodule Astarte.Housekeeping.Config do
...
+    ]
+
+  @envdoc "The port for the AMQP connection."
+  app_env :amqp_port, :astarte_housekeeping, :amqp_port,
+    os_env: "HOUSEKEEPING_AMQP_PORT",
+    type: :integer,
+    env_overrides: [
+      dev: [default: 15672],
+      test: [default: 15672]
+    ]
```
noaccOS: this is actually the dafult amqp port so here it's fine to make a guess

```suggestion
  app_env :amqp_port, :astarte_housekeeping, :amqp_port,
    os_env: "HOUSEKEEPING_AMQP_PORT",
    type: :integer,
    default: 15672
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -84,6 +84,36 @@ defmodule Astarte.Housekeeping.Config do
...
           default: :expose
 
+  @envdoc "The host for the AMQP connection."
+  app_env :amqp_host, :astarte_housekeeping, :amqp_host,
+    os_env: "HOUSEKEEPING_AMQP_HOST",
+    type: :binary,
+    env_overrides: [
+      dev: [default: "localhost"],
+      test: [default: "localhost"]
+    ]
```
noaccOS: let's have this required so that it fails on startup if we didn't configure this correctly. also, I'm not sure whether we should also add a default for the `ci` env, how are unit tests passing in ci right now?

```suggestion
  app_env :amqp_host, :astarte_housekeeping, :amqp_host,
    os_env: "HOUSEKEEPING_AMQP_HOST",
    type: :binary,
    env_overrides: [
      prod: [required: true],
      dev: [default: "localhost"],
      test: [default: "localhost"]
    ]
```

apps/astarte_housekeeping/lib/astarte_housekeeping/vhosts/vhosts.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Astarte.Housekeeping.Vhosts do
+  alias Astarte.Housekeeping.Config
+
+  def create_vhost(realm) do
+    realm = vhost_name(realm)
+
+    rabbit_client()
+    |> ExRabbitMQAdmin.Vhost.put_vhost(realm)
```
noaccOS: I've already replied to the comment but here it's more visible

```suggestion
    vhost_name = vhost_name(realm)

    rabbit_client()
    |> ExRabbitMQAdmin.Vhost.put_vhost(vhost_name)
```

apps/astarte_housekeeping/test/astarte_housekeeping/realms/realms_test.exs
```diff
@@ -113,6 +115,12 @@ defmodule Astarte.Housekeeping.RealmsTest do
         # Test realm creation
         realm = realm_fixture(%{realm_name: name})
 
+        case(AMQP.get("/api/vhosts/#{Vhost.vhost_name(name)}")) do
+          {:ok, %HTTPoison.Response{status_code: 201}} -> assert true
+          {:ok, %HTTPoison.Response{status_code: 204}} -> assert true
+          _ -> assert false
+        end
+
```
noaccOS: nit: also, funny format

```suggestion
        assert {:ok, %HTTPoison.Response{status_code: status}} = AMQP.get("/api/vhosts/#{Vhost.vhost_name(name)}")
        assert status in [201, 204]
```

PR #1485, opened 2025-08-21
feat(e2e): amqp device trigger roundtrip

emirh1:
amqp device trigger roundtrip

Fixes #

* [ ] Yes
* [ ] No

```docs

```

### DIFF
--- a/tools/astarte_e2e/lib/astarte_e2e/amqp_device_trigger.ex
+++ b/tools/astarte_e2e/lib/astarte_e2e/amqp_device_trigger.ex
@@ -17,44 +17,44 @@
 #
 
 defmodule AstarteE2E.AmqpDeviceTrigger do
-  use GenServer
+  use GenServer, restart: :temporary
   require Logger
 
   alias AstarteE2E.Config
   alias AstarteE2E.Device
+  alias AstarteE2E.AmqpTriggers.Consumer
+  alias Astarte.Core.Triggers.SimpleEvents.SimpleEvent
+
+  @events [:device_connected]
+
+  @routing_key "e2e_device_triggers"
+  @exchange_suffix Config.amqp_trigger_exchange_suffix!()
 
   def name, do: "AmqpDeviceTrigger"
 
   def start_link(opts) do
-    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+    device_id = Astarte.Core.Device.random_device_id()
+    realm = Config.realm!()
+    init_arg = opts |> Keyword.merge(device_id: device_id, realm: realm)
+    GenServer.start_link(__MODULE__, init_arg)
   end
 
   @impl true
   def init(opts) do
-    with {:ok, realm} <- Config.realm() do
-      device_id = Astarte.Core.Device.random_device_id()
-      interfaces = Keyword.fetch!(opts, :interfaces)
-      device_opts = [realm: realm, device_id: device_id, interfaces: interfaces]
+    realm = Keyword.fetch!(opts, :realm)
+    device_id = Keyword.fetch!(opts, :device_id)
+    interfaces = Keyword.fetch!(opts, :interfaces)
 
-      case Device.start_link(device_opts) do
-        {:ok, device_pid} ->
-          state = %{
-            device_id: device_id,
-            realm: realm,
-            device_pid: device_pid
-          }
+    state = %{
+      device_id: device_id,
+      realm: realm,
+      interfaces: interfaces,
+      messages: nil,
+      device_pid: nil,
+      consumer_pid: nil
+    }
 
-          {:ok, state, {:continue, :install_triggers}}
-
-        {:error, reason} ->
-          Logger.error("Failed to start device: #{inspect(reason)}")
-          {:stop, reason}
-      end
-    else
-      {:error, reason} ->
-        Logger.error("Failed to fetch realm: #{inspect(reason)}")
-        {:stop, reason}
-    end
+    {:ok, state, {:continue, :install_triggers}}
   end
 
   @impl true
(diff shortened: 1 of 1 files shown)


tools/astarte_e2e/lib/astarte_e2e/amqp_device_trigger.ex
```diff
@@ -22,37 +22,42 @@ defmodule AstarteE2E.AmqpDeviceTrigger do
...
-            device_id: device_id,
-            realm: realm,
-            device_pid: device_pid
-          }
+    realm = Keyword.fetch!(opts, :realm)
+    device_id = Keyword.fetch!(opts, :device_id)
+    interfaces = Keyword.fetch!(opts, :interfaces)
+    device_opts = [realm: realm, device_id: device_id, interfaces: interfaces]
+
+    case Device.start_link(device_opts) do
```
noaccOS: I'm not sure if this is the culprit but in the http device trigger we first install the trigger and only after start the device (so that it generates a connection event)

you can take a look at `device_trigger.ex` to see how we handle that there

tools/astarte_e2e/lib/astarte_e2e/amqp_device_trigger.ex
```diff
@@ -22,56 +22,107 @@ defmodule AstarteE2E.AmqpDeviceTrigger do
...
       {:error, reason} ->
-        Logger.error("Failed to install amqp data triggers: #{inspect(reason)}")
-        {:stop, {:failed_to_install_triggers, reason}, state}
+        Logger.error("Device Trigger: consumer failed to start: #{inspect(reason)}")
+        {:stop, {:failed_to_start_consumer, reason}, state}
+    end
+  end
+
+  @impl true
+  def handle_continue(:publish_events, state) do
```
noaccOS, line 101: can you try starting the device here? there is a "no default consumer" error, so I think it tries to read the amqp message but the consumer is not started yet, making sure it's started before starting the device makes sense to me

tools/astarte_e2e/lib/astarte_e2e/amqp_device_trigger.ex
```diff
@@ -17,61 +17,135 @@
...
+
+  @impl true
+  def handle_call({:handle_message, payload, _meta}, _from, state) do
+    case SimpleEvent.decode(payload) do
+      %SimpleEvent{event: {:device_connected_event, _ev}} ->
+        Logger.info("AMQP Device Trigger: device connected",
+          device_id: state.device_id
+        )
+
+        {:stop, :normal, :ok, state}
```
noaccOS: this should not stop but instead reply with ok and continue with stop, so that we ack the message and close the channel before stopping (which has shown to create problems)

tools/astarte_e2e/lib/astarte_e2e/amqp_device_trigger.ex
```diff
@@ -17,61 +17,135 @@
...
+      %SimpleEvent{event: {:device_disconnected_event, _ev}} ->
+        Logger.info("AMQP Device Trigger: device disconnected",
+          device_id: state.device_id
+        )
+
+        {:stop, :normal, :ok, state}
+
+      %SimpleEvent{event: {:incoming_data_event, _ev}} ->
+        Logger.debug("AMQP Device Trigger: incoming_data_event", device_id: state.device_id)
+        {:stop, :normal, :ok, state}
```
noaccOS: I don't think these are valid: we only have a trigger for device_connected


```suggestion
```

tools/astarte_e2e/lib/astarte_e2e/amqp_device_trigger.ex
```diff
@@ -17,61 +17,135 @@
...
+  @events [:device_connected, :device_disconnected]
 
   def name, do: "AmqpDeviceTrigger"
 
   def start_link(opts) do
-    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+    device_id = Astarte.Core.Device.random_device_id()
+    realm = Config.realm!()
+    init_arg = opts |> Keyword.merge(device_id: device_id, realm: realm)
+    GenServer.start_link(__MODULE__, init_arg, name: via_tuple(realm, device_id))
```
noaccOS: why does this need a registered name?

tools/astarte_e2e/lib/astarte_e2e/amqp_device_trigger.ex
```diff
@@ -17,61 +17,124 @@
...
-          {:stop, reason}
-      end
+  @impl true
+  def handle_continue(:install_triggers, state) do
+    %{
+      device_id: device_id,
+      realm: realm
+    } = state
+
+    with :ok <- install_triggers(realm, device_id) do
```
noaccOS: nit: this should be a `case`

tools/astarte_e2e/lib/astarte_e2e/amqp_device_trigger.ex
```diff
@@ -113,13 +175,15 @@ defmodule AstarteE2E.AmqpDeviceTrigger do
 
   defp generate_triggers(realm, device_id) do
     encoded_device_id = Astarte.Core.Device.encode_device_id(device_id)
+    exchange = "astarte_events_#{realm}_#{Config.amqp_trigger_exchange_suffix!()}"
+    routing_key = "e2e_device_triggers"
```
noaccOS: can this be pinned as a module attribute? (ie, `@routing_key "e2e_device_trigger"`)
this is because it should be reused with the same name in multiple places, so it would be better to have a single source of truth, either as a module attribute or passed down from the init function

tools/astarte_e2e/lib/astarte_e2e/amqp_triggers/consumer.ex
```diff
@@ -78,7 +78,7 @@ defmodule AstarteE2E.AmqpTriggers.Consumer do
   def handle_info({:basic_deliver, payload, meta}, state) do
     case state.message_handler.(payload, meta) do
       :ok -> Basic.ack(state.channel, meta.delivery_tag)
-      :error  -> Basic.nack(state.channel, meta.delivery_tag)
+      :error -> Basic.nack(state.channel, meta.delivery_tag)
```
noaccOS: this should probably be removed as it was merged in the other pr

PR #1487, opened 2025-08-21
feat(dup): add per-vhost pool management and dynamic connection handling

eddbbt:
## Summary

This PR implements the behavior defined in the RFC one realm - one vhost. 
The DUP now builds one RabbitMQ pool connection per vhost (realm) and publishes using the corresponding exchange.

## Changes

- Introduced a new `AMQPTriggersSupervisor` responsible for:
  - Initializing a dedicated `ExRabbitPool.PoolSupervisor` per realm.
  - Managing `AMQPTriggersProducer` processes bound to the correct vhost.
- Changed `AMQPTriggersProducer` behavior:
  - Creates a connection only when an AMQP trigger needs to publish.
  - Reuses existing connections when available.
  - Closes idle connections after a `timeout`.
- A `DynamicSupervisor` is used to spawn and monitor processes dynamically per realm.
- Ensured supervisors use `:one_for_one` strategy and gracefully shut down their pool when producers exit with `{:normal, :timeout}`.

## Motivation

Previously, all realms shared a single pool of RabbitMQ connections. 

The new design:
- Improves isolation (each realm manages its own vhost connection).
- Reduces resource usage by dropping unused connections.
- Ensures scalability and clearer supervision strategies.

## Breaking Changes

- Connection management logic has changed: instead of a shared pool, each realm now maintains its own dedicated pool and producer. This may introduce incompatibility with older installations (realms needs a migration procedure)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -32,19 +32,26 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
-    GenServer.start_link(__MODULE__, args, name: __MODULE__)
+    realm_name = Keyword.fetch!(args, :realm)
+
+    producer = producer_from_registry(realm_name)
+
+    Logger.info("AMQPTriggers producer start link for realm #{realm_name}",
+      tag: "amqp_triggers_consumer_start_link"
+    )
+
+    GenServer.start_link(__MODULE__, args, name: producer)
```
noaccOS: name conflicts are expected, we can handle them at this level so that in the rest of the code we can just match with `{:ok, pid}`

```suggestion
    with {:error, {:already_started, pid}} <- GenServer.start_link(__MODULE__, args, name: producer) do
      {:ok, pid}
    end
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_supervisor.ex
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  def start_and_initialize_children(realm_name) do
+    {:ok, sup} =
+      case start_link(%{realm_name: realm_name}) do
+        {:ok, pid} ->
+          {:ok, pid}
+
+        {:error, {:already_registered, pid}} ->
+          {:ok, pid}
```
noaccOS: is this a thing?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_supervisor.ex
```diff
@@ -0,0 +1,109 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Astarte.DataUpdaterPlant.AMQPTriggersSupervisor do
```
noaccOS: I'm pretty sure this should just be a supervisor as it always has 2 children. the api also would look much cleaner

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_supervisor.ex
```diff
@@ -0,0 +1,109 @@
...
+    )
+
+    trigger_producer_conf =
+      Config.amqp_triggers_producer_options!()
+      |> Keyword.put(:virtual_host, vhost_name(realm_name))
+
+    # String.to_atom is unfortuantely required due to library limitation
+    trigger_pool_conf =
+      Config.triggers_producer_pool_config!()
+      |> Keyword.put(:name, {:local, String.to_atom(vhost_name(realm_name))})
```
noaccOS: I'm not sold on this. I think we should start with a single amqp connection and add more if we need them. probably one connection is enough as it will be just for its realm. this will also circumvent the issue with `String.to_atom`

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/application.ex
```diff
@@ -50,7 +50,8 @@ defmodule Astarte.DataUpdaterPlant.Application do
        [Config.cluster_topologies!(), [name: Astarte.DataUpdaterPlant.ClusterSupervisor]]},
       Astarte.DataUpdaterPlantWeb.Telemetry,
       {Astarte.DataAccess, data_access_opts},
-      Astarte.DataUpdaterPlant.DataPipelineSupervisor
+      Astarte.DataUpdaterPlant.DataPipelineSupervisor,
+      {Registry, keys: :unique, name: Astarte.DataUpdaterPlant.VhostRegistry}
```
noaccOS: this registry should be placed
- above its user
- in a `rest_for_one` supervisor
- possibly, only one level above its user in the supervision tree

the reason being:
above its user and in a rest for one supervisor, means that if the registry crashes, it makes sense for the amqp triggers producer to also crash.
placing it only one level above its user is for isolation of concerns

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/producers_supervisor.ex
```diff
@@ -46,15 +45,15 @@ defmodule Astarte.DataUpdaterPlant.ProducersSupervisor do
         id: :events_producer_pool
       )
 
-    triggers_pool =
+    generic_pool =
```
noaccOS: what are we using this for?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
```diff
@@ -50,13 +50,15 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandler do
 
   alias Astarte.Core.Triggers.SimpleTriggersProtobuf.AMQPTriggerTarget
 
-  def register_target(%AMQPTriggerTarget{exchange: nil} = _target) do
+  def register_target(%AMQPTriggerTarget{exchange: nil} = _target, _realm_name) do
```
noaccOS, line 53: Are we sure we don't need to declare the vhost in this instance?

docker-compose.yml
```diff
@@ -196,6 +196,9 @@ services:
     restart: on-failure
     volumes:
       - rabbitmq-data:/var/lib/rabbitmq
+    ports: 
+      -  15672:15672
+      -  5672:5672
```
noaccOS: I believe this is leftover from tests?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -70,65 +79,104 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
-      {:ok, channel}
+  defp init_producer(realm_name) do
+    vhost = AMQPTriggersSupervisor.vhost_name(realm_name)
+    # String.to_atom is unfortuantely required due to library limitation
+    conn = ExRabbitPool.get_connection_worker(String.to_atom(vhost))
+
+    case checkout_channel(conn) do
+      {:ok, %Channel{pid: channel_pid} = channel} ->
+        Process.monitor(channel_pid)
+        {:ok, {channel, realm_name}}
```
noaccOS: nit: you don't need to return the realm_name, the caller already has access to it as it's this function's parameter

eddbbt: its for further use inside the state, i don't think it will hurt

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_vhost_supervisor.ex
```diff
@@ -0,0 +1,48 @@
...
+      # {:error, {:shutdown, {:failed_to_start_child, _child_id, {:already_started, pid}}}} ->
+      #   {:ok, pid}
+
+      {:error, {:already_started, pid}} ->
+        {:ok, pid}
+
+      {:error, reason} ->
+        Logger.error("Failed to start child #{inspect(child)}: #{inspect(reason)}")
+    end
+  end
```
noaccOS: this supervisor should only be responsible for starting a child and returning it

```suggestion
  def start_for_realm(realm_name) do
    name = {:via, Registry, {@registry, {:realm_producer, realm_name}}}
    child = {Astarte.DataUpdaterPlant.AMQPTriggersSupervisor, [realm_name: realm_name, name: name]}

    case DynamicSupervisor.start_child(__MODULE__, child) do
      {:ok, pid} ->
        {:ok, pid}

      {:error, {:already_started, pid}} ->
        {:ok, pid}

      {:error, reason} ->
        Logger.error("Failed to start realm producerr #{inspect(child)}: #{inspect(reason)}")
        {:error, reason}
    end
  end
```

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
eddbbt:
```
   # Default exchange, no need to declare it
```

quite sure, but I'll check

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -70,81 +81,123 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
   end
 
   @impl GenServer
-  def handle_call({:publish, exchange, routing_key, payload, opts}, _from, chan) do
+  def handle_call(
+        {:publish, exchange, routing_key, payload, opts},
+        _from,
+        {conn, chan, realm_name}
+      ) do
+    Astarte.DataUpdaterPlant.VHostSupervisor.start_for_realm(realm_name)
```
noaccOS: the `handle_call` is already handled within the server process, your code would crash before trying to start the server here

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
```diff
@@ -557,22 +575,22 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandler do
...
+  defp wait_ok_publish(server, exchange, routing_key, payload, opts) do
+    publish(server, exchange, routing_key, payload, opts)
+    |> wait_backoff_and_publish(server, 1, nil, routing_key, payload, opts)
   end
 
-  defp publish(nil, routing_key, payload, opts) do
+  defp publish(_server, nil, routing_key, payload, opts) do
     Config.events_exchange_name!()
     |> AMQPEventsProducer.publish(routing_key, payload, opts)
   end
```
noaccOS: I think `nil` here only means we use the default exchange, not that we're discarding the value

```suggestion
  defp publish(server, nil, routing_key, payload, opts) do
    exchange = Config.events_exchange_name!()
    AMQPEventsProducer.publish(server, exchange, routing_key, payload, opts)
  end
```

I believe it would be better to just join the two `publish` bodies with

```elixir
  defp publish(server, exchange, routing_key, payload, opts) do
    exchange = exchange || Config.events_exchange_name!()
    AMQPEventsProducer.publish(server, exchange, routing_key, payload, opts)
  end
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
```diff
@@ -627,8 +645,16 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandler do
     opts = Enum.filter(opts_with_nil, fn {_k, v} -> v != nil end)
 
     payload = SimpleEvent.encode(simple_event)
+    {_, server} = Astarte.DataUpdaterPlant.VHostSupervisor.start_for_realm(simple_event.realm)
```
noaccOS: Let's match for `:ok` here:
- it's fine to crash if our state is nuked (it's weird that the server can't start, we can just crash)
- if the result was an `:error`, your logic here will try to `GenServer.call` on an error value (which is not a process) and lead to hard to debug errors

 ```suggestion
     {:ok, server} = Astarte.DataUpdaterPlant.VHostSupervisor.start_for_realm(simple_event.realm)
```

File: apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/producers_supervisor.ex [hunk 14]
```diff
@@ -46,15 +45,12 @@ defmodule Astarte.DataUpdaterPlant.ProducersSupervisor do
...
-         connection_pools: [Config.triggers_producer_pool_config!()]},
-        id: :triggers_producer_pool
-      )
-
-    children = [events_pool, triggers_pool, AMQPEventsProducer, AMQPTriggersProducer]
+    children = [
+      events_pool,
+      AMQPEventsProducer,
+      Astarte.DataUpdaterPlant.VHostSupervisor,
+      {Registry, keys: :unique, name: Astarte.DataUpdaterPlant.VhostRegistry}
```
noaccOS: the registry should be placed above the supervisor, so that if the registry crashes the supervisor also crashes

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_vhost_supervisor.ex
```diff
@@ -0,0 +1,44 @@
...
+    Logger.info("AMQPTriggers dynamic supervisor init.", tag: "amqp_triggers_sup_init")
+    DynamicSupervisor.init(strategy: :one_for_one)
+  end
+
+  def start_link(init_arg) do
+    DynamicSupervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
+  end
+
+  def start_for_realm(realm_name) do
+    server_name = server_from_realm(realm_name) |> dbg()
```
noaccOS:
```suggestion
    server_name = server_from_realm(realm_name)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_vhost_supervisor.ex
```diff
@@ -0,0 +1,44 @@
...
+  def init(_init_arg) do
+    Logger.info("AMQPTriggers dynamic supervisor init.", tag: "amqp_triggers_sup_init")
+    DynamicSupervisor.init(strategy: :one_for_one)
+  end
+
+  def start_link(init_arg) do
+    DynamicSupervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
+  end
+
+  def start_for_realm(realm_name) do
```
noaccOS: this reuses existing process so `start` is misleading

```suggestion
  def for_realm(realm_name) do
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -28,23 +28,30 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
-    GenServer.start_link(__MODULE__, args, name: __MODULE__)
+    realm_name = Keyword.fetch!(args, :realm)
+
+    Logger.info("AMQPTriggers producer start link for realm #{realm_name}",
+      tag: "amqp_triggers_consumer_start_link"
+    )
+
+    with {:error, {:already_started, pid}} <-
+           GenServer.start_link(__MODULE__, args, name: server) do
+      {:ok, pid}
```
noaccOS: this is now handled at the supervisor level so it can be simplified here

```suggestion
    GenServer.start_link(__MODULE__, args, name: server)
```

noaccOS (changes requested):
A couple things to fix but it's looking better and better with each revision

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
eddbbt: that's eventsProducers, not triggersProducers, this time we shall discard it

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
noaccOS: yes, @noaccOS, the vhost is created by housekeeping

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -70,81 +76,114 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
 
     ## Parameters
     - `exchange`: The name of the exchange to declare.
-
+    - `realm_name`: The name of the realm interested to this exchange.
     This function uses a longer timeout to allow RabbitMQ to process requests even if loaded.
   """
-  def declare_exchange(exchange) do
+  def declare_exchange(server, exchange, realm_name) do
```
noaccOS: `publish` and `declare_exchange` both need to add all added parameters (ie `server`: the server handle)

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -70,81 +76,114 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
     # Use a longer timeout to allow RabbitMQ to process requests even if loaded
-    GenServer.call(__MODULE__, {:declare_exchange, exchange}, 60_000)
+    GenServer.call(
+      server,
+      {:declare_exchange, %{exchange: exchange, realm: realm_name}},
```
noaccOS: this is the only call with a map; you're also not using the realm name, is it needed?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/server.ex
```diff
@@ -223,6 +223,11 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Server do
     {:stop, :monitored_process_died, state}
   end
 
+  @impl GenServer
+  def handle_info({:EXIT, _pid, :normal}, state) do
+    {:noreply, state}
+  end
+
```
noaccOS: - who is sending you this?
- why are you ignoring this exit request?

docker-compose.yml
```diff
@@ -196,6 +196,8 @@ services:
     restart: on-failure
     volumes:
       - rabbitmq-data:/var/lib/rabbitmq
+    ports: 
+      -  5672:5672 # required for e2e to pass
```
noaccOS: it's not, e2e runs inside docker and does not need port forwarding

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/time_based_actions_test.exs
```diff
@@ -31,7 +31,7 @@ defmodule Astarte.DataUpdaterPlant.TimeBasedActionsTest do
   alias Astarte.DataUpdaterPlant.DataUpdater
   alias Astarte.DataUpdaterPlant.TimeBasedActions
   alias Astarte.Helpers.Database
-
+  @moduletag astarte_instance_id: ""
```
noaccOS: I'm not sure we want this, and even if we did we should pass it as a parameter to astarte.cases.data

apps/astarte_data_updater_plant/test/test_helper.exs
```diff
@@ -18,6 +18,7 @@
 
 Mimic.copy(Astarte.Core.Mapping.ValueType)
 Mimic.copy(Astarte.DataAccess.Config)
+Mimic.copy(Astarte.DataUpdaterPlant.RPC.VMQPlugin.ClientMock)
```
noaccOS: why do you need this? previously we were doing just fine mocking vmqplugin as a whole

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/triggers_handler_test.exs
```diff
@@ -67,6 +68,7 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandlerTest do
   @custom_policy_routing_key "#{@realm}_#{@custom_policy_name}"
 
   setup_all do
+    Astarte.Housekeeping.AMQP.Vhost.create_vhost(@realm)
```
noaccOS: nit: use the alias you added pls

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -69,82 +76,115 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
-
     This function uses a longer timeout to allow RabbitMQ to process requests even if loaded.
   """
-  def declare_exchange(exchange) do
+  def declare_exchange(server, exchange) do
     # Use a longer timeout to allow RabbitMQ to process requests even if loaded
-    GenServer.call(__MODULE__, {:declare_exchange, exchange}, 60_000)
+    GenServer.call(
+      server,
+      {:declare_exchange, %{exchange: exchange}},
```
noaccOS:
```suggestion
      {:declare_exchange, exchange},
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -69,82 +76,115 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
 
-    {:reply, reply, chan}
+    {:reply, reply, {conn, chan, realm_name}, 60_000}
   end
 
   @impl GenServer
-  def handle_call({:declare_exchange, exchange}, _from, chan) do
-    # TODO: we need to decide who is responsible of deleting the exchange once it is
-    # no longer needed
+  def handle_call({:declare_exchange, %{exchange: exchange}}, _from, {conn, chan, realm_name}) do
```
noaccOS:
```suggestion
  def handle_call({:declare_exchange, exchange}, _from, {conn, chan, realm_name}) do
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -69,82 +76,115 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
+    :ok = Connection.close(conn)
+
+    case init_producer(realm_name) do
+      {:ok, {new_conn, new_channel, realm_name}} ->
+        {:noreply, {new_conn, new_channel, realm_name}, 60_000}
 
       {:error, _reason} ->
         schedule_connect()
-        {:noreply, :not_connected}
+        {:noreply, :not_connected, 60_000}
```
noaccOS: - we don't actually need a timeout here, as we're sure to receive a message after this (sent by `schedule_connect`) and we don't want to timeout before we receive it (technically we may set the reconnection delay to a value higher than 60 seconds)
- we can't use `:not_connected` as a state, as we're losing the realm name information, which we will need after a successful reconnection.


```suggestion
        {:noreply, {:not_connected, realm_name}}
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -69,82 +76,115 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
     )
 
-    case init_producer() do
-      {:ok, channel} ->
-        {:noreply, channel}
+    :ok = Connection.close(conn)
+
+    case init_producer(realm_name) do
+      {:ok, {new_conn, new_channel, realm_name}} ->
+        {:noreply, {new_conn, new_channel, realm_name}, 60_000}
```
noaccOS: nit


```suggestion
      {:ok, new_state} ->
        {:noreply, new_state, 60_000}
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -69,82 +76,115 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
+      |> Keyword.put(:virtual_host, vhost)
+
+    with {:ok, connection} <- Connection.open(trigger_producer_conf) do
+      case Channel.open(connection) do
+        {:ok, %Channel{pid: channel_pid} = channel} ->
+          Process.monitor(channel_pid)
+          {:ok, {connection, channel, realm_name}}
+
+        {:error, reason} ->
+          Logger.warning("Connessione non disponibile per realm #{realm_name}, riprovo...",
```
noaccOS: italian??

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
```diff
@@ -69,82 +76,115 @@ defmodule Astarte.DataUpdaterPlant.AMQPTriggersProducer do
...
+            tag: "event_producer_channel_checkout_fail"
+          )
+
+          schedule_connect()
```
noaccOS: it's fine to handle this outside (it's currently handled at line 140) as we want to simply stop the server if it happens during the init process. it will be started again when we request it later




```suggestion
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
```diff
@@ -557,22 +575,22 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandler do
...
-  defp publish(nil, routing_key, payload, opts) do
+  defp publish(_server, nil, routing_key, payload, opts) do
     Config.events_exchange_name!()
     |> AMQPEventsProducer.publish(routing_key, payload, opts)
   end
 
-  defp publish(exchange, routing_key, payload, opts) do
-    AMQPTriggersProducer.publish(exchange, routing_key, payload, opts)
+  defp publish(server, exchange, routing_key, payload, opts) do
+    AMQPTriggersProducer.publish(server, exchange, routing_key, payload, opts)
```
noaccOS, line 594: as it was already stated, let's put the server initialization here so that we don't start the amqp connection for a realm which is only using http triggers


```suggestion
    {:ok, server} = ...
    AMQPTriggersProducer.publish(server, exchange, routing_key, payload, opts)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
```diff
@@ -627,8 +645,16 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandler do
     opts = Enum.filter(opts_with_nil, fn {_k, v} -> v != nil end)
 
     payload = SimpleEvent.encode(simple_event)
+    {:ok, server} = Astarte.DataUpdaterPlant.VHostSupervisor.for_realm(simple_event.realm)
 
-    result = wait_ok_publish(target_exchange, routing_key, payload, [{:headers, headers} | opts])
+    result =
+      wait_ok_publish(
+        server,
```
noaccOS:
```suggestion
    result =
      wait_ok_publish(
        simple_event.realm,
```

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/time_based_actions_test.exs
```diff
@@ -17,7 +17,7 @@
 #
 
 defmodule Astarte.DataUpdaterPlant.TimeBasedActionsTest do
-  use Astarte.Cases.Data, async: true
+  use Astarte.Cases.Data, async: true, astarte_instance_id: ""
```
noaccOS: this is technically not needed if you move the server initialization as suggested by the two comments above

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/amqp_triggers_producer.ex
eddbbt: ops

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/triggers_handler.ex
```diff
@@ -537,11 +539,27 @@ defmodule Astarte.DataUpdaterPlant.TriggersHandler do
     end
   end
 
-  defp wait_backoff_and_publish(:ok, _retry, _exchange, _routing_key, _payload, _opts) do
+  defp wait_backoff_and_publish(
+         :ok,
+         _server,
```
noaccOS:
```suggestion
         _realm,
```

PR #1491, opened 2025-08-28
chore: forward port release-1.2 into release-1.3

noaccOS:


noaccOS:
Apparently the certificate in pairing tests is expired
image

This will be fixed later but for now you can just trust the `@csr_valid_until 1_755_940_320_000` which happens to be `Saturday, August 23, 2025 9:12:00 AM`

PR #1494, opened 2025-09-02
fix(rm): faulty checks for existing database resources

### LINKED ISSUE #1493: Devices stuck in "in deletion" status
When deleting a device from the Astarte Dashboard, the device remains in the `In deletion` status indefinitely.

The HTTP DELETE request to Realm Management on the device correctly returned 204. 
However, the backend logged this error:

```
** (CaseClauseError) no case clause matching: {:ok, true}
(astarte_realm_management 1.2.1-alpha.0) lib/astarte_realm_management/device_removal/device_remover.ex:35
```

We're currently running a snapshot image of Astarte v1.2.

davidebriani:
There were some faulty checks for existing database resources, due to the used `Repo.some?` custom function which now returns a `{:ok, bool}` tuple instead of directly a boolean, as in previous iterations.

Indeed, the tuple was either checked in `if` statements where it always computed to a trueish value, of it was pattern matched to boolean values which resulted in an match error and a crash.

The latter one is also the case of a bug in the DeviceRemover task, which crashed just after launch and resulted in devices to be stuck in the "in deletion"
status.

Fixes #1493.

Fixes #

* [ ] Yes
* [ ] No

```docs

```


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+## [1.2.1] - Unreleased
+### Fixed
+- [astarte_realm_management] Faulty checks for existing database resources. This fix also addresses a bug where devices got stuck in the "in deletion" status: [#1493](https://github.com/astarte-platform/astarte/issues/1493).
```
noaccOS: nit:
1. the user does not care that the error was faulty checks, the user cares that device deletion works
2. the only solved issue is, in fact, device deletion

PR #1500, opened 2025-09-09
chore: remove unused references to `astarte_rpc`

noaccOS:
this removes compilation warnings. the configuration of astarte_rpc
was removed from services which don't use astarte_rpc anymore


eddbbt (approved):

lusergit (approved):
:+1:

PR #1501, opened 2025-09-09
fix(e2e): crash on amqp data triggers for unexpected messages

noaccOS:
enhanced the logging while solving a crash while trying to print a tuple
without inspect


eddbbt (approved):

lusergit (approved):
🥇

PR #1504, opened 2025-09-09
wip add timestamp to server owned call

eddbbt:


apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -544,14 +569,17 @@ defmodule Astarte.AppEngine.API.Device do
...
     case Keyword.fetch!(opts, :type) do
       :properties ->
         DataTransmitter.set_property(
           realm,
           device_id,
           interface,
           path,
-          value
+          value,
+          timestamp: timestamp
```
noaccOS, line 582: the previous behavior did _not_ set the timestamp, resulting in a map with only the `:v` key being set

after this change, we will always set the `:t` key in the value we send to the device

I believe the device only sets the timestamp when the value has explicit timestamp, my gut feeling says we should do the same BUT we can ask the embedded team for their opinion on the matter

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -369,9 +375,15 @@ defmodule Astarte.AppEngine.API.Device do
...
+        raw_value,
+        explicit_timestamp
       ) do
-    now = DateTime.utc_now()
+    # check for explicit timestamp
+    now =
+      case explicit_timestamp do
+        false -> DateTime.utc_now()
+        {:ok, timestamp} -> timestamp
+      end
```
noaccOS, line 386: the fact that this value can be either `false` (a boolean) or `{:ok, timestamp}` (a tuple) is quite cursed. I suggest either making it a classic result

```elixir
case explicit_timestamp do
  {:ok, timestamp} -> timestamp
  :error -> DateTime.utc_now()
end
```

or just having a nullable datetime

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -369,9 +375,15 @@ defmodule Astarte.AppEngine.API.Device do
         device_id,
         interface_descriptor,
         path,
-        raw_value
+        raw_value,
+        explicit_timestamp
       ) do
-    now = DateTime.utc_now()
+    # check for explicit timestamp
+    now =
```
noaccOS, line 382: maybe `value_timestamp` better reflects the intent, `now` worked when it was only "now"

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -482,21 +496,32 @@ defmodule Astarte.AppEngine.API.Device do
          {:ok, interface_descriptor} <- InterfaceDescriptor.from_db_result(interface_row),
          {:ownership, :server} <- {:ownership, interface_descriptor.ownership},
          path <- "/" <> no_prefix_path do
+
+          explicit_timestamp =
+            with true <-
+                   Queries.interface_has_explicit_timestamp?(realm_name, interface_row.interface_id),
```
noaccOS, line 502: in their respective functions (`update_{individual,object}_interface_values`), we're already querying for the mappings

I think we can avoid this extra query by just delaying setting the timestamp until after we've read the mappings from the db

PR #1510, opened 2025-09-12
refactor(appengine): Resolved Credo's complains on appEngine

AliouneGaye21:

Fixes #

* [ ] Yes
* [x] No

```docs

```


eddbbt (approved):

apps/astarte_appengine_api/lib/astarte_appengine_api_web/plug/guardian_authorize_path.ex
```diff
@@ -22,7 +22,7 @@ defmodule Astarte.AppEngine.APIWeb.Plug.GuardianAuthorizePath do
   defp authorize(conn, opts) do
     with %User{authorizations: authorizations} <- AuthGuardian.Plug.current_resource(conn),
          {:ok, auth_path} <- build_auth_path(conn),
-         :ok <- is_path_authorized?(conn.method, auth_path, authorizations) do
+         :ok <- path_authorized?(conn.method, auth_path, authorizations) do
```
noaccOS: as we're reworking naming anyway, it's a good time to make sure they're not only accepted by credo, but that they also make sense

in this instance, a function named with a question mark at the end is expected to return a boolean, while this function really returns a result tuple

I suggest changing its name to better reflect this behavior. Usually, we just add `check_` at the start


```suggestion
         :ok <- check_path_authorized(conn.method, auth_path, authorizations) do
```

apps/astarte_appengine_api/lib/astarte_appengine_api/device/aliases.ex
```diff
@@ -124,7 +124,7 @@ defmodule Astarte.AppEngine.API.Device.Aliases do
   defp apply_delete(%Changeset{valid?: false} = changeset, _delete_aliases),
     do: changeset
 
-  defp apply_delete(changeset, delete_aliases) when length(delete_aliases) == 0,
+  defp apply_delete(changeset, delete_aliases) when delete_aliases == [],
```
noaccOS: (this is related to all similar funcitons)

in general, guards add quite a bit of verbosity. sometimes it makes things clearer, but in this instance I think the shorter version is just as readable


```suggestion
  defp apply_delete(changeset, [] = _delete_aliases),
```

feel free to _not_ add the alias, I think it helps clear what everything is in more complex scenarios but it's not something we always (or even often) do

noaccOS (changes requested):
Just a couple nitpicks, but the changes in general look good

although, pr title and commit description don't follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) so they may need to be changed

PR #1511, opened 2025-09-15
Extract AMQP triggers producer into standalone library and update workflows

eddbbt:
This PR extracts the AMQP triggers producer into a new standalone library to improve modularity and reuse.
It also updates docker compose and GitHub workflows to support the /libs directory.
The old code in astarte_data_updater_plant has been removed and references have been updated to rely on the new library.


.github/workflows/pr-container-build-workflow.yaml
```diff
@@ -16,7 +16,6 @@ jobs:
       matrix:
         app:
         - astarte_appengine_api
-        - astarte_data_updater_plant
```
noaccOS, line 19: Why can't we just generalize the current job?

docker-compose.yml
```diff
@@ -99,11 +99,13 @@ services:
...
     image: astarte/astarte_data_updater_plant:1.3-snapshot
     build:
-      context: apps/astarte_data_updater_plant
+      context: .
+      dockerfile: apps/astarte_data_updater_plant/Dockerfile
     env_file:
       - ./.env
     environment:
       DATA_UPDATER_PLANT_AMQP_CONSUMER_HOST: "rabbitmq"
+      AMQP_TRIGGERS_AMQP_CONSUMER_HOST: "rabbitmq"
```
noaccOS: this does not exist in the first commit, I would move it to the same commit where it's being declared

apps/astarte_data_updater_plant/Dockerfile
```diff
@@ -1,12 +1,16 @@
...
 # install build dependencies
 # --allow-releaseinfo-change allows to pull from 'oldstable'
 RUN apt-get update --allow-releaseinfo-change -y && apt-get install -y build-essential git \
     && apt-get clean && rm -f /var/lib/apt/lists/*_*
 
 WORKDIR /app
 
+COPY apps/astarte_data_updater_plant/ ./
+COPY libs/ ../libs/
+
```
noaccOS, line 12: why do you need to copy stuff around? if we're using the root directory for the context, should we not already have access to all the files?

.github/workflows/pr-container-build-workflow.yaml
eddbbt: this require to check other services Dockerfile also, a little out of scope for now (but not for the future)

apps/astarte_data_updater_plant/Dockerfile
eddbbt: yes, but "all the files" include also other astarte services, meaning that if we add `.` the resulting image would include code that we do not care about

i'll try with ADD, but shall be the same

before, the workflow used `ADD . . ` that added the entire folder
without adding, the files are not mounted

.github/workflows/pr-container-build-workflow.yaml
noaccOS: makes sense, but I still don't like having to duplicate the job.

instead, how about using docker compose to handle the build? it contains all the definitions already

```
run: docker compose build ${{ matrix.app }}
```

it's not like we're re-using the `$CONTAINER_IMAGE_NAME:ci-test` anyway

.github/workflows/pr-container-build-workflow.yaml
```diff
@@ -15,16 +15,15 @@ jobs:
...
+        - astarte-realm-management
+        - astarte-trigger-engine
     steps:
     - uses: actions/checkout@v4
     - name: Building Docker Image
       env:
         CONTAINER_IMAGE_NAME: gh_actions_test
-      working-directory: ./apps/${{ matrix.app }}
-      run: docker build -t $CONTAINER_IMAGE_NAME:ci-test .
+      run: docker compose build ${{ matrix.app }}
```
noaccOS: missing newline at the end of file

apps/astarte_data_updater_plant/Dockerfile
```diff
@@ -1,12 +1,16 @@
 FROM hexpm/elixir:1.15.7-erlang-26.1-debian-bookworm-20230612-slim AS builder
 
+
```
noaccOS: a newline spawned here

libs/astarte_events/lib/astarte_events/amqp_triggers/producer.ex
```diff
@@ -0,0 +1,139 @@
...
+
+    case init_producer(realm_name) do
+      {:ok, state} -> {:ok, state, 60_000}
+      {:error, reason} -> {:stop, reason}
+    end
+  end
+
+  @impl true
+  def handle_call({:publish, exchange, routing_key, payload, opts}, _from, {conn, chan, realm}) do
+    dbg(@adapter)
```
noaccOS: leftover

libs/astarte_events/lib/astarte_events/amqp_triggers/producer.ex
```diff
@@ -0,0 +1,139 @@
...
+    reply = @adapter.publish(chan, exchange, routing_key, payload, opts)
+    {:reply, reply, {conn, chan, realm}, 60_000}
+  end
+
+  @impl true
+  def handle_call({:declare_exchange, exchange}, _from, {conn, chan, realm}) do
+    dbg(@adapter)
```
noaccOS: leftover

libs/astarte_events/lib/astarte_events/amqp_triggers/producer.ex
```diff
@@ -0,0 +1,139 @@
...
+    GenServer.call(server, {:publish, exchange, routing_key, payload, opts}, 60_000)
+  end
+
+  def declare_exchange(server, exchange) when is_binary(exchange) do
+    GenServer.call(server, {:declare_exchange, exchange}, 60_000)
+  end
+
+  @impl true
+  def init(opts) do
+    dbg(@adapter)
```
noaccOS: leftover

libs/astarte_events/lib/astarte_events/amqp_triggers/producer.ex
```diff
@@ -0,0 +1,139 @@
...
+#
+
+defmodule Astarte.Events.AMQPTriggers.Producer do
+  use GenServer, restart: :transient
+  require Logger
+
+  alias AMQP.{Channel, Connection}
+  alias Astarte.Events.Config
+
+  @adapter Config.amqp_adapter!()
```
noaccOS: m2c: let's just use `ExRabbitPool.RabbitMQ` here, as it only broke stuff when we used it this way in dup

we can use mimic to stub it with `ExRabbitPool.FakeRabbitMQ` in tests if we need it

```suggestion
```

libs/astarte_events/lib/astarte_events/amqp_triggers/producer.ex
```diff
@@ -0,0 +1,139 @@
...
+# limitations under the License.
+#
+
+defmodule Astarte.Events.AMQPTriggers.Producer do
+  use GenServer, restart: :transient
+  require Logger
+
+  alias AMQP.{Channel, Connection}
+  alias Astarte.Events.Config
+
```
noaccOS:
```suggestion
  alias ExRabbitPool.RabbitMQ
```

libs/astarte_events/lib/astarte_events/amqp_triggers/producer.ex
```diff
@@ -0,0 +1,139 @@
...
+    case init_producer(realm_name) do
+      {:ok, state} -> {:ok, state, 60_000}
+      {:error, reason} -> {:stop, reason}
+    end
+  end
+
+  @impl true
+  def handle_call({:publish, exchange, routing_key, payload, opts}, _from, {conn, chan, realm}) do
+    dbg(@adapter)
+    reply = @adapter.publish(chan, exchange, routing_key, payload, opts)
```
noaccOS:
```suggestion
    reply = RabbitMQ.publish(chan, exchange, routing_key, payload, opts)
```

libs/astarte_events/lib/astarte_events/amqp_triggers/producer.ex
```diff
@@ -0,0 +1,139 @@
...
+    {:reply, reply, {conn, chan, realm}, 60_000}
+  end
+
+  @impl true
+  def handle_call({:declare_exchange, exchange}, _from, {conn, chan, realm}) do
+    dbg(@adapter)
+    reply = @adapter.declare_exchange(chan, exchange, type: :direct, durable: true)
```
noaccOS:
```suggestion
    reply = RabbitMQ.declare_exchange(chan, exchange, type: :direct, durable: true)
```

libs/astarte_events/lib/astarte_events/config.ex
```diff
@@ -0,0 +1,178 @@
...
+    else
+      server_name = amqp_consumer_ssl_custom_sni!() || amqp_consumer_host!()
+      Keyword.put(ssl_options, :server_name_indication, to_charlist(server_name))
+    end
+  end
+
+  def amqp_adapter!() do
+    Application.get_env(:astarte_events, :amqp_adapter) || ExRabbitPool.RabbitMQ
+  end
+
```
noaccOS:
```suggestion
```

libs/astarte_events/lib/astarte_events/config.ex
```diff
@@ -0,0 +1,178 @@
...
+          {:username, String.t()}
+          | {:password, String.t()}
+          | {:virtual_host, String.t()}
+          | {:host, String.t()}
+          | {:port, integer()}
+          | {:ssl_options, ssl_options}
+          | {:channels, integer()}
+  @envdoc "The host for the AMQP consumer connection."
+  app_env(:amqp_consumer_host, :astarte_events, :amqp_consumer_host,
+    os_env: "ASTARTE_EVENTS_AMQP_CONSUMER_HOST",
```
noaccOS: this is actually an AMQP producer


```suggestion
    os_env: "ASTARTE_EVENTS_PRODUCER_AMQP_HOST",
```

same for all other variables

libs/astarte_events/mix.exs
```diff
@@ -0,0 +1,83 @@
...
+  defp astarte_required_modules("true") do
+    [
+      {:astarte_data_access, in_umbrella: true}
+    ]
+  end
+
+  defp astarte_required_modules(_) do
+    [
+      {:astarte_data_access,
+       github: "astarte-platform/astarte_data_access", branch: "release-1.3"}
```
noaccOS: let's add it to standard deps and drop this `astarte_required_modules` function
we will include it in the monorepo soon enough anyway

noaccOS (changes requested):
Left a few comments but nothing major, good job

noaccOS (approved):

AliouneGaye21 (approved):

PR #1522, opened 2025-09-19
Extract AMQP events producer into astarte_events

eddbbt:
like https://github.com/astarte-platform/astarte/pull/1511 but with amqp events


libs/astarte_events/lib/astarte_events/amqp_events/amqp_events_producer.ex
```diff
@@ -16,17 +16,15 @@
 # limitations under the License.
 #
 
-defmodule Astarte.DataUpdaterPlant.AMQPEventsProducer do
+defmodule Astarte.Events.AMQPEvents.AMQPEventsProducer do
   require Logger
-  use GenServer
+  use GenServer, restart: :transient
```
noaccOS: This is incorrect: for the realm-only connection it makes sense, as it is restarted automagically when needed
On the other hand the assumption here is for the amqp events producer to always be available, so it makes sense that it is restarted in case of error


```suggestion
  use GenServer
```

libs/astarte_events/lib/astarte_events/amqp_events/producers_supervisor.ex
```diff
@@ -18,16 +18,16 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-defmodule Astarte.DataUpdaterPlant.ProducersSupervisor do
+defmodule Astarte.Events.AMQPEvents.ProducersSupervisor do
```
noaccOS: IMO we can restore the status from before 96600300, ie all amqp resources under the same supervisor

what this would mean in practice is deleting this module, and moving its children to the application module

libs/astarte_events/lib/astarte_events/amqp_events/amqp_events_producer.ex
```diff
@@ -16,17 +16,15 @@
 # limitations under the License.
 #
 
-defmodule Astarte.DataUpdaterPlant.AMQPEventsProducer do
+defmodule Astarte.Events.AMQPEvents.AMQPEventsProducer do
```
noaccOS: I think we don't need additional layering: if we remove the producerssupervisor, this is the only module for amqp events (it also was already in the name)


```suggestion
defmodule Astarte.Events.AMQPEventsProducer do
```

libs/astarte_events/lib/astarte_events/config.ex
```diff
@@ -169,6 +175,18 @@ defmodule Astarte.Events.Config do
     end
   end
 
+  # Since we have only one producer, this is not configurable
+  def events_producer_connection_number!(), do: 1
+
+  def events_producer_pool_config!() do
```
noaccOS: nit: we only produce


```suggestion
  def events_connection_number!(), do: 1

  def events_pool_config!() do
```

PR #1523, opened 2025-09-23
refactor: extract event dispatching logic to astarte_events

eddbbt:
depends on https://github.com/astarte-platform/astarte/pull/1522


libs/astarte_events/lib/astarte_events/triggers_handler.ex
```diff
@@ -0,0 +1,242 @@
...
+       ) do
+    %SimpleEvent{
+      simple_trigger_id: simple_trigger_id,
+      parent_trigger_id: parent_trigger_id,
+      realm: realm,
+      device_id: device_id,
+      timestamp: timestamp,
+      event: {event_type, event}
+    }
+  end
```
noaccOS, line 46: `do_make_simple_event` has become useless, it can be inlined


```suggestion
  def dispatch_event(event, event_type, target, realm, device_id, timestamp, policy) do
    %SimpleEvent{
      simple_trigger_id: target.simple_trigger_id,
      parent_trigger_id: target.parent_trigger_id,
      realm: realm,
      device_id: device_id,
      timestamp: timestamp,
      event: {event_type, event}
    }
    |> do_dispatch_event(target, policy)
  end
```

libs/astarte_events/lib/astarte_events/triggers_handler.ex
```diff
@@ -0,0 +1,242 @@
...
+        target_exchange,
+        routing_key,
+        payload,
+        [{:headers, headers} | opts]
+      )
+
+    Logger.debug("headers: #{inspect(headers)}, routing key: #{inspect(routing_key)}")
+
+    :telemetry.execute(
+      [:astarte, :data_updater_plant, :triggers_handler, :published_event],
```
noaccOS: well this is a problem
I'm not sure what the long term best solution is (having the application be configurable?), for now I'd say simple is best, let's avoid emitting telemetry here and add a wrapper to the function in dup

```elixir
defp dispatch_event_with_telemetry(event, event_type, target, realm, device_id, timestamp, policy) do
  result = TriggersHandler.dispatch_event(event, event_type, target, realm, device_id, timestamp, policy)
  :telemetry.execute(...)
  result
end

File: libs/astarte_events/lib/astarte_events/triggers_handler.ex [hunk 3]
```diff
@@ -0,0 +1,242 @@
...
+           message_persistent: message_persistent
+         },
+         policy_name
+       ) do
+    {event_type, _event_struct} = simple_event.event
+
+    simple_trigger_id_str =
+      simple_event.simple_trigger_id
+      |> :uuid.uuid_to_string()
+      |> to_string()
```
noaccOS: nit: let's avoid the erlang module and use the native elixir one


```suggestion
      |> UUID.binary_to_string()
```

you may need to change the dependency in mix.exs to elixir_uuid

noaccOS (approved):

PR #1524, opened 2025-09-25
ci: don't skip dialyzer plt build with cache hit

noaccOS:
dialyzer is smart enough to avoid rebuilding cache for already built modules

fixes issue with dialyzer step failing for local dependencies because they didn't get rebuilt


eddbbt (approved):

AliouneGaye21 (approved):

lusergit (approved):
:merge_demon:

PR #1526, opened 2025-09-26
refactor: increase astarte_events scope

noaccOS:
add more functionality needed to use triggers from other services

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/trigger.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/trigger.ex
@@ -27,132 +27,42 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.Trigger do
   alias Astarte.Core.Device
   alias Astarte.DataUpdaterPlant.DataUpdater.Core
   alias Astarte.DataUpdaterPlant.DataUpdater.State
-  alias Astarte.Core.Triggers.SimpleTriggersProtobuf.AMQPTriggerTarget
   alias Astarte.DataUpdaterPlant.TriggersHandler
   alias Astarte.DataUpdaterPlant.DataUpdater.EventTypeUtils
   alias Astarte.DataUpdaterPlant.DataUpdater.Queries
   alias Astarte.Core.Triggers.SimpleTriggersProtobuf
   alias Astarte.Core.Triggers.DataTrigger
-  alias Astarte.Core.InterfaceDescriptor
-  alias Astarte.Core.Mapping.EndpointsAutomaton
   alias Astarte.DataAccess.Interface, as: InterfaceQueries
-  alias Astarte.DataUpdaterPlant.TriggerPolicy.Queries, as: PolicyQueries
-  alias Astarte.Core.Triggers.SimpleTriggersProtobuf.DeviceTrigger, as: ProtobufDeviceTrigger
+  alias Astarte.DataAccess.Realms.SimpleTrigger
   alias Astarte.Core.Triggers.SimpleTriggersProtobuf.DataTrigger, as: ProtobufDataTrigger
   alias Astarte.Core.Triggers.SimpleTriggersProtobuf.Utils, as: SimpleTriggersProtobufUtils
   alias Astarte.DataUpdaterPlant.MessageTracker
+  alias Astarte.Events.Triggers
 
   def populate_triggers_for_object!(state, object_id, object_type) do
     %{realm: realm} = state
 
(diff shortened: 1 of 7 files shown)


libs/astarte_events/lib/astarte_events/triggers_handler/core.ex
```diff
@@ -16,24 +16,23 @@
...
-  """
-  import Bitwise, only: [<<<: 2]
-  require Logger
-  @max_backoff_exponent 8
-  @max_rand trunc(:math.pow(2, 32) - 1)
-
+defmodule Astarte.Events.TriggersHandler.Core do
   alias Astarte.Core.Triggers.SimpleEvents.SimpleEvent
   alias Astarte.Core.Triggers.SimpleTriggersProtobuf.AMQPTriggerTarget
+  alias Astarte.Core.Triggers.SimpleTriggersProtobuf.AMQPTriggerTarget
```
eddbbt: duplicated

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/trigger.ex
```diff
@@ -49,112 +47,22 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.Trigger do
...
-        load_trigger(state_acc, trigger_data, trigger_target)
-      end)
+    new_state = Map.merge(state, new_data)
 
     Enum.reduce(new_state.volatile_triggers, new_state, fn {{obj_id, obj_type},
                                                             {simple_trigger, trigger_target}},
                                                            state_acc ->
       if obj_id == object_id and obj_type == object_type_int do
-        load_trigger(state_acc, simple_trigger, trigger_target)
+        EventsTriggersHandler.Core.load_trigger(realm, simple_trigger, trigger_target, state_acc)
```
Annopaolo: Here it is not clear why we're using a function that looks almost the same from `EventsTriggersHandler` and `EventsTriggersHandler.Core`, but actually is quite different. My 2 cents:
- `s/load_trigger/cache_trigger/g`
- A `Core` module is usually exposed to its "API" module, so
```elixir
# In EventsTriggersHandler
defdelegate cache_simple_trigger_and_target, to: Core

# In EventsTriggersHandler.Core
def cache_simple_trigger_and_target(...) do
```

This piece of code would then become:
```suggestion
    new_data =
      EventsTriggersHandler.cache_triggers(realm, simple_triggers, Map.from_struct(state))
    new_state = Map.merge(state, new_data)
    
        Enum.reduce(new_state.volatile_triggers, new_state, fn {{obj_id, obj_type},
                                                            {simple_trigger, trigger_target}},
                                                           state_acc ->
      if obj_id == object_id and obj_type == object_type_int do
              EventsTriggersHandler.cache_simple_trigger_and_target(realm, simple_trigger, trigger_target, state_acc)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/trigger.ex
```diff
@@ -366,11 +250,13 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.Trigger do
 
       case trigger do
         {:data_trigger, %ProtobufDataTrigger{match_path: "/*"}} ->
-          {:ok, load_trigger(new_state, trigger, target)}
+          {:ok,
+           EventsTriggersHandler.Core.load_trigger(new_state.realm, trigger, target, new_state)}
```
Annopaolo: See above comment, here it would be 

```suggestion
           EventsTriggersHandler.cache_simple_trigger_and_target(new_state.realm, trigger, target, new_state)}
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/trigger.ex
noaccOS: It's actually 99% a copy/paste of that function, but I'll look into this

noaccOS: For one thing I dislike the `load` -> `cache` suggestion because it implies we're saving it _somewhere_, but we're actually just returning them
If eventually (as is my hope) the whole caching mechanism will be refactored into astarte_events (using concache or something similar), then the proposed change would make sense

Annopaolo: For the same reason I'm not fond of `load` either (we're not loading data from one place to the other, we're just putting them in the process state). No big deal though, we can look for another verb. `store` maybe?
Or let's just wait to see where we'll be going in the future and change accordingly

PR #1527, opened 2025-09-29
feat(pairing)!: device registration trigger

eddbbt:
emit an event when a device is correctly registered

BREAKING CHANGE: pairing now requires
ASTARTE_EVENTS_PRODUCER_AMQP_HOST to be set in order to successfully
generate events

depends on https://github.com/astarte-platform/astarte/pull/1533

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - `HOUSEKEEPING_ASTARTE_KEYSPACE_NETWORK_REPLICATION_MAP` - Datacenter replication map when using NetworkTopologyStrategy (no default, required when using network strategy)
 - Added database events handling configuration across all services:
   - `DATABASE_EVENTS_HANDLING_METHOD` - Controls how database events are handled: "expose" (via telemetry) or "log" (to logs) (default: "expose")
+- [astarte_pairing] Added device registration triggers
 
 ### Changed
 - BREAKING: Merged API services into main services, eliminating separate containers:
@@ -37,6 +38,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - `HOUSEKEEPING_AMQP_USERNAME` - AMQP username (default: guest)
   - `HOUSEKEEPING_AMQP_PASSWORD` - AMQP password (default: guest)
   - `HOUSEKEEPING_AMQP_MANAGEMENT_PORT` - AMQP management API port (default: 15672)
+- BREAKING: [astarte_pairing] AMQP Producer configuration is now mandatory using the `ASTARTE_EVENTS_PRODUCER_AMQP_*` environment variables
 
 ## [1.2.1] - Unreleased
 ### Fixed
--- a/apps/astarte_pairing/lib/astarte_pairing/agent/agent.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/agent/agent.ex
@@ -25,6 +25,12 @@ defmodule Astarte.Pairing.Agent do
   alias Astarte.Pairing.Agent.DeviceRegistrationRequest
   alias Astarte.Pairing.Agent.DeviceRegistrationResponse
   alias Astarte.Pairing.Engine
+  alias Astarte.Pairing.Config
+
+  alias Astarte.Core.Triggers.SimpleEvents.DeviceDeletionStartedEvent
+  alias Astarte.Events.Triggers
+  alias Astarte.Events.TriggersHandler
+  @cache_name Config.trigger_cache_name!()
 
   def register_device(realm, attrs \\ %{}) do
     changeset =
@@ -44,7 +50,10 @@ defmodule Astarte.Pairing.Agent do
              }
            end),
          {:ok, credentials_secret} <-
-           Engine.register_device(realm, hw_id, initial_introspection: initial_introspection) do
+           Engine.register_device(realm, hw_id, initial_introspection: initial_introspection),
+         {:ok, decoded_id} <- Device.decode_device_id(hw_id, allow_extended_id: true) do
+      dispatch_device_registration_trigger(realm, decoded_id)
+
       {:ok, %DeviceRegistrationResponse{credentials_secret: credentials_secret}}
     end
   end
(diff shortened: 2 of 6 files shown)


apps/astarte_pairing/lib/astarte_pairing/agent/agent.ex
```diff
@@ -44,7 +50,10 @@ defmodule Astarte.Pairing.Agent do
              }
            end),
          {:ok, credentials_secret} <-
-           Engine.register_device(realm, hw_id, initial_introspection: initial_introspection) do
+           Engine.register_device(realm, hw_id, initial_introspection: initial_introspection),
+         {:ok, decoded_id} <- Device.decode_device_id(hw_id, allow_extended_id: true) do
+      device_registration_trigger(realm, decoded_id)
```
Annopaolo: Nit: it is not clear what's the action in the call. What about this?
```suggestion
      dispatch_device_registration_trigger(realm, decoded_id)
```

apps/astarte_pairing/lib/astarte_pairing/agent/agent.ex
```diff
@@ -54,4 +63,43 @@ defmodule Astarte.Pairing.Agent do
...
+        device_id,
+        timestamp,
+        policy
+      )
+    end)
+    |> Enum.all?(&(&1 == :ok))
+    |> case do
+      true -> :ok
+      false -> :error
+    end
```
Annopaolo: It seems we're not using the return value of this function, so why checking it?

apps/astarte_pairing/lib/astarte_pairing/config.ex
```diff
@@ -79,6 +79,10 @@ defmodule Astarte.Pairing.Config do
           type: Astarte.Pairing.Config.TelemetryType,
           default: :expose
 
+  app_env :trigger_cache, :astarte_pairing, :trigger_cache,
+    type: :atom,
+    default: :trigger_cache
```
Annopaolo: If this is something we want to keep configurable, let's add some documentation for it as the rest of config options

apps/astarte_pairing/lib/astarte_pairing/application.ex
```diff
@@ -26,6 +26,8 @@ defmodule Astarte.Pairing.Application do
   require Logger
 
   @app_version Mix.Project.config()[:version]
+  @trigger_lifetime_ttl :timer.seconds(60)
+  @trigger_cache_id Config.trigger_cache!()
```
Annopaolo: Nit: let's specify whether it is a name, an id or something else.
```suggestion
  @trigger_cache_name Config.trigger_cache_name!()
```

apps/astarte_pairing/lib/astarte_pairing/application.ex
```diff
@@ -46,7 +48,13 @@ defmodule Astarte.Pairing.Application do
     children = [
       Astarte.PairingWeb.Telemetry,
       {Astarte.Pairing.CredentialsSecret.Cache, []},
-      Astarte.PairingWeb.Endpoint
+      Astarte.PairingWeb.Endpoint,
+      {ConCache,
+       [
+         name: @trigger_cache_id,
```
Annopaolo: See comment aboce

apps/astarte_pairing/mix.exs
```diff
@@ -96,9 +96,10 @@ defmodule Astarte.Pairing.Mixfile do
       {:dialyxir, "~> 1.0", only: [:dev, :ci], runtime: false},
       {:skogsra, "~> 2.2"},
       {:cors_plug, "~> 2.0"},
-      {:telemetry_metrics, "~> 0.4"},
-      {:telemetry_poller, "~> 0.4"},
-      {:telemetry_metrics_prometheus_core, "~> 0.4"},
+      {:telemetry, "~> 1.0"},
+      {:telemetry_metrics, "~> 0.6.1", override: true},
+      {:telemetry_poller, "~> 1.0", override: true},
+      {:telemetry_metrics_prometheus_core, "~> 1.0.2", override: true},
```
Annopaolo: I suspect this was changed in order to build due to the fact that Realm Management (which has updated versions of telemetry deps) is now a dependency. If so, let's not mix two different actions in one PR and open a small one just to make Pairing compile once again

eddbbt: here you go https://github.com/astarte-platform/astarte/pull/1540

apps/astarte_pairing/lib/astarte_pairing/agent/agent.ex
```diff
@@ -54,4 +63,38 @@ defmodule Astarte.Pairing.Agent do
...
+        event,
+        event_type,
+        target,
+        realm_name,
+        device_id,
+        timestamp,
+        policy
+      )
+    end)
+  end
```
Annopaolo, line 88: I personally would do the following, but it's not an issue
```suggestion
  Enum.map(targets, fn {target, policy} ->
      TriggersHandler.dispatch_event(
        event,
        event_type,
        target,
        realm_name,
        device_id,
        timestamp,
        policy
      )
    end)
  end
```

apps/astarte_pairing/lib/astarte_pairing/agent/agent.ex
```diff
@@ -54,4 +63,38 @@ defmodule Astarte.Pairing.Agent do
       Engine.unregister_device(realm, device_id)
     end
   end
+
+  defp dispatch_device_registration_trigger(realm_name, device_id) do
```
noaccOS, line 67: I'm not sold on this function living in the agent, it would make more sense for it to be in its own module just like for other services

with that said astarte pairing is the problem child anyway as it's the only service with an `Engine` module still present so we'll probably have to refactor it down the line anyway

CHANGELOG.md
```diff
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - `HOUSEKEEPING_ASTARTE_KEYSPACE_NETWORK_REPLICATION_MAP` - Datacenter replication map when using NetworkTopologyStrategy (no default, required when using network strategy)
 - Added database events handling configuration across all services:
   - `DATABASE_EVENTS_HANDLING_METHOD` - Controls how database events are handled: "expose" (via telemetry) or "log" (to logs) (default: "expose")
-
+- [astarte_pairing] added ASTARTE_EVENTS_PRODUCER_AMQP_HOST for triggers connection, used to dispatch :on_device_registration trigger
```
noaccOS: 1. don't delete existing newlines (we need a formatter check for markdown tbh)
2. we already have pending changes to the changelog which include documenting the `ASATARTE_EVENTS_PRODUCER_AMQP_*` variables. I think we should start from there, as your current description also does not say anything about other `ASTARTE_EVENTS_` variables. An adequate entry could be something like

```markdown
### Added
- [astarte_pairing] Added device registration triggers

### Changed
- BREAKING: [astarte_pairing] AMQP Producer configuration is now mandatory using the `ASTARTE_EVENTS_PRODUCER_AMQP_*` environment variables
```

3. (and this is related to our workflow not to your pr) we should really decide how to handle the changelog. currently, we're "populating" it through our commit messages, but then environment variables are the exception and need to be populated manually right away. I think we need to sort this out as it's creating friction within the team

PR #1540, opened 2025-10-10
chore(pairing): dependencies upgrade

eddbbt:
upgrade dependencies to bare minimum for libs to work
does not introduce any breaking changes


apps/astarte_pairing/mix.exs
```diff
@@ -96,9 +96,10 @@ defmodule Astarte.Pairing.Mixfile do
       {:dialyxir, "~> 1.0", only: [:dev, :ci], runtime: false},
       {:skogsra, "~> 2.2"},
       {:cors_plug, "~> 2.0"},
-      {:telemetry_metrics, "~> 0.4"},
-      {:telemetry_poller, "~> 0.4"},
-      {:telemetry_metrics_prometheus_core, "~> 0.4"},
+      {:telemetry, "~> 1.0"},
+      {:telemetry_metrics, "~> 0.6.1", override: true},
+      {:telemetry_poller, "~> 1.0", override: true},
+      {:telemetry_metrics_prometheus_core, "~> 1.0.2", override: true},
```
Annopaolo: Would this be feasible? This way we would be on par with Realm Management, without any `override` (it might require some `mix deps.unlock`)
```suggestion
      {:telemetry, "~> 1.0"},
      {:telemetry_metrics, "~> 1.1"},
      {:telemetry_poller, "~> 1.3"},
      {:telemetry_metrics_prometheus_core, "~> 1.2"},
```

eddbbt: looks like the deprecation warnings that I got with the other PR are not showing here

so, i guess it's safe to say yes

PR #1553, opened 2025-10-21
refactor: let astarte_events handle trigger cache

noaccOS:
Move the trigger cache to astarte_events instead of handling it ah-hoc in rm and pairing.
the structure has been generalized so that it can work for Data Updater Plant as well, although the logic to handle data trigger has not been implemented yet


libs/astarte_events/lib/astarte_events/triggers/cache.ex
```diff
@@ -0,0 +1,118 @@
...
+    end)
+    |> Map.get(event_key, [])
+  end
+
+  def fetch_triggers(realm_name, object_id, object_type) do
+    simple_triggers =
+      Queries.query_simple_triggers!(realm_name, object_id, object_type)
+      |> Enum.map(&Core.deserialize_simple_trigger/1)
+
+    # FIXME: this does not work for data triggers because they need the interfaces data
```
eddbbt, line 84: is that something we want addressed in this pr?

noaccOS: no because this pr only tries to give a general structure for a common cache between realm management and pairing, and it works without issues in that regard.
this will be addressed when the work on dup requires this to be implemented, but for the current use case it works flawlessly

PR #1558, opened 2025-10-29
feat(events): handle volatile triggers

noaccOS:
as the volatile triggers are not fetched from the database, the cache
can be optimized by including the event_key in the cache key.

the cache structure is optimized for reading data, as that is by far the
most used operation on the trigger cache.

an additional cache is needed for the deletion operation (in order to
avoid having to iterate the whole ets table to find the trigger_id).

volatile triggers don't use a tll as they can't be read from the database
at expiry.

### DIFF
--- a/libs/astarte_events/.formatter.exs
+++ b/libs/astarte_events/.formatter.exs
@@ -1,4 +1,4 @@
 [
   inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"],
-  import_deps: [:skogsra]
+  import_deps: [:skogsra, :ecto]
 ]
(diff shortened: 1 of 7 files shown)


libs/astarte_events/lib/astarte_events/amqp_events/supervisor.ex
```diff
@@ -0,0 +1,44 @@
...
+        {ExRabbitPool.PoolSupervisor,
+         rabbitmq_config: Config.amqp_options!(), connection_pools: [Config.events_pool_config!()]},
+        id: :events_producer_pool
+      )
+
+    [
+      events_pool,
+      Producer
+    ]
+    |> Supervisor.init(strategy: :rest_for_one)
```
Annopaolo, line 42: This does not look very legible

Annopaolo (approved):
A very minor comment, LGTM

PR #1560, opened 2025-10-29
feat: Dockerize FDO Rendezvous server

mizzet1:
This PR introduce support for the FDO Rendezvous Server, which can now be started and communicate with the other Astarte services via Docker. In particular the Pairing Service, which will be the one responbile for communicating with the Rendezvous Server to perform the FDO protocols.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/config.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/config.ex
@@ -49,6 +49,12 @@ defmodule Astarte.Pairing.Config do
     type: CQExNodes,
     default: [{"localhost", 9042}]
 
+  @envdoc "The URL to access the FDO Rendezvous Server."
+  app_env :fdo_rendezvous_url, :astarte_pairing, :fdo_rendezvous_url,
+    os_env: "PAIRING_FDO_RENDEZVOUS_URL",
+    type: :binary,
+    default: "http://rendezvous:8041"
+
   def init! do
     if {:ok, nil} == ca_cert() do
       case CFSSLCredentials.ca_cert() do
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,6 +87,7 @@ services:
       - ./.env
     environment:
       APPENGINE_API_ROOMS_AMQP_CLIENT_HOST: "rabbitmq"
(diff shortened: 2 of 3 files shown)


AliouneGaye21 (approved):

docker-compose.yml
```diff
@@ -76,8 +76,7 @@ services:
       - "traefik.http.routers.astarte-pairing.service=astarte-pairing"
       - "traefik.http.middlewares.astarte-pairing.stripprefix.prefixes=/pairing"
       - "traefik.http.middlewares.astarte-pairing.stripprefix.forceSlash=false"
-      - "traefik.http.services.astarte-pairing.loadbalancer.server.port=4003"
-
+      - "traefik.http.services.astarte-pairing.loadbalancer.server.port=4003"
```
Annopaolo: Let's keep a newline here

apps/astarte_pairing/lib/astarte_pairing/config.ex
```diff
@@ -49,6 +49,10 @@ defmodule Astarte.Pairing.Config do
     type: CQExNodes,
     default: [{"localhost", 9042}]
 
+  @envdoc "URL to the local dockerized FDO rendezvous server."
```
Annopaolo: This env will set the Rendezvous server address for any Astarte deployment, so it might not always be the case that it is local, nor that it is dockerized. What about this?

```suggestion
  @envdoc "The URL to access the FDO Rendezvous Server"
```

apps/astarte_pairing/lib/astarte_pairing/config.ex
```diff
@@ -49,6 +49,10 @@ defmodule Astarte.Pairing.Config do
...
+  app_env :fdo_rendezvous_url, :astarte_pairing, :fdo_rendezvous_url,
```
Annopaolo, line 53: Let's also expose this as an env var similar to the others, in accordance to the [12-factor app](https://12factor.net/) manifesto. What about `PAIRING_FDO_URL`?
Also, please add a line to the `CHANGELOG.md` documenting the addition of the env var

Annopaolo (changes requested):
Code is good, just a couple nits!

PR #1562, opened 2025-10-29
feat: Handle the first TO0 message

mizzet1:
This PR adds the logic for the Pairing Service to send the first TO0 Protocol message to the Rendezvous Server and to receive the associated HelloAck as response.
An associated module to test the added logic is also included.

### DIFF
--- a/apps/astarte_pairing/README.md
+++ b/apps/astarte_pairing/README.md
@@ -33,7 +33,8 @@ database, (we suggest scylla)
 
 ``` shell
 docker run --rm -d -p 9042:9042 --name scylla scylladb/scylla
-docker run --rn -d --net=host -p 8080/tcp ispirata/docker-alpine-cfssl-autotest:astarte
+docker run --rm  -d -p 5672:5672 -p 15672:15672 --name rabbit rabbitmq:3.12.0-management
+docker run --rm -d --net=host -p 8080/tcp ispirata/docker-alpine-cfssl-autotest:astarte
 ```
 
 
@@ -49,3 +50,29 @@ these resources are located.
 ``` shell
 CASSANDRA_NODES=localhost CFSSL_API_URL=http://localhost:8080 mix test
 ```
+
+# Test FDO
+
+To test FDO, the manufacturer and Device CA keys are required and
+can be generated from the following tools:
+
+# Generate manufacturer keys
+docker run --rm \
+  -v $(pwd)/compose/fdo-keys:/keys \
+  quay.io/fido-fdo/admin-cli:latest \
+  generate-key-and-cert manufacturer \
+  --destination-dir /keys
+
+# Generate device CA keys
+docker run --rm \
+  -v $(pwd)/compose/fdo-keys:/keys \
+  quay.io/fido-fdo/admin-cli:latest \
+  generate-key-and-cert device-ca \
+  --destination-dir /keys
+
+# Set permissions
+chmod 644 compose/fdo-keys/*.pem
+chmod 600 compose/fdo-keys/*.der
+
+An ownership voucher and owner private key are also required.
+For testing purposes, a mock key and voucher have been stored in the priv folder of the Pairing Service.
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
@@ -1,0 +1,69 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2017-2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Pairing.FDOClient do
+  alias Astarte.Pairing.Config
+  alias Astarte.Pairing.TO0Util
+
+  require Logger
+
+  @doc """
+  TO0.Hello - Type 20 message to initiate TO0 protocol
+  Sends an empty array as per FDO specification section 5.3.1
+  Returns decoded TO0.HelloAck (message 21) with rendezvous nonce
+  """
+  def to0_hello() do
+    url = "#{fdo_rendezvous_url!()}/fdo/101/msg/20"
+    headers = [{"Content-Type", "application/cbor"}, {"Content-Length", "0"}]
+    request_body = CBOR.encode([])
+
+    Logger.debug("Sending TO0.Hello to FDO rendezvous server...", url: url)
+
+    case http_client().post(url, request_body, headers) do
+      {:ok, %{status_code: 200, headers: headers, body: body}} ->
+        Logger.debug("TO0.Hello completed successfully")
+        Logger.debug(inspect(headers))
+
+        case TO0Util.get_nonce_from_hello_ack(body) do
+          {:ok, nonce} ->
+            # TODO: to0owner_sign
+            {:ok, nonce}
+
+          {:error, reason} ->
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -30,8 +30,8 @@ services:
   astarte-realm-management:
     image: astarte/astarte_realm_management:1.3-snapshot
     build:
-        dockerfile: apps/astarte_realm_management/Dockerfile
-        context: .
+      dockerfile: apps/astarte_realm_management/Dockerfile
+      context: .
```
eddbbt: avoid unnecessary refactoring, if possible, in order to not pollute the history

docker-compose.yml
```diff
@@ -55,13 +55,14 @@ services:
...
-        dockerfile: apps/astarte_pairing/Dockerfile
-        context: .
+      dockerfile: apps/astarte_pairing/Dockerfile
+      context: .
     env_file:
       - ./.env
     environment:
       PAIRING_CFSSL_URL: "http://cfssl:8080"
       ASTARTE_EVENTS_PRODUCER_AMQP_HOST: "rabbitmq"
+      PAIRING_FDO_RENDEZVOUS_URL: "http://rendezvous:8041"
```
eddbbt: it's the same as the default value, we can
- avoid to define the env in the dockerfile
- set the default value to localhost (see PAIRING_CFSSL_URL)

docker-compose.yml
```diff
@@ -199,6 +199,25 @@ services:
           - dashboard.${DOCKER_COMPOSE_ASTARTE_BASE_DOMAIN}
           - grafana.${DOCKER_COMPOSE_ASTARTE_BASE_DOMAIN}
           - broker.${DOCKER_COMPOSE_ASTARTE_BASE_DOMAIN}
+  
+  rendezvous:
+    image: quay.io/fido-fdo/rendezvous-server:latest
+    container_name: fdo-rendezvous
+    restart: on-failure
+    ports:
+      - "8041:8041"
```
eddbbt: also, exposing the port(s) shall be enough to access the service via localhost, no need to use the container name (test it, to be sure)

apps/astarte_pairing/README.md
```diff
@@ -33,7 +33,8 @@ database, (we suggest scylla)
 
 ``` shell
 docker run --rm -d -p 9042:9042 --name scylla scylladb/scylla
-docker run --rn -d --net=host -p 8080/tcp ispirata/docker-alpine-cfssl-autotest:astarte
+docker run --rm  -d -p 5672:5672 -p 15672:15672 --name rabbit rabbitmq:3.12.0-management
+docker run --rm -d --net=host -p 8080/tcp ispirata/docker-alpine-cfssl-autotest:astarte
```
eddbbt: shall we start also the rendezvous server?

mizzet1: This change was done because with the previous commands mix test was not working on Pairing.
The rendezvous is not required here

apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
```diff
@@ -0,0 +1,69 @@
...
+  Returns decoded TO0.HelloAck (message 21) with rendezvous nonce
+  """
+  def to0_hello() do
+    url = "#{fdo_rendezvous_url!()}/fdo/101/msg/20"
+    headers = [{"Content-Type", "application/cbor"}, {"Content-Length", "0"}]
+    request_body = CBOR.encode([])
+
+    Logger.debug("Sending TO0.Hello to FDO rendezvous server...", url: url)
+
+    case http_client().post(url, request_body, headers) do
```
noaccOS, line 37: this is good as a first step, but I think eventually we will find that creating an ad-hoc httpoison base may be more ergonomic

for reference, we're already using this pattern for [AMQP](https://github.com/astarte-platform/astarte/blob/release-1.3/apps/astarte_housekeeping/lib/astarte_housekeeping/amqp/amqp.ex) and [vhosts](https://github.com/astarte-platform/astarte/blob/release-1.3/apps/astarte_housekeeping/lib/astarte_housekeeping/amqp/vhosts.ex)

apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
```diff
@@ -0,0 +1,69 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2017-2025 SECO Mind Srl
```
noaccOS, line 4: Beware the copyright

```suggestion
# Copyright 2025 SECO Mind Srl
```

apps/astarte_pairing/test/astarte_pairing/fdo/to0util_test.exs
```diff
@@ -0,0 +1,43 @@
+defmodule Astarte.Pairing.TO0UtilTest do
```
noaccOS, line 1: this module is missing the copyright

apps/astarte_pairing/test/astarte_pairing/fdo/to0util_test.exs
```diff
@@ -0,0 +1,43 @@
...
+  use ExUnit.Case, async: true
+
+  alias Astarte.Pairing.TO0Util
+
+  describe "get_nonce_from_hello_ack/1" do
+    test "returns nonce for actual FDO HelloAck CBOR payload (binary nonce)" do
+      valid_nonce = <<32, 54, 127, 243, 66, 48, 228, 115, 59, 186, 230, 246, 198, 179, 113, 78>>
```
noaccOS, line 8: a few things for future reference
- try to keep the test's body as small as possible, and to move every precondition you can to the setup step
- use helper functions and names to describe the behavior of the test: you should be able to understand what it does without knowing implementation details (encoded cbors and the like)
- the data is assumed to be valid unless otherwise stated. this helps getting rid of all the `valid_` prefixes throughout the test suite
- try to always test one thing at a time, if you do two things with two assertions it should be two tests
- don't override the meaning of something: if we define the meaning of `<<1, 2, 3, 4, 5, 6, 7, 8>>` to be a(n invalid) nonce, I may (it happened) spend some time trying to understand why it would behave differently for a `nonce` specifically. keep it as simple as possible

an example below:

```elixir
setup_all do
  nonce = <<32, 54, 127, 243, 66, 48, 228, 115, 59, 186, 230, 246, 198, 179, 113, 78>>
  invalid_nonce = <<1, 2, 3, 4, 5, 6, 7, 8>>

  cbor =
    CBOR.encode([
      %CBOR.Tag{
        tag: :bytes,
        value: <<32, 54, 127, 243, 66, 48, 228, 115, 59, 186, 230, 246, 198, 179, 113, 78>>
      },
      %CBOR.Tag{
        tag: :bytes,
        value: <<32, 54, 127, 243, 66, 48, 228, 115, 59, 186, 230, 246, 198, 179, 113, 78>>
      }
    ])

  %{nonce: nonce, invalid_nonce: invalid_nonce, not_hello_ack_cbor: cbor}
end

defp hello_ack(nonce) do
  CBOR.encode([%CBOR.Tag{tag: :bytes, value: nonce}])
end

describe "get_nonce_from_hello_ack/1" do
  test "returns nonce for actual FDO HelloAck CBOR payload (binary nonce)", %{nonce: nonce} do
    ack = hello_ack(nonce)
    assert {:ok, ^nonce} = TO0Util.get_nonce_from_hello_ack(ack)
  end

  test "fails with wrong length CBOR body", %{invalid_nonce: invalid_nonce} do
    invalid_ack = hello_ack(invalid_nonce)
    assert {:error, {:wrong_cbor_size, invalid_nonce}} == TO0Util.get_nonce_from_hello_ack(invalid_ack)
  end

  test "only decodes valid cbor binaries" do
    assert {:error, {:unexpected_body_format, _}} = TO0Util.get_nonce_from_hello_ack(<<>>)
  end

  test "fails for cbors with unexpected format", %{not_hello_ack_cbor: cbor} do
     assert {:error, {:unexpected_body_format, _}} =
               TO0Util.get_nonce_from_hello_ack(cbor)
  end
end
```

apps/astarte_pairing/README.md
noaccOS, line 37: this change is correct, but is not relevant to the fdo changes
which I don't strictly mind, except it should go in 1.2 (rn -> rm) and 1.3 (rabbit) instead of the feature branch

PR #1565, opened 2025-11-03
feat: TO0.OwnerSign

mizzet1:
This PR adds the logic to build and send the TO0.OwnerSign message after recieving the HelloAck from the Rendezvous during the TO0 protocol.
New tests for this new functionality have also been included.


apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
```diff
@@ -0,0 +1,69 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2017-2025 SECO Mind Srl
```
eddbbt, line 4: check the copyright header, this file is not from 2017

apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
```diff
@@ -0,0 +1,69 @@
...
+        )
+
+      {:error, reason} ->
+        Logger.error("TO0.Hello HTTP request failed", reason: inspect(reason))
+    end
+  end
+
+  defp fdo_rendezvous_url! do
+    Config.fdo_rendezvous_url!()
+  end
```
eddbbt, line 121: just use the config version

apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
```diff
@@ -59,6 +64,58 @@ defmodule Astarte.Pairing.FDOClient do
...
+
+      headers = [
+        {"Content-Type", "application/cbor"},
+        {"Content-Length", byte_size(request_body)},
+        {"Authorization", bearer}
+      ]
+
+      Logger.debug("Sending TO0.OwnerSign", payload_size: byte_size(request_body))
+
+      case http_client().post(url, request_body, headers) do
```
noaccOS, line 101: As with the other pr, I'm afraid this module is doing a little too too much: it's performing both validation and http requests, it gets a little harder to follow

apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
```diff
@@ -59,6 +64,58 @@ defmodule Astarte.Pairing.FDOClient do
...
+        {:ok, %{status_code: 200, body: _body}} ->
+          Logger.debug("TO0.OwnerSign completed successfully!")
+
+        {:ok, %{status_code: status_code, body: body}} ->
+          Logger.error(
+            "TO0.OwnerSign failed with status #{status_code}, response: #{inspect(body)}"
+          )
+
+        {:error, reason} ->
+          Logger.error("TO0.OwnerSign request failed", reason: reason)
```
noaccOS, line 111: `Logger` functions always return `:ok`, you're not just `tap`ping the result with a log

this means that to0 hello will return `:ok` even if the signing failed

apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
```diff
@@ -59,6 +64,58 @@ defmodule Astarte.Pairing.FDOClient do
...
+      end
+    else
+      {:error, reason} ->
+        Logger.error("Failed to prepare TO0.OwnerSign message", reason: reason)
```
noaccOS, line 115: as a rule of thumb, prefer adding logging to the single functions instead of adding an `else` clause to the with

apps/astarte_pairing/lib/astarte_pairing/to0_util.ex
```diff
@@ -41,4 +43,159 @@ defmodule Astarte.Pairing.TO0Util do
...
+      |> String.replace("-----BEGIN OWNERSHIP VOUCHER-----", "")
+      |> String.replace("-----END OWNERSHIP VOUCHER-----", "")
+      |> String.replace(~r/\s/, "")
+
+    with {:ok, cbor_data} <- Base.decode64(ov_data),
+         {:ok, result, _rest} <- CBOR.decode(cbor_data) do
+      {:ok, result}
+    else
+      {:error, reason} -> {:error, reason}
+      other -> {:error, "PEM parsing failed: #{inspect(other)}"}
```
noaccOS, line 63: this does not make it clear where the error comes from.
try to avoid `else` clauses in `with` blocks when possible.

in this case, you could do something like

```elixir
defp cbor_from_ownership_voucher(ov_data) do
  case Base.decode64(ov_data) do
    {:ok, cbor_data} -> {:ok, cbor_data}
    :error -> {:error, "PEM parsing failed"}
  end
end

defp voucher_data_from_pem(voucher_pem) do
  voucher_pem
  |> String.replace("-----BEGIN OWNERSHIP VOUCHER-----", "")
  |> String.replace("-----END OWNERSHIP VOUCHER-----", "")
  |> String.replace(~r/\s/, "")
end

def decode_ownership_voucher(voucher_pem) do
  ov_data = voucher_data_from_pem(voucher_pem)

  with {:ok, cbor} <- cbor_from_ownership_voucher(ov_data),
       {:ok, ownership_voucher, _rest} <- CBOR.decode(cbor) do
    {:ok, ownership_voucher}
  end
end
```

apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
```diff
@@ -59,6 +64,58 @@ defmodule Astarte.Pairing.FDOClient do
...
+  @doc """
+  TO0.OwnerSign - Type 22 message to register ownership.
+  Sends ownership voucher and waits for response from rendezvous server.
+  Returns decoded TO0.AcceptOwner (message 23) with negotiated wait time.
+  """
+  def to0_owner_sign(nonce, bearer) do
+    Logger.debug("Sending TO0.OwnerSign...")
+
+    with {:ok, owner_key} <- TO0Util.get_owner_private_key(),
+         {:ok, ownership_voucher} <- TO0Util.get_ownership_voucher(),
```
noaccOS, line 87: there's a few things I don't understand:
- why do we need mock data? I would understand it in tests but here I think the logic should be generic
- I believe when we start to0 we're already doing it for a specific ownership voucher, so I would expect to0 hello to be parametric on the ov and to get it passed down to owner_sign
- we already have to do the whole cleaning up of the ownership voucher to insert it into the database (`Astarte.Pairing.FDO.OwnershipVoucher.clean_and_b64_decode_voucher_blob`). would it make sense to decode it _once_ and keeping the decoded version in the database/use the decoded version directly in to0 hello? Or do we need the encoded version somewhere down the line?

apps/astarte_pairing/lib/astarte_pairing/to0_util.ex
```diff
@@ -41,4 +43,159 @@ defmodule Astarte.Pairing.TO0Util do
...
+    end
+  end
+
+  def get_owner_private_key() do
+    MockFDOApi.get_owner_private_key()
+  end
+
+  def sign_with_owner_key(data, owner_key) do
+    case :public_key.pem_decode(owner_key) do
+      [{:ECPrivateKey, der_data, :not_encrypted}] ->
```
noaccOS, line 73: this could use a wrapper, so that you could add it to the with

```elixir
with {:ok, der_data} <- decode_owner_key_pem(owner_key),
       ... # rest of with block
```

also, are we sure that `pem_decode`, contrary to `der_decode` does not throw exceptions?

apps/astarte_pairing/lib/astarte_pairing/to0_util.ex
```diff
@@ -41,4 +43,159 @@ defmodule Astarte.Pairing.TO0Util do
...
+    end
+  end
+
+  def safe_sign(data, priv_key) do
+    # ECDSA with SHA-256 requires a 32-byte private key secp256r1
+    if is_binary(priv_key) and byte_size(priv_key) == 32 do
+      try do
+        {:ok, :crypto.sign(:ecdsa, :sha256, data, [priv_key, :secp256r1])}
+      rescue
+        e -> {:error, {:signing_failed, e}}
```
noaccOS, line 112: I think in general it's better to avoid multiple layer errors

```suggestion
        e ->
          Logger.warning("ownership voucher signing failed: #{inspect(e)}")

          {:error, :signing_failed}
```

apps/astarte_pairing/lib/astarte_pairing/to0_util.ex
```diff
@@ -41,4 +43,159 @@ defmodule Astarte.Pairing.TO0Util do
...
+    case byte_size(bin) do
+      32 -> {:ok, bin}
+      n when n < 32 -> {:ok, :binary.copy(<<0>>, 32 - n) <> bin}
+      n -> {:error, {:key_too_long, n}}
+    end
+  end
+
+  def safe_sign(data, priv_key) do
+    # ECDSA with SHA-256 requires a 32-byte private key secp256r1
+    if is_binary(priv_key) and byte_size(priv_key) == 32 do
```
noaccOS, line 108: if the private key comes from the database, it should be validated when we first receive the input, not down the line in here
in general, input validation should be the first operation to perform, and then you can build the rest of your code with the assumption that the data is valid

apps/astarte_pairing/lib/astarte_pairing/to0_util.ex
```diff
@@ -41,4 +43,159 @@ defmodule Astarte.Pairing.TO0Util do
...
+
+  defp build_to1d_rv(entries) do
+    to1d_rv = CBOR.encode([entries])
+    {:ok, to1d_rv}
+  end
+
+  defp build_to0d(ov, wait_seconds, nonce) do
+    to0d = CBOR.encode([ov, wait_seconds, nonce])
+    {:ok, to0d}
+  end
```
noaccOS, line 133: why do we need to wrap in a result something that always succeeds?

apps/astarte_pairing/lib/astarte_pairing/to0_util.ex
```diff
@@ -41,4 +43,159 @@ defmodule Astarte.Pairing.TO0Util do
...
+
+  defp add_cbor_tag(payload) do
+    %CBOR.Tag{tag: :bytes, value: payload}
+  end
+
+  defp build_to1d_to0d_hash(to0d) do
+    to1d_to0d_hash_value = :crypto.hash(:sha256, to0d)
+    # 47 is the key for SHA-256
+    to1d_to0d_hash = CBOR.encode([47, to1d_to0d_hash_value])
```
noaccOS, line 142: usually it's best to have your code speak for itself rather than adding comments

```suggestion
    to1d_to0d_hash = CBOR.encode([@sha256, to1d_to0d_hash_value])
```

apps/astarte_pairing/lib/astarte_pairing/to0_util.ex
```diff
@@ -41,4 +43,159 @@ defmodule Astarte.Pairing.TO0Util do
...
+        add_cbor_tag(payload),
+        add_cbor_tag(raw_signature)
+      ]
+
+      # Tag 18 is associated with COSE_Sign1
+      cose_sign1 = %CBOR.Tag{tag: 18, value: cose_sign1_array}
+      {:ok, cose_sign1}
+    else
+      {:error, reason} -> {:error, {:signing_error, reason}}
+    end
```
noaccOS, line 173: a `with` with only one clause and an `else` block should be a `case`

apps/astarte_pairing/lib/astarte_pairing/to0_util.ex
```diff
@@ -41,4 +43,159 @@ defmodule Astarte.Pairing.TO0Util do
...
+  end
+
+  def get_astarte_rv_to2_addr_entries() do
+    with {:ok, rv_entry1} <- build_rv_to2_addr_entry(CBOR.encode([]), "pippo", 8080, 3),
+         {:ok, rv_entry2} <- build_rv_to2_addr_entry(CBOR.encode([]), "paperino", 8080, 3) do
+      {:ok, [rv_entry1, rv_entry2]}
+    else
+      {:error, reason} -> {:error, reason}
```
noaccOS, line 181: this is the default behavior

```suggestion
```

apps/astarte_pairing/test/astarte_pairing/fdo/to0util_test.exs
```diff
@@ -1,3 +1,21 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2017-2025 SECO Mind Srl
```
noaccOS, line 4: this header should have been amended on the first commit

apps/astarte_pairing/test/astarte_pairing/fdo/to0util_test.exs
```diff
@@ -40,4 +58,137 @@ defmodule Astarte.Pairing.TO0UtilTest do
...
+  defp get_mock_astarte_rv_to2_addr_entries() do
+    with {:ok, rv_entry1} <- TO0Util.build_rv_to2_addr_entry(CBOR.encode([]), "pippo", 8080, 3),
+         {:ok, rv_entry2} <- TO0Util.build_rv_to2_addr_entry(CBOR.encode([]), "paperino", 8080, 3) do
+      {:ok, [rv_entry1, rv_entry2]}
+    else
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  defp get_mock_nonce() do
```
noaccOS, line 133: all the functions below should rather be data set in a setup block and passed in the context to the tests
also no use wrapping in a result tuple

noaccOS (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo_client.ex
mizzet1: The mock data can be removed from the main logic, it was only used to test the to0 flow before the implementation of the  keys retrival api.
I agree that passing the decoded version is a good idea because the encoded version is not needed down the line.

PR #1577, opened 2025-11-10
refactor(dup): use astarte_events for device_error

noaccOS:
removed no-longer relevant test, as astarte_events is now responsible
to find targets

### DIFF
--- a/apps/astarte_data_updater_plant/test/support/cases/trigger.ex
+++ b/apps/astarte_data_updater_plant/test/support/cases/trigger.ex
@@ -19,11 +19,59 @@
 defmodule Astarte.Cases.Trigger do
   use ExUnit.CaseTemplate
 
+  alias Astarte.Core.Triggers.SimpleTriggersProtobuf.AMQPTriggerTarget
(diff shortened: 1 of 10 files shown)


nedimtokic (approved):

apps/astarte_data_updater_plant/test/support/cases/trigger.ex
```diff
@@ -19,11 +19,59 @@
...
+    deserialized_simple_trigger =
+      case protobuf_trigger do
+        %DeviceTrigger{} -> {{:device_trigger, protobuf_trigger}, trigger_target}
+        %DataTrigger{} -> {{:data_trigger, protobuf_trigger}, trigger_target}
+      end
+
+    Astarte.Events.TriggersHandler
+    |> Mimic.stub(:dispatch_event, fn
+      event, event_type, ^trigger_target, realm, hw_id, timestamp, policy ->
+        validation_function &&
```
Annopaolo, line 63: Not a fan of this, as `validation_function` is not a boolean, but as long as Elixir lets you do it...

Annopaolo (approved):

PR #1583, opened 2025-11-11
fix(dup): fix possible crash on interface-specific volatile triggers

noaccOS:
previously, when
1. installing a volatile trigger for a specific interface
2. the interface results in a cache miss but is valid for the device

the volatile trigger would be installed but the device status would not be updated.

this results in a crash when the trigger generated events, as the logic expects the interfaces to be in the data updater state.

by just loading the interfaces normally, we avoid this possible crash


Annopaolo (commented):
I would suggest adding a test that specifically tests (obv) this corner case, in order to be sure not to have regressions in the future. Apart from this, gg!

PR #1584, opened 2025-11-12
chore: remove unused code after astarte_events rework

noaccOS:


noaccOS:
depends on https://github.com/astarte-platform/astarte/pull/1583 to be forward-ported for tests to pass

PR #1586, opened 2025-11-12
test(pairing): Add tests for Rendezvous client

nedimtokic:
Add tests for FDO Rendezvous API client.

### DIFF
--- a/apps/astarte_pairing/test/astarte_pairing/fdo/rendezvous/rendezvous_test.exs
+++ b/apps/astarte_pairing/test/astarte_pairing/fdo/rendezvous/rendezvous_test.exs
@@ -1,0 +1,75 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/rendezvous/rendezvous.ex
```diff
@@ -22,6 +22,8 @@ defmodule Astarte.Pairing.FDO.Rendezvous do
   alias Astarte.Pairing.FDO.Rendezvous.Client
```
eddbbt, line 22: alias no longer required

apps/astarte_pairing/lib/astarte_pairing/fdo/rendezvous/rendezvous.ex
```diff
@@ -22,6 +22,8 @@ defmodule Astarte.Pairing.FDO.Rendezvous do
...
   require Logger
 
+  @client Application.compile_env(:astarte_pairing, :fdo_client, Client)
+
```
eddbbt: I'm not fully sold to the idea that we can have more than one fdo http client
Also, why don't we use mimic? tests would be a lot easier

noaccOS: If we need another client in tests, let's use mimic
- prevents polluting lib code with tests logic
- simpler for my smol brain to follow (doesn't need a config entry + behavior)
- lsp go to definition works

PR #1588, opened 2025-11-12
refactor(pairing): rework ownership voucher api

noaccOS:
now the `/ownership` endpoint both saves the voucher to database and performs to0 with the rendezvous.

the voucher data saved to the database is now cbor encoded, respecting the "blob" database type.

a new module is initialized for the request parameters, just like device registration request and credentials request, which allows using ecto changesets for the data validation.

tests have been removed from the ownership voucher file as the module is no longer the one performing data validation.

Fixes #

* [ ] Yes
* [ ] No

```docs

```


apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/core.ex
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  @doc """
+    Returns the cbor encoded ownership voucher from a given ownership voucher certificate
+  """
+  def binary_voucher(ownership_voucher_certificate) do
+    with %{"voucher" => voucher} <-
+           Regex.named_captures(@ownership_voucher_regex, ownership_voucher_certificate),
+         {:ok, decoded_base64} <- voucher |> String.replace("\n", "") |> Base.decode64() do
+      {:ok, decoded_base64}
```
eddbbt: Base.decode64() already returns {:ok, result}

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/core.ex
```diff
@@ -0,0 +1,78 @@
...
+    else
+      _ -> :error
```
eddbbt: i'll say, same for the else

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/core.ex
```diff
@@ -0,0 +1,78 @@
...
+
+  @doc """
+    Returns the decoded ownership voucher from a given ownership voucher certificate
+  """
+  def decode_ownership_voucher(ownership_voucher_certificate) do
+    with {:ok, decoded_base64} <- Core.binary_voucher(ownership_voucher_certificate),
+         {:ok, decoded_voucher, _rest} <- CBOR.decode(decoded_base64) do
+      {:ok, decoded_voucher}
+    else
+      _ -> :error
```
eddbbt: unnecessary else

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/core.ex
noaccOS: the else is (was) needed because `Regex.named_captures/2` returs `map() | nil`
but you're right that after `binary_voucher` was split to its own function (it was together with `decode_ownership_voucher/1` in a previous version) it no longer makes sense to have the `with` block at all

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/core.ex
noaccOS:
```elixir
@spec decode(binary()) :: {:ok, any(), binary()} | {:error, atom}
```

we change the error from `{:error, atom}` to `:error`

noaccOS: changed the function to return `{:error, atom}` instead so we don't need the `else`.
also, added specs to make it clearer

PR #1589, opened 2025-11-12
chore(pairing): move private key validation and decode during voucher creation

eddbbt:


AliouneGaye21 (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/rendezvous/core.ex
```diff
@@ -101,6 +86,15 @@ defmodule Astarte.Pairing.FDO.Rendezvous.Core do
...
   end
 
+  defp safe_sign(data, private_key) do
+    try do
+      {:ok, :public_key.sign(data, :sha256, private_key)}
+    rescue
+      e ->
+        {:error, :signing_error}
+    end
+  end
```
noaccOS: I'm not sure about this change
If our key retrieval logic is correct, public key sign shouldn't panic
if it's not, we may as well panic

PR #1594, opened 2025-11-14
fix(pairing): add headers to TO0.OwnerSign request

nedimtokic:
Sending TO0.OwnerSign with empty headers fails.
Get headers from HelloAck and use it for sending TO0.OwnerSign message.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/fdo.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/fdo.ex
@@ -22,10 +22,10 @@ defmodule Astarte.Pairing.FDO do
   alias Astarte.Pairing.Queries
 
   def claim_voucher(realm_name, device_id) do
-    with {:ok, nonce} <- hello(),
+    with {:ok, %{nonce: nonce, headers: headers}} <- hello(),
          {:ok, owner_private_key} <- Queries.get_owner_private_key(realm_name, device_id),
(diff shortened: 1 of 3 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/fdo.ex
```diff
@@ -35,8 +35,9 @@ defmodule Astarte.Pairing.FDO do
   Returns decoded TO0.HelloAck (message 21) with rendezvous nonce.
   """
   def hello() do
-    with {:ok, body} <- Rendezvous.send_hello() do
-      RendezvousCore.get_body_nonce(body)
+    with {:ok, %{body: body, headers: headers}} <- Rendezvous.send_hello(),
+         nonce <- RendezvousCore.get_body_nonce(body) do
+      {:ok, %{nonce: nonce, headers: headers}}
```
noaccOS: nit: simple assignments do not need `<-` (they always match)
also, try to keep in the with block assignments as few logic as possible: with blocks are the "fallible" step in the code, so the programmer/reviewer needs extra attention while looking at them, we try to make their job as easy as possible 💫



```suggestion
    with {:ok, %{body: body, headers: headers}} <- Rendezvous.send_hello() do
      nonce = RendezvousCore.get_body_nonce(body)
      {:ok, %{nonce: nonce, headers: headers}}
```

PR #1601, opened 2025-11-19
feat(pairing): Expose endpoint for TO2.HelloDevice

nedimtokic:
This PR implements the POST endpoint for the FDO TO2.HelloDevice message, enabling the secure establishment of communication between the new owner and the device. 
It integrates a dedicated CBOR Plug pipeline to handle the application/cbor request body, ensuring the raw binary data is passed correctly to the controller. 
The controller then decodes the required array of six elements.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/core.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/core.ex
@@ -1,0 +1,51 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Pairing.FDO.OwnerOnboarding.Core do
+  alias Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice
+
+  def decode_hello_device(cbor_hello_device) do
+    case CBOR.decode(cbor_hello_device) do
+      {:ok, decoded_hello_device, ""} ->
+        parse_hello_device(decoded_hello_device)
+
+      _ ->
+        {:error, :invalid_hello_device_message}
+    end
+  end
+
+  defp parse_hello_device(decoded_hello_device) do
+    case decoded_hello_device do
+      [max_size, device_id, nonce_hello_device, kex_name, cipher_name, easig_info] ->
+        hello_device =
+          %HelloDevice{
+            max_size: max_size,
+            device_id: device_id,
+            nonce: nonce_hello_device,
+            kex_name: kex_name,
+            cipher_name: cipher_name,
+            easig_info: easig_info
+          }
+
+        {:ok, hello_device}
+
+      _ ->
+        {:error, :invalid_hello_device_format}
+    end
+  end
+end
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
@@ -1,0 +1,37 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
+  @moduledoc """
+  HelloDevice structure as per FDO specification.
+  """
+  use TypedStruct
+
+  @type sign_info :: {String.t(), binary()}
+
+  typedstruct enforce: true do
+    @typedoc "A hello device message structure."
+
+    field :max_size, non_neg_integer()
+    field :device_id, binary()
+    field :nonce, binary()
+    field :kex_name, String.t()
+    field :cipher_name, String.t()
+    field :easig_info, sign_info()
+  end
+end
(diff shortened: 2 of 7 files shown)


apps/astarte_pairing/lib/astarte_pairing_web/plug/verify_fdo.ex
```diff
@@ -0,0 +1,43 @@
...
+
+  import Plug.Conn
+
+  def init(_opts) do
+    nil
+  end
+
+  def call(conn, _opts) do
+    if get_req_header(conn, "content-type") == ["application/cbor"] do
+      case Plug.Conn.read_body(conn) do
```
noaccOS, line 30: do we have an expected size (or maximum size) for these requests?
by default read_body reads `8_000_000` and if larger returns `{:more, binary(), Conn.t()}`, is it something we need to handle?

apps/astarte_pairing/lib/astarte_pairing_web/router.ex
```diff
@@ -33,6 +33,10 @@ defmodule Astarte.PairingWeb.Router do
     plug Astarte.PairingWeb.Plug.LogHwId
   end
 
+  pipeline :fdo_cbor do
```
noaccOS: I think the cbor part is an implementation detail and I'd leave it out of the pipeline name. Further plugs (if we'll ever add them) may have nothing to do with cbor


```suggestion
  pipeline :fdo do
```

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
```diff
@@ -0,0 +1,39 @@
...
+  use Astarte.PairingWeb, :controller
+
+  alias Astarte.Pairing.FDO.OwnerOnboarding
+
+  require Logger
+
+  action_fallback Astarte.PairingWeb.FallbackController
+
+  def hello_device(conn, _params) do
+    realm_name = conn.params["realm_name"]
```
noaccOS: `map[key]` may hide parameters disappearing after a refactor and result in hard to debug errors

prefer more assertive alternatives such as the dot notation (`conn.assigns.cbor_body`) or `Map.fetch!`

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -0,0 +1,33 @@
...
+# limitations under the License.
+#
+
+defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
+  @moduledoc """
+  HelloDevice structure as per FDO specification.
+  """
+
+  @enforce_keys [:max_size, :device_id, :nonce, :kex_name, :cipher_name, :easig_info]
+  defstruct [
```
noaccOS: would it make sense to use [typedstruct](https://hexdocs.pm/typedstruct/readme.html) here instead? it's usually easier to read with enforced keys, default values and types

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/core.ex
```diff
@@ -0,0 +1,45 @@
...
+             }}
+
+          _ ->
+            {:error, :invalid_hello_device_format}
+        end
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
```
noaccOS, line 50: nit: this is fine but it's a little hard to read due to the many levels of indentation. I suggest splitting up the function a little more


```suggestion
  def decode_hello_device(cbor_hello_device) do
    with {:ok, decoded_hello_device, _} <- CBOR.decode(cbor_hello_device) do
      parse_hello_device(decoded_hello_device)
    end
  end

  defp parse_hello_device do
    case decoded_hello_device do
      [max_size, device_id, nonce_hello_device, kex_name, cipher_name, easig_info] ->
        hello_device = 
          %HelloDevice{
            max_size: max_size,
            device_id: device_id,
            nonce: nonce_hello_device,
            kex_name: kex_name,
            cipher_name: cipher_name,
            easig_info: easig_info
          }

        {:ok, hello_device}

      _ ->
        {:error, :invalid_hello_device_format}
    end
  end
```

apps/astarte_pairing/lib/astarte_pairing_web/router.ex
```diff
@@ -43,6 +47,13 @@ defmodule Astarte.PairingWeb.Router do
       post "/", OwnershipVoucherController, :create
     end
 
+    # Maybe shoule be moved to its own scope?
```
noaccOS: I think it's fine under the current realm api.
I'll double check though thank you

apps/astarte_pairing/lib/astarte_pairing_web/plug/verify_fdo.ex
nedimtokic: For hello_device message we expect to be small size of message, and for following messages we can define maximum size and send it to device on message 61:

See FIDO message 61 specification:
_The maxOwnerMessageSize indicates the maximum sized FIDO Device Onboard message the Owner is able to
receive, buffer, and decode. The Device may use this value to adjust the size of messages sent to the Owner. A
value of zero indicates the default message size._

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/core.ex
```diff
@@ -0,0 +1,47 @@
...
+      [max_size, device_id, nonce_hello_device, kex_name, cipher_name, easig_info] ->
+        hello_device =
+          %HelloDevice{
+            max_size: max_size,
+            device_id: device_id,
+            nonce: nonce_hello_device,
+            kex_name: kex_name,
+            cipher_name: cipher_name,
+            easig_info: easig_info
+          }
```
noaccOS, line 43: the types make sense, however I'm afraid we're missing some cbor tags here.
I think only time will tell

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
```diff
@@ -0,0 +1,39 @@
...
+  def hello_device(conn, _params) do
+    realm_name = Map.fetch!(conn.params, "realm_name")
+    cbor_hello_device = conn.assigns.cbor_body
+
+    {:ok, response_msg} =
+      OwnerOnboarding.hello_device(cbor_hello_device, realm_name)
+
+    conn
+    |> put_resp_content_type("application/cbor")
+    |> send_resp(200, CBOR.encode(response_msg))
```
noaccOS, line 37: this will probably become a with but it's fine to leave it as it is now as we don't know what `OwnerOnboarding.hello_device` will look like

eddbbt (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/core.ex
nedimtokic: I don't have exact example what device will send on `hello device.` Can be fixed after manual testing?

noaccOS: yes of course, implementation details aren't set in stone
however, it may usually be worth leaving a note to future developers (which may as well be yourself in two weeks) that something may be broken. the easiest way is by adding a `TODO` comment

PR #1604, opened 2025-11-20
feat(pairing): Implement FDO ProveOVHdr message 

nedimtokic:
This PR adds the full implementation of the `ProveOVHdr` message, including all steps from voucher retrieval to COSE_Sign1 response generation. It introduces helper functions for OV header extraction, HMAC generation, payload assembly, and hashing of the incoming CBOR message.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
@@ -17,11 +17,119 @@
 #
 
 defmodule Astarte.Pairing.FDO.OwnerOnboarding do
+  @moduledoc """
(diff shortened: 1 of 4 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -17,11 +17,149 @@
...
+        {:ok, public_key_record, private_key_record}
+
+      _ ->
+        Logger.error("Key exchange suite not supported: #{kex_suite_name}")
+        {:error, :unsupported_kex_suite}
+    end
+  end
+
+  def generate_rsa_2048_key() do
+    public_exponent = 65_537
```
eddbbt, line 148: lets put this as a tag

PR #1607, opened 2025-11-21
chore: rudimentary session management

noaccOS:
create a session as defined by section 4.3 of the FDO 1.1 spec using
a single identifier in the authorization header

the current implementation works but has a few caveats:
- it stores the private key to the database (we never did it!)
- uses `System.unique_integer` as the identifier. this is acceptable by
  the spec but not ideal

alternatives considered: storing the sessions in cache
unfortunately, we need the database to store sessions because we don't
have any guarantee the same pairing replica will handle subsequent
messages.

an horde registry might work tho!

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
@@ -1,0 +1,29 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
+  alias Astarte.Pairing.Queries
+
+  def new(realm_name, device_id, public_key, private_key) do
+    key = UUID.uuid4(:raw)
+
+    with :ok <- Queries.store_session(realm_name, key, device_id, public_key, private_key) do
+      {:ok, UUID.binary_to_string!(key)}
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/cache.ex
```diff
@@ -0,0 +1,2 @@
+defmodule AstartePairing.Fdo.OwnerOnboarding.Cache do
```
eddbbt: typo 
```suggestion
defmodule Astarte.Pairing.Fdo.OwnerOnboarding.Cache do
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
+  alias Astarte.Pairing.Queries
+
+  def new(realm_name, device_id, public_key, private_key) do
+    # TODO: find something appropriate
+    key = System.unique_integer()
```
eddbbt: maybe ? idk
```suggestion
    key = :crypto.strong_rand_bytes(16)
```

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/create_request.ex
```diff
@@ -68,31 +69,12 @@ defmodule Astarte.Pairing.FDO.OwnershipVoucher.CreateRequest do
     # SAFETY: we've validated the field is required and we only accept valid changesets
     private_key = fetch_field!(changeset, :private_key)
 
-    case validate_end_extract_private_key(private_key) do
+    case FDOCore.extract_private_key(private_key) do
```
eddbbt, line 72: what does it extract from the private key? imho function name is not descriptive, could be better

noaccOS: it converts the pem string into an erlang's [public_key.private_key()](https://www.erlang.org/doc/apps/public_key/public_key.html#t:private_key/0)

though this is temporary and will change with https://github.com/astarte-platform/astarte/pull/1608

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
noaccOS: it's now using an uuid4 as multiple replicas could still use the same "unique integer" multiple times

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/cache.ex
noaccOS: oh yeah leftover file
removed!

PR #1611, opened 2025-11-25
chore: key exchange

noaccOS:
for now only ECDH256 is implemented

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
@@ -17,13 +17,57 @@
 #
 
 defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
+  use TypedStruct
+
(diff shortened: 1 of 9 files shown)


eddbbt (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
```diff
@@ -17,13 +17,47 @@
...
+
+  alias Astarte.Pairing.FDO.OwnerOnboarding.Session
+  alias Astarte.Pairing.FDO.OwnerOnboarding.SessionKey
   alias Astarte.Pairing.Queries
 
-  def new(realm_name, device_id, public_key, private_key) do
+  typedstruct do
+    field :key, String.t()
+    field :device_id, Astarte.DataAccess.UUID
+    field :nonce, binary()
```
frossq: Better to specify already the nonce type (prove_ov)? Since other nonces will be saved in the session

frossq (approved):

PR #1612, opened 2025-11-25
chore(pairing): automatic session loading

noaccOS:
load the session from the database and reject unauthenticated
connections

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
@@ -19,6 +19,7 @@
 defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
   use TypedStruct
 
+  alias Astarte.DataAccess.FDO.TO2Session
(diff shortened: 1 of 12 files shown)


nedimtokic (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
```diff
@@ -58,4 +59,25 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
...
+        owner_random: owner_random,
+        secret: secret
+      } = database_session
+
+      session = %Session{
+        key: session_key,
+        nonce: prove_ov_nonce,
+        kex_suite_name: kex_suite_name,
+        owner_random: owner_random,
+        secret: secret
```
eddbbt, line 93: missing device_id in struct (?)

noaccOS: right, it was removed for a while and then added back, must have forgotten to add it back here

AliouneGaye21 (approved):

PR #1613, opened 2025-11-26
chore: setup key derivation function

noaccOS:


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
```diff
@@ -17,13 +17,111 @@
...
+    field :xa, binary()
+    field :secret, binary() | nil
+    field :sevk, binary() | nil
+    field :svk, binary() | nil
+    field :sek, binary() | nil
+  end
+
+  def new(realm_name, device_id, kex, cipher_suite, owner_key) do
     key = UUID.uuid4(:raw)
+    prove_ov_nonce = :crypto.strong_rand_bytes(16)
```
frossq, line 44: ProveDv is the nonce that we generate on Astarte side and then save to session; ProveOv is the one that arrives from HelloDevice message

noaccOS: good input! however, this pr does not modify its name, it was from a change in #1611, only the last commit is relevant (unfortunately github does not update existing prs when fast forwarding)

I'd say let's fix that in another pr

PR #1622, opened 2025-11-27
fix(pairing) rewrite payload for message 65

eddbbt:


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -207,63 +206,25 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
...
-    protected_header = %{1 => 5}
-    protected_bin = CBOR.encode(protected_header)
-    mac_structure = ["MAC0", protected_bin, <<>>, payload_bin]
-    to_be_maced = CBOR.encode(mac_structure)
-    tag = :crypto.mac(:hmac, :sha256, session_key, to_be_maced)
+    protected_header_map = %{alg: :es256}
 
-    cose_mac0 = [protected_bin, %{}, payload_bin, tag]
+    res =
+      %Sign1{payload: payload_array, phdr: protected_header_map}
```
noaccOS: the payload may need to be cbor encoded, as `CoseSignature` is defined as

```
CoseSignature = #6.18(CoseSignatureBase)

CoseSignatureBase  = [
   protected  : bytes .cbor $$COSEProtectedHeaders,
   unprotected: $$COSEUnprotectedHeaders
   payload:     bytes .cbor $COSEPayloads, <= this
   signature:   bstr
]
```

(noticed just because I just had the same problem for owner sign)


```suggestion
      payload_array
      |> CBOR.encode()
      |> Sign1.build(protected_header_map)
```

(sign1.build already sets udhr to `%{}` (because struct defaults are out of fashion) and tags the payload as bytes)

AliouneGaye21 (approved):

PR #1623, opened 2025-11-27
chore: validate signature info and save details in the session

noaccOS:


eddbbt (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
```diff
@@ -34,4 +52,66 @@ defmodule Astarte.Pairing.FDO.OwnershipVoucher do
...
+  def decode_cbor(cbor) do
+    case CBOR.decode(cbor) do
+      {:ok, message, _} -> decode(message)
+      _ -> :error
+    end
+
+    with {:ok, message, _} <- CBOR.decode(cbor) do
+      decode(message)
+    end
+  end
```
frossq, line 78: Redundant?

noaccOS: yes, it's a mistake
however, that's from #1620, let's keep this review focused on the last commit

frossq (approved):

PR #1628, opened 2025-11-27
66 and 67

eddbbt:


apps/astarte_pairing/mix.lock
noaccOS, line 18: this should be rebase on the current traget branch instead of duplicating the change

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -0,0 +1,87 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Pairing.FDO.ServiceInfo do
+  alias COSE.Messages.Encrypt0
+
+  @owner_max_service_info 4096
+
+  def handle_msg_66(raw_cose_body, sevk, old_voucher, session_cipher) do
```
noaccOS: it's passable this time, but in general: remember that we have made structures for all to2 messages!

- encoding/decoding should be performed in their own modules, to keep things tidy(er)
- you may find that for the specific messages you're handling, the encoding/decoding you need was already done, so make sure to look out for them!

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -0,0 +1,87 @@
...
+    with {:ok, %{payload: payload, uhdr: uhdr, phdr: phdr}} <-
+           Encrypt0.decrypt_decode(raw_cose_body, session_cipher, sevk),
+         {:ok, [replacement_hmac, _device_max_size], _} <- CBOR.decode(payload),
+         {:ok, _new_voucher} <-
+           generate_replacement_voucher(old_voucher, replacement_hmac) do
+      # TODO: Store `new_voucher` and `device_max_size` in the Session or DB.
+
+      msg_67_payload = [@owner_max_service_info]
+      current_iv = Map.get(uhdr, :iv)
```
noaccOS: this allows current_iv to be nil, which would violently fail on generate_msg_67

instead, you could add to the with block something like

```elixir
{:ok, %CBOR.Tag{tag: :bytes, value: iv}} <- Map.fetch(uhdr, :iv)
```

to fail fast in this function. remember: all validation should be performed asap and you should have your core functionality safe from the outside world

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -0,0 +1,87 @@
...
+      iv.value
+    )
+  end
+
+  defp generate_replacement_voucher(_old_voucher, nil) do
+    # If ReplacementHMac is null, we don't create a new voucher (Credential Reuse)
+    {:ok, nil}
+  end
+
+  defp generate_replacement_voucher(old_voucher_cbor, new_dev_id_hmac) do
```
noaccOS: this function should be in the ownership voucher module, and preferably accept/return `OwnershipVoucher.t()`s (we also have those!) and leave cbor encoding/decoding part out of the equation (it should also be handled by OwnershipVoucher)

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -0,0 +1,87 @@
...
+      _ ->
+        {:error, :failed_66}
+    end
+  end
+
+  def generate_msg_67(payload, sevk, session_cipher, iv, phdr, uhdr) do
+    payload = CBOR.encode(payload)
+
+    Encrypt0.encrypt_encode(
+      %{payload: payload, phdr: phdr, uhdr: uhdr, aad: COSE.tag_as_byte(<<>>)},
```
noaccOS: nit:
```suggestion
      Encrypt0.build(payload, phdr, uhdr),
```

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -0,0 +1,87 @@
...
+      current_iv = Map.get(uhdr, :iv)
+      {:ok, generate_msg_67(msg_67_payload, sevk, session_cipher, current_iv, phdr, uhdr)}
+    else
+      _ ->
+        {:error, :failed_66}
+    end
+  end
+
+  def generate_msg_67(payload, sevk, session_cipher, iv, phdr, uhdr) do
+    payload = CBOR.encode(payload)
```
noaccOS: this is just my assumption but normally when you cbor encode your payload you also need to use `|> COSE.tag_as_byte()` otherwise the CBOR library goes mad

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
noaccOS: it's fine for the demo, but this doesn't take into consideration how to distinguish between sevk and sek+svk

PR #1630, opened 2025-11-27
feat(pairing): TO2 Done2 msg

frossq:
Implementing TO2 Done/Done2 messages (70 & 71), added tests.
Make sure that all nonces are checked correctly and the SetupDvNonce
is saved to TO2 session when needed (msg. 64).

* [X] I have added or ran the appropriate tests

Handle TO2 Done/Done2 msgs

Fixes #

Be kind :smile_cat: 

* [ ] Yes
* [X] No

```docs

```

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
@@ -22,12 +22,12 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
   @known_payload_claims %{
     10 => :nonce,
     256 => :ueid,
-    -257 => :fdo,
-    -259 => :euphnonce
+    -257 => :fdo
   }
 
   @known_unprotected_header_claims %{
-    -258 => :maroeprefix
+    -258 => :maroeprefix,
+    -259 => :euphnonce
   }
 
   def verify_decode_cbor(
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/prove_device.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/prove_device.ex
@@ -22,7 +22,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.ProveDevice do
   From Device ROE to Owner Onboarding Service.
 
   This message contains a COSE_Sign1 object representing an EAT (Entity Attestation Token).
-  It authenticates the device via signature verification and provides the second part 
+  It authenticates the device via signature verification and provides the second part
   of the Key Exchange (xB).
 
   Spec Reference: 5.5.6 TO2.ProveDevice, Type 64 [cite: 3148]
@@ -36,7 +36,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.ProveDevice do
   # --- IANA & FDO Constants (Appendix E & Spec 3.3.6) ---
 
   # EAT-FDO: FDO-specific claim wrapping the FDO payload (Label -257) [cite: 3456]
-  @eat_fdo_label -257
+  @eat_fdo_label :fdo
 
   typedstruct enforce: true do
     @typedoc "Decoded content of the TO2.ProveDevice message."
@@ -44,7 +44,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.ProveDevice do
     # 1. xBKeyExchange
     # The Device's public key contribution to the Key Exchange.
     # Extracted from the EAT Payload inside the EAT-FDO claim.
-    # Spec: TO2ProveDevicePayload = [ xBKeyExchange ] 
+    # Spec: TO2ProveDevicePayload = [ xBKeyExchange ]
     field :xb_key_exchange, binary()
 
     # 2. NonceTO2ProveDv
@@ -68,7 +68,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.ProveDevice do
   end
 
   @doc """
-  Decodes the TO2.ProveDevice message, verifies the COSE signature using the 
+  Decodes the TO2.ProveDevice message, verifies the COSE signature using the
   Device Public Key, and extracts the required FDO fields.
   """
   @spec decode(binary(), any()) :: {:ok, t()} | {:error, atom()}


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -266,4 +268,32 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
 
     {:ok, CBOR.encode(cose_mac0)}
   end
+
+  def done(to2_session, cbor_body) do
+    # retrieve nonce NonceTO2ProveDv from session and check against incoming nonce from device
+    # if == retrieve NonceTO2SetupDv from session and send back to device
+    with {:ok, %DonePayload{nonce_to2_prove_dv: prove_dv_nonce_challenge}} <-
+           DonePayload.decode(cbor_body),
+         true <- prove_dv_nonce_challenge == to2_session.prove_dv_nonce,
```
Annopaolo: Nit: let's use the match operator!

```suggestion
         prove_dv_nonce_challenge = to2_session.prove_dv_nonce,
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -266,4 +268,32 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
...
+      :error ->
+        {:error, "CBOR payload decode error"}
+
+      {:error, err} ->
+        {:error, err}
+
+      false ->
+        # non-matching proveDv nonces
+        # TODO build error msg 101 ??
+        {:error, "ProveDv nonce mismatch"}
```
Annopaolo: Let's build the error here (msg 101)

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -266,4 +268,32 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
...
+    end
+  end
+
+  defp build_done2_message(setup_dv_nonce) do
+    %Done2Payload{:nonce_to2_setup_dv => setup_dv_nonce} |> Done2Payload.encode()
+
+    # TODO put msg in secure tunnel
```
Annopaolo: Nit: putting the message in the secure tunnel should be done in the controller:
- We keep simmetry with what we do for fetching the message from the tunnel
- We clearly isolate the component that act as a boundary with the external world and the internal implementation

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
```diff
@@ -75,4 +75,22 @@ defmodule Astarte.PairingWeb.FDOOnboardingController do
...
     end
   end
+
+  def done(conn, _params) do
+    # TODO extract message from secure tunnel
+    cbor_body = conn.assigns.cbor_body
+    to2_session = conn.assigns.to2_session
+
+    with {:ok, response_msg} <-
+           OwnerOnboarding.done(to2_session, cbor_body) do
```
Annopaolo: Nit: this can be just a case:
```suggestion
    case OwnerOnboarding.done(to2_session, cbor_body) do
      {:ok, response_msg} -> # etc etc...
      {:error, reason} -> # yada yada ...
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
frossq: Not working as expected, leaving the == check

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
frossq: Probably better another PR in which we build the error struct and propagate a more appropriate error handling

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
noaccOS: should've probably been `^prove_dv_nonce_challange <- to2_session.prove_dv_nonce` but it's style

AliouneGaye21 (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -230,4 +232,27 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
...
+    with {:ok, %DonePayload{nonce_to2_prove_dv: prove_dv_nonce_challenge}} <-
+           DonePayload.decode(cbor_body),
+         true <- prove_dv_nonce_challenge == to2_session.prove_dv_nonce do
+      done2_message = build_done2_message(to2_session.setup_dv_nonce)
+      {:ok, done2_message}
+    else
+      :error ->
+        {:error, "CBOR payload decode error"}
+
+      false ->
```
noaccOS: nit: in general, try to avoid matching on the `else` block of `with` s
in this case, you could

- wrap your check to an helper function
```elixir
# in the with block
:ok <- check_dv_nonce(prove_dv_nonce_challange, to2_session.prove_dv_nonce)

# helper function
defp check_dv_nonce(prove_dv_nonce_challange, prove_dv_nonce) do
  # no I'll never use an if block for booleans
  case prove_dv_nonce_challange == prove_dv_nonce do
    true ->
      :ok

    false ->
      # TODO build error message 101 ??
      {:error, "ProveDv nonce mismatch"}
  end
end
```

- match all the errors and return a single error. this is not something you can always do (if you have to build the error message only on once branch of the with, you can't do this obviously)
```elixir
with :ok <- check_a(),
     :ok <- check_b() do
else
  _ -> :error
end
```

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
```diff
@@ -75,4 +75,23 @@ defmodule Astarte.PairingWeb.FDOOnboardingController do
...
+    to2_session = conn.assigns.to2_session
+
+    case OwnerOnboarding.done(to2_session, cbor_body) do
+      {:ok, response_msg} ->
+        conn
+        |> put_resp_content_type("application/cbor")
+        # TODO put msg in secure tunnel
+        |> send_resp(200, response_msg)
+
+      # TODO maybe TO2 session cleanup now
```
noaccOS: probably doing it in the core functionality of your modules make more sense, best to leave only api-level functionality here

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/done.ex
```diff
@@ -39,12 +39,16 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Done do
...
+  Decodes the raw CBOR binary into the Done struct.
   """
-  def to_cbor_list(%__MODULE__{} = t) do
-    [
-      t.nonce_to2_prove_dv
-    ]
+  @spec decode(binary()) :: {:ok, %__MODULE__{}} | {:error, atom()}
+  def decode(cbor_binary) do
+    case CBOR.decode(cbor_binary) do
+      {:ok, [nonce_to2_prove_dv], _rest} when is_binary(nonce_to2_prove_dv) ->
```
noaccOS: just noticed, but cbor decode returns binaries wrapped in a tag


```suggestion
      {:ok, [%CBOR.Tag{tag: :bytes, value: nonce_to2_prove_dv}], _rest} ->
```

frossq: All the encoding/decoding of CBOR payloads needs to be done including the tags, right?

libs/astarte_data_access/lib/astarte_data_access/fdo/to2_session.ex
```diff
@@ -27,6 +27,7 @@ defmodule Astarte.DataAccess.FDO.TO2Session do
     field :epid_group, :binary
     field :device_public_key, :binary
     field :prove_dv_nonce, :binary
+    field :setup_dv_nonce, :binary
```
noaccOS, line 48: do you not need to save this in the database?

PR #1639, opened 2025-11-28
feat(pairing): encode new device credentials and configurations

AliouneGaye21:

Implement CBOR encoding for the payload (to be signed)

Fixes #

* [ ] Yes
* [x] No

```docs

```

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/setup_device.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/setup_device.ex
@@ -32,6 +32,9 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SetupDevicePayload do
   use TypedStruct
   alias Astarte.Pairing.FDO.Types.PublicKey
 
+  alias Astarte.Pairing.FDO.Types.PublicKey
+  alias COSE.Messages.Sign1
+
   typedstruct enforce: true do
     @typedoc "The 4-element payload to be signed inside COSE_Sign1."
 
@@ -63,8 +66,8 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SetupDevicePayload do
   def to_cbor_list(%__MODULE__{} = p) do
     [
       p.rendezvous_info,
-      p.guid,
-      p.nonce_setup_device,
+      COSE.tag_as_byte(p.guid),
+      COSE.tag_as_byte(p.nonce_setup_device),
       PublicKey.encode(p.owner2_key)
     ]
   end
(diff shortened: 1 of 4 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/setup_device.ex
```diff
@@ -40,26 +41,23 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SetupDevicePayload do
...
 
     # 4. Owner2Key
-    # The new Owner Public Key.
-    # The device will store this key and use it to verify future ownership commands.
-    # Note: The COSE_Sign1 wrapping this payload MUST be signed by the private key
-    # corresponding to this public key.
-    field :owner2_key, binary()
+    # The New Owner Public Key.
+    # This is the key the device will trust from now on.
+    field :owner2_key, binary() | map()
```
noaccOS: it's actually a `Astarte.Pairing.FDO.Types.PublicKey.t()`

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/setup_device.ex
```diff
@@ -69,4 +67,14 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SetupDevicePayload do
       p.owner2_key
```
noaccOS: and here we should encode it (by reversing the existing decode function)

noaccOS (changes requested):

apps/astarte_pairing/lib/astarte_pairing/fdo/types/public_key.ex
```diff
@@ -68,4 +68,40 @@ defmodule Astarte.Pairing.FDO.Types.PublicKey do
...
+      encoding: enc,
+      body: body
+    } = pk
+
+    with {:ok, cbor_type} <- encode_type(type),
+         {:ok, cbor_encoding} <- encode_encoding(enc) do
+      cbor_body = %CBOR.Tag{tag: :bytes, value: body}
+
+      [cbor_type, cbor_encoding, cbor_body]
+    end
```
noaccOS: if we're encoding, I would say we're confident our data is valid, so I'd say we can have `encode_type` and `encode_encoding` return the value itself instead (or crash trying) of a result tuple

alternatively, the return value should be changed from `list() | :error` to `{:ok, list()} | :error` and a check should be added in `SetupDevice.to_cbor_list` that the value is `:ok`. I think we can just go with the first option though

noaccOS (changes requested):

apps/astarte_pairing/lib/astarte_pairing/fdo/types/public_key.ex
noaccOS: still relevant @AliouneGaye21

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/setup_device.ex
noaccOS:
```suggestion
p.rendezvous_info,
COSE.tag_as_byte(p.guid),
COSE.tag_as_byte(p.nonce_setup_device),
```

PR #1643, opened 2025-11-28
chore(pairing): device -> owner secure tunnel

noaccOS:
add a pipeline to automagically extract the encoded payload, effectively
performing the device -> owner secure tunnel

depends on #1642

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
@@ -25,6 +25,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
   alias Astarte.Pairing.FDO.OwnerOnboarding.Session
   alias Astarte.Pairing.FDO.OwnerOnboarding.SessionKey
   alias Astarte.Pairing.Queries
+  alias COSE.Messages.Encrypt0
 
   typedstruct do
     field :key, String.t()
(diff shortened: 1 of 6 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
```diff
@@ -103,11 +104,23 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
...
+         :ok <- Queries.add_session_keys(realm_name, session_key, db_sevk, db_svk, db_sek) do
       {:ok, %{session | sevk: sevk, svk: svk, sek: sek}}
     end
   end
 
+  def decrypt_and_verify(%Session{sevk: sevk}, message) when not is_nil(sevk) do
+    cipher = sevk.alg
+
+    with {:ok, enc0} <- Encrypt0.decrypt_decode(message, cipher, sevk),
+         {:ok, body, _} <- CBOR.decode(enc0.payload) do
```
eddbbt, line 117: in some instances we should use also  uhdr and phdr from the the request, not just the decoded body, so i'll add them as well

eddbbt, line 117: in some instances we should use also  uhdr and phdr from the the request, not just the decoded body, so i'll add them as well

noaccOS: I remember looking at the spec and I did not find any such instance
Can you show an example?

eddbbt: 8 was thinking about IV, clearly looking at the wrong spot, your next PR takes care of that aspect

PR #1650, opened 2025-12-01
fix(rm): deletion for long disconnected devices

noaccOS:
dup's rpc server expects an encoded device id for the device deletion
rpc, but rm was sending the decoded device id

### DIFF
--- a/apps/astarte_realm_management/lib/astarte_realm_management/rpc/data_updater_plant/client.ex
+++ b/apps/astarte_realm_management/lib/astarte_realm_management/rpc/data_updater_plant/client.ex
@@ -17,9 +17,13 @@
 #
 
 defmodule Astarte.RealmManagement.RPC.DataUpdaterPlant.Client do
+  alias Astarte.Core.Device
+
   def start_device_deletion_rpc(realm_name, decoded_id) do
+    encoded_device_id = Device.encode_device_id(decoded_id)
+
     server_via_tuple()
-    |> GenServer.call({:start_device_deletion, {realm_name, decoded_id}})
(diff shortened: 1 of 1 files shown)


apps/astarte_realm_management/lib/astarte_realm_management/rpc/data_updater_plant/client.ex
```diff
@@ -17,9 +17,13 @@
 #
 
 defmodule Astarte.RealmManagement.RPC.DataUpdaterPlant.Client do
+  alias Astarte.Core.Device
+
   def start_device_deletion_rpc(realm_name, decoded_id) do
+    hw_id = Device.encode_device_id(decoded_id)
```
Annopaolo: Entia non sunt multiplicanda praeter necessitatem

```suggestion
    encoded_device_id = Device.encode_device_id(decoded_id)
```

noaccOS: from astarte repo's root

```
❯ rg encoded_device_id --only-matching  | wc -l
378
❯ rg hw_id --only-matching  | wc -l
202
```

Annopaolo: An `hw_id` is not the same as an Astarte ID. The Astarte ID is well-defined as a 128-bit binary etc etc, while an "hardware id" can be any binary (that's why we allow for "extended IDs" that are internally collapsed into an Astarte ID).
What we're referring to here is the Astarte Device ID, encoded as defined in astarte_core

PR #1658, opened 2025-12-01
feat(pairing) add max payload size to session and save accordingly

eddbbt:


apps/astarte_pairing/test/astarte_pairing/fdo/service_info_test.ex
```diff
@@ -17,18 +17,40 @@
...
+    {:ok, device_random, xb} = SessionKey.new(hello_device.kex_name, device_key)
+
+    %{
+      hello_device: hello_device,
+      ownership_voucher: ownership_voucher,
+      owner_key: owner_key,
+      device_key: device_key,
+      device_random: device_random,
+      xb: xb
+    }
```
Annopaolo, line 51: Would it make sense to add a whole session here rather than a subset of it?

Annopaolo (approved):

apps/astarte_pairing/test/astarte_pairing/fdo/service_info_test.ex
eddbbt: I agree it's not the best kind of correct rn, but at least it's working. I'll fix later those kind of issues

AliouneGaye21 (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -19,19 +19,23 @@
...
-          max_owner_service_info_sz: _device_max_size
+          max_owner_service_info_sz: device_max_size
         },
         %OwnershipVoucher{} = old_voucher
       ) do
     with {:ok, _new_voucher} <-
-           OwnershipVoucher.generate_replacement_voucher(old_voucher, replacement_hmac) do
-      # TODO: Store `new_voucher` and `device_max_size` in the Session or DB.
+           OwnershipVoucher.generate_replacement_voucher(old_voucher, replacement_hmac),
+         :ok <- Queries.update_session_max_payload(realm_name, session_key, device_max_size) do
```
noaccOS, line 37: eventually it would make sense for all session-updating logic to pass through the session module instead

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -19,19 +19,23 @@
 defmodule Astarte.Pairing.FDO.ServiceInfo do
   alias Astarte.Pairing.FDO.OwnershipVoucher
   alias Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfoReady
+  alias Astarte.Pairing.Queries
 
   @owner_max_service_info 4096
 
   def handle_msg_66(
+        realm_name,
+        session_key,
```
noaccOS, line 28: why not giving this function the whole session struct instead?

apps/astarte_pairing/test/astarte_pairing/fdo/service_info_test.ex
noaccOS: why not creating the session in the setup (not setup_all) and using it in the tests instead? no need to reinvent the wheel, you can copy from `session_test`

noaccOS (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
eddbbt: right, superseeded by https://github.com/astarte-platform/astarte/pull/1665

PR #1667, opened 2025-12-02
feat(pairing): decoding for FDO TO2.DeviceServiceInfo

AliouneGaye21:

Fixes #

* [ ] Yes
* [ ] No

```docs

```

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/device_service_info.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/device_service_info.ex
@@ -25,6 +25,8 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfo do
   It is part of a loop with TO2.OwnerServiceInfo and handles flow control.
   """
   use TypedStruct
+  alias Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfo
+  alias Astarte.Pairing.FDO.Types.ServiceInfo
 
   typedstruct enforce: true do
     @typedoc "Structure for TO2.DeviceServiceInfo message."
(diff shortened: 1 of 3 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/device_service_info.ex
```diff
@@ -47,10 +48,47 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfo do
...
+        {:error, :invalid_structure}
+
+      {:error, _reason} ->
+        {:error, :invalid_cbor}
+    end
+  end
+
+  defp validate_and_build(is_more, info) do
+    with :ok <- validate_bool(is_more),
+         :ok <- validate_list(info) do
```
noaccOS: service info values are cbor encoded from the spec, here we're not decoding them
I suggest adding the type `%Astarte.Pairing.FDO.Types.ServiceInfo{}` (based on [the spec](https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#ServiceInfo)) which, when decoded, decodes the value

```elixir
defmodule Astarte.Pairing.FDO.Types.ServiceInfo do
  use TypedStruct
  alias Astarte.Pairing.FDO.Types.ServiceInfo

  typedstruct do
    field :key, String.t()
    field :value, term()
  end

  def decode(service_info) do
    with [key, value] <- service_info,
         true <- is_binary(key),
         {:ok, value, _} <- CBOR.decode(value) do
      service_info =
        %ServiceInfo{
          key: key,
          value: value
        }

      {:ok, service_info}
    else
      _ ->
        # fallback controller: error 100
        {:error, :message_body_error}
    end
  end

  def decode_map(service_info_list) do
    decoded = service_info_list |> Enum.map(&decode/1)

    Enum.find(decoded, {:ok, decoded}, fn {tag, _} -> tag == :error end)
    |> case do
      {:ok, list} ->
        map =
          Map.new(fn {:ok, value} ->
            %ServiceInfo{key: key, value: value} = value
            {key, value}
          end)

        {:ok, map}

      error ->
        error
    end
  end
end
```

then, you can substitue the existing `:ok <- validate_list(info)` with `{:ok, service_info_map} <- ServiceInfo.decode_map(info)` and use the returned value instead

noaccOS (changes requested):

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/device_service_info_ready_test.exs
```diff
@@ -0,0 +1,98 @@
...
+defmodule Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfoReadyTest do
+  use ExUnit.Case, async: true
+
+  alias Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfoReady
+
+  import Astarte.Helpers.FDO
+
+  describe "decode/1" do
+    test "correctly decodes a full valid payload" do
+      hmac = :crypto.strong_rand_bytes(16)
```
frossq:
```suggestion
      hmac = %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(16)}
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/device_service_info.ex
noaccOS: updated to return a map instead of a list of serviceinfo

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/device_service_info_ready_test.exs
noaccOS: good find! hmacs are not even binaries, they are defined as the Hash type (and we have it already!)

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/device_service_info_test.exs
```diff
@@ -0,0 +1,109 @@
...
+    test "correctly decodes a valid payload (IsMore=false, Empty List)" do
+      cbor_payload = CBOR.encode([false, []])
+
+      assert {:ok, %DeviceServiceInfo{} = msg} = DeviceServiceInfo.decode(cbor_payload)
+      assert msg.is_more_service_info == false
+      assert msg.service_info == []
+    end
+
+    test "correctly decodes a valid payload with data (IsMore=true)" do
+      info_list = ["devmod:os", "linux", "devmod:version", "1.0"]
```
frossq: Vals of the ServiceInfo probably need a CBOR Tag (to be confirmed)

PR #1671, opened 2025-12-03
fix(pairing): handle CBOR tags in message 64

nedimtokic:
Some data are in CBOR tags and needs to be properly extracted

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
@@ -55,8 +55,10 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
   defp verify_payload(payload, extra_claims) do
     known_claims = Map.merge(@known_payload_claims, extra_claims)
 
-    case CBOR.decode(payload) do
-      {:ok, raw_claims = %{}, _} -> {:ok, translate_claims(raw_claims, known_claims)}
+    with %CBOR.Tag{tag: :bytes, value: cbor_payload} <- payload,
+         {:ok, raw_claims = %{}, _} <- CBOR.decode(cbor_payload) do
+      {:ok, translate_claims(raw_claims, known_claims)}
+    else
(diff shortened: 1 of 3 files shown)


apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/prove_device_test.exs
```diff
@@ -40,10 +40,10 @@ defmodule Astarte.Pairing.OwnerOnboarding.Onboarding.ProveDevice do
     eat_claims = %{
       10 => prove_dv_nonce_val,
       256 => guid_val,
-      -257 => [<<>>]
+      -257 => [CBOR.encode([<<>>]) |> COSE.tag_as_byte()]
```
noaccOS:
```suggestion
      -257 => [COSE.tag_as_byte([<<>>])]
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -52,7 +52,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
     translate_claims(unprotected_headers, known_claims)
   end
 
-  defp verify_payload(payload, extra_claims) do
+  defp verify_payload(%CBOR.Tag{tag: :bytes, value: payload}, extra_claims) do
```
noaccOS: I don't love the fact that we're matching on a binary tag here, could this check be moved to the verification?

```elixir
with %CBOR.Tag{tag: :bytes, value: payload} <- payload,
      {:ok, raw_claims = %{}, _} <- CBOR.decode(payload) do
  {:ok, translate_claims(raw_claims, known_claims)}
else
  _ -> :error
end
```

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/prove_device_test.exs
nedimtokic: Chnaged to `[COSE.tag_as_byte(<<>>)]`

PR #1675, opened 2025-12-03
fix(pairing): use secure view for encrypted messages

noaccOS:
properly derive key from session in prove_device


apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/prove_device_test.exs
```diff
@@ -56,40 +103,49 @@ defmodule Astarte.Pairing.OwnerOnboarding.Onboarding.ProveDevice do
         # SetupDvNonce is in unencrypted hdrs
         uhdr: %{-259 => COSE.tag_as_byte(setup_dv_nonce_val)}
       },
-      priv_key_struct
+      priv_key_struct |> dbg()
```
frossq: Intentional?

noaccOS: removed

PR #1676, opened 2025-12-04
fix(fdo): make aead encryption work

joshuachp:


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/core.ex
```diff
@@ -17,17 +17,53 @@
 #
 
 defmodule Astarte.Pairing.FDO.OwnerOnboarding.Core do
-  def counter_mode_kdf(mac_type, mac_subtype, n, secret, context, l) do
-    do_counter_mode_kdf(mac_type, mac_subtype, n, secret, context, l, <<>>)
+  # h – The length of the output of a single invocation of the PRF in bits
+  # r – The length of the binary representation of the counter i
```
noaccOS, line 21: let's go with https://github.com/astarte-platform/astarte/pull/1677 for the kdf

PR #1685, opened 2025-12-04
fix(pairing): fix setup device message

noaccOS:
- use 12 bytes iv for encryption
- properly encode the payload

changes taken from #1676 without the kdf rework


noaccOS:
superseded by #1684 and #1639

PR #1696, opened 2025-12-05
fix(pairing): service info messages

noaccOS:
with this messages 78/79 work

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/device_service_info.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/device_service_info.ex
@@ -60,30 +60,32 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfo do
   Decodes the raw CBOR payload into the struct.
   Validates that IsMoreServiceInfo is a boolean and ServiceInfo is a list.
   """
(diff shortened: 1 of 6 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/device_service_info.ex
```diff
@@ -60,28 +60,30 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfo do
...
-      {:ok, _decoded, _rest} ->
-        {:error, :invalid_structure}
+      {:ok, payload, _} -> decode(payload)
+      _ -> {:error, :message_body_error}
+    end
+  end
 
-      {:error, _reason} ->
-        {:error, :invalid_cbor}
+  def decode(payload) do
```
Annopaolo, line 71: Nit: the difference between `cbor_decode` and `decode` is unclear, I suggest:
```suggestion
  def decode_payload(payload) do
```

noaccOS: It is consistent with the cose library (not only my modifications :innocent:) and by extensions other parts of the pairing codebase

noaccOS: we can think of a general rename later, for now I'll go with my change as it's consistent

PR #1702, opened 2025-12-05
test(pairing): FDO controller tests

nedimtokic:
- The `case` statement in the `done` handler was replaced with a with construct to ensure errors are correctly routed to the fallback controller.
- Added tests for FDO controller.


apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+    conn
+    |> put_req_header("content-type", "application/cbor")
+    |> put_req_header("accept", "application/cbor")
+  end
+
+  defp assert_cbor_error(conn, code, message_id) do
+    assert conn.status == 500
+    assert get_resp_header(conn, "message-type") == ["255"]
+
+    {:ok, [^code, ^message_id, _, _, _], _} = CBOR.decode(conn.resp_body)
```
noaccOS: let's use standard function to fetch the response


```suggestion
    assert get_resp_header(conn, "message-type") == ["255"]
    response = response(conn, 500)

    {:ok, [^code, ^message_id, _, _, _], _} = CBOR.decode(response)
```

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+  end
+
+  defp assert_cbor_ok(conn, expected_map) do
+    assert conn.status == 200
+    {:ok, decoded, ""} = CBOR.decode(conn.resp_body)
+    assert decoded == expected_map
+  end
```
noaccOS: instead of hiding your assertions in private functions, let's try to keep them explicit
in this case, you could


```suggestion
  defp cbor_response(conn) do
    resp = response(conn, 200)
    assert {:ok, decoded, ""} = CBOR.decode(resp)
    decoded
  end
```

this way the user could write

```elixir
assert cbor_response(conn) == expected_map
```

which is more explicit in what you are asserting

(this could and should be applied to the cbor error function, eg by returning a tuple or map with `{code, message_id}`)

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+  alias Astarte.Pairing.FDO.ServiceInfo
+  alias Astarte.Pairing.FDO.OwnerOnboarding.Session
+
+  setup :verify_on_exit!
+
+  defp add_cbor_headers(conn) do
+    conn
+    |> put_req_header("content-type", "application/cbor")
+    |> put_req_header("accept", "application/cbor")
+  end
```
noaccOS: Nit: we could have a `CBORConnCase` to do this

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+
+    %{
+      auth_conn: conn,
+      create_path: fdo_onboarding_path(conn, action, realm),
+      realm_name: realm,
+      message_id: message_id
+    }
+  end
+
+  describe "/hello_device" do
```
noaccOS: this looks like but is not representative of the actual path in the router. I suggest changing it to avoid ambiguity. My pick would be the naming from the FDO spec, but feel free to propose other namings


```suggestion
  describe "HelloDevice" do
```

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+    setup [:hello_device]
+
+    defp hello_device(ctx),
+      do: setup_onboarding(ctx, :hello_device, 60)
+
+    test "returns CBOR ok", %{auth_conn: conn, create_path: path, message_id: id} do
```
noaccOS: each test should be explicative of a single behavior of the code, in this instance it would be more interesting to test that 
- it returns a valid proveovhdr for valid hellodevice messages, or
- it calls/expects something from the hello device function from the owneronboarding controller, which we assume is tested elsewhere

your test is structured more like the second option, so I suggest changing the name to reflect what this does


```suggestion
    test "calls `OwnerOnboarding.hello_device/2`", %{auth_conn: conn, create_path: path, message_id: id} do
```

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+    end
+
+    test "returns error 100", %{auth_conn: conn, create_path: path, message_id: id} do
+      expect(OwnerOnboarding, :hello_device, fn _, _ ->
+        {:error, :message_body_error}
+      end)
+
+      conn = post(conn, path, CBOR.encode(%{hello: "device"}))
+      assert_cbor_error(conn, 100, id)
+    end
```
noaccOS: this test is not interesting as it testing the fallback controller more than anything else
with that said, if you remove the Mimic expectation, you at least test that it expects valid hello device messages


```suggestion
    test "returns message body error when it called with something other than a hello device", %{auth_conn: conn, create_path: path, message_id: id} do
      conn = post(conn, path, CBOR.encode(%{hello: "device"}))
      assert_cbor_error(conn, 100, id)
    end
```

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+
+    test "returns error 500", %{auth_conn: conn, create_path: path, message_id: id} do
+      expect(OwnerOnboarding, :hello_device, fn _, _ ->
+        {:error, %{response: false}}
+      end)
```
noaccOS: this is testing a behavior that shouldn't ever happen. currently this _may_ return 500, such as when we can't find the device id, but this is a bug and should be fixed (Riccardo was looking on this), not something we should test

I wouldn't mind adding a test like this on the fdo fallback controller tests, but I find these unnecessary to have on each function

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+      conn = post(conn, path, CBOR.encode(%{hello: "device"}))
+      assert_cbor_error(conn, 500, id)
+    end
+  end
+
+  describe "/ov_next_entry" do
+    setup [:ov_setup]
+
+    defp ov_setup(ctx),
+      do: setup_authenticated(ctx, :ov_next_entry, 62)
```
noaccOS: nit: I don't find `defp`s inside `describe` particularily easy to read


```suggestion
    setup context do
      setup_authenticated(context, :ov_next_entry, 62)
    end
```

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+
+      expect(DeviceServiceInfoReady, :decode, fn _ -> {:ok, decoded} end)
+
+      expect(ServiceInfo, :handle_msg_66, fn _, _, _ ->
+        {:ok, response}
+      end)
+
+      stub(Session, :encrypt_and_sign, fn _, cbor -> cbor end)
+
+      conn = post(conn, path, CBOR.encode(decoded))
```
noaccOS, line 196: instead of `stub`bing the encrypt and decrypt functions, how about having proper encode/decode in your tests?
if you feel like it's too much work for this pr (understandable), you may add a `TODO` for this and in later iterations we can add a case with a session (as we're already setupping sessions in many test files) and a conn with said session

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
+defmodule Astarte.PairingWeb.FDOOnboardingControllerTest do
```
noaccOS, line 19: missing copyright header

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+      expect(OwnerOnboarding, :hello_device, fn _, _ ->
+        {:error, %{response: false}}
+      end)
+
+      conn = post(conn, path, CBOR.encode(%{hello: "device"}))
+      assert_cbor_error(conn, 500, id)
+    end
+  end
+
+  describe "/ov_next_entry" do
```
noaccOS: most of my comments from hello device apply here as well

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+      expect(OwnerOnboarding, :ov_next_entry, fn _, _, _ ->
+        {:error, %{response: false}}
+      end)
+
+      conn = post(conn, path, CBOR.encode(%{next: "entry"}))
+      assert_cbor_error(conn, 500, id)
+    end
+  end
+
+  describe "/prove_device" do
```
noaccOS: most of my comments apply here as well

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+    end
+
+    test "decode error → 500", %{auth_conn: conn, create_path: path, message_id: id} do
+      expect(DeviceServiceInfoReady, :decode, fn _ -> {:error, :invalid_format} end)
+      conn = post(conn, path, CBOR.encode(%{bad: true}))
+      assert_cbor_error(conn, 500, id)
+    end
+  end
+
+  describe "/service_info_end" do
```
noaccOS: most of my comments apply here as well

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,310 @@
...
+
+    test "500 error", %{auth_conn: conn, create_path: path, message_id: id} do
+      expect(DeviceServiceInfo, :decode, fn _ -> {:error, :invalid_format} end)
+
+      conn = post(conn, path, CBOR.encode(%{}))
+      assert_cbor_error(conn, 500, id)
+    end
+  end
+
+  describe "/done" do
```
noaccOS: most of my comments apply here as well

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
```diff
@@ -71,10 +71,9 @@ defmodule Astarte.PairingWeb.FDOOnboardingController do
   def done(conn, _params) do
     to2_session = conn.assigns.to2_session
 
-    case OwnerOnboarding.done(to2_session, conn.assigns.body) do
-      {:ok, response_msg} ->
-        conn
-        |> render("secure.cbor", %{cbor_response: response_msg})
+    with {:ok, response_msg} <- OwnerOnboarding.done(to2_session, conn.assigns.body) do
+      conn
+      |> render("secure.cbor", %{cbor_response: response_msg})
```
noaccOS, line 76: this technically fixes a crash when we error out on `OwnerOnboarding.done/2`, and since `fix` > `test` the whole commit should be a `fix` (or preferably, you could split this in a separate commit)

noaccOS (changes requested):

apps/astarte_pairing/test/support/cases/cbor_conn_case.ex
```diff
@@ -0,0 +1,58 @@
...
+    end
+  end
+
+  setup _tags do
+    conn = Phoenix.ConnTest.build_conn()
+
+    auth_conn =
+      conn
+      |> Plug.Conn.put_req_header("content-type", "application/cbor")
+      |> Plug.Conn.put_req_header("accept", "application/cbor")
```
noaccOS: this conn is not actually authorized but instead just has some headers set
I would suggest
- adding this headers to `conn`
- not having an `auth_conn` right now, we can do that in a second iteration once we add the `session` to the cases

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -0,0 +1,277 @@
...
+      assert conn.assigns.message_id == id
+    end
+
+    test "returns message body error when it called with something other than a ProveDevice", %{
+      auth_conn: conn,
+      create_path: path,
+      message_id: id
+    } do
+      conn = post(conn, path, CBOR.encode(%{prove: "device"}))
+      # TODO this should return error 100?
```
noaccOS: nit, these should be `FIXME` because they are indeed errors

PR #1705, opened 2025-12-08
chore(pairing): Rendezvous info struct

nedimtokic:
Add Rendezvous info struct and implement decoding and encoding its values.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
@@ -1,0 +1,61 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Pairing.FDO.OwnershipVoucher.RendezvousInfo do
+  @moduledoc """
+  Implementation of FDO RendezvousInfo type.
+  """
+
+  use TypedStruct
+  alias Astarte.Pairing.FDO.OwnershipVoucher.RendezvousInfo
+  alias Astarte.Pairing.FDO.OwnershipVoucher.RendezvousInfo.RendezvousDirective
+
+  typedstruct enforce: true do
+    field :directives, [RendezvousDirective.t()]
+  end
+
+  @doc """
+  Decode CBOR-decoded RendezvousInfo into structs.
+  """
+  @spec decode(list(list())) :: {:ok, t()} | {:error, any()}
+  def decode(directives) when is_list(directives) do
+    directives
+    |> Enum.reduce_while([], fn instructions, acc ->
+      case RendezvousDirective.decode(instructions) do
+        {:ok, directive} -> {:cont, [directive | acc]}
+        {:error, reason} -> {:halt, {:error, reason}}
+      end
+    end)
+    |> case do
+      {:error, reason} ->
+        {:error, reason}
+
+      list ->
+        result = %RendezvousInfo{directives: Enum.reverse(list)}
+        {:ok, result}
+    end
+  end
+
+  def decode(_), do: {:error, :invalid_ownership_voucher}
+
+  @doc "Encode RendezvousInfo into CBOR-compatible terms."
+  @spec encode(RendezvousInfo.t()) :: list()
+  def encode(rv_info) do
+    Enum.map(rv_info.directives, &RendezvousDirective.encode/1)
+  end
+end
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info/rendezvous_directive.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info/rendezvous_directive.ex
@@ -1,0 +1,58 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Pairing.FDO.OwnershipVoucher.RendezvousInfo.RendezvousDirective do
+  @moduledoc """
+  One RendezvousDirective = a list of RendezvousInstr.
+
+  Represents a single directive grouping as defined in the FDO spec,
+  i.e. a list of rendezvous instructions interpreted together.
+  """
+  use TypedStruct
+  alias Astarte.Pairing.FDO.OwnershipVoucher.RendezvousInfo.RendezvousDirective
+  alias Astarte.Pairing.FDO.OwnershipVoucher.RendezvousInfo.RendezvousInstr
+
+  typedstruct enforce: true do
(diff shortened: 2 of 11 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/types/rv_variable.ex
```diff
@@ -0,0 +1,46 @@
...
+          | :medium
+          | :protocol
+          | :delaysec
+          | :bypass
+          | :ext_rv
+
+  @rv_reverse Map.new(@rv_map, fn {k, v} -> {v, k} end)
+
+  def rv_from_u8(u8), do: Map.fetch(@rv_map, u8)
+  def rv_to_u8(atom), do: Map.fetch!(@rv_reverse, atom)
```
noaccOS: 1. the `rv_` prefix doesn't add information as the module name already is `RVVariable`
2. we're currently using `encode` and `decode` for these actions, so I'd suggest `rv_from_u8` -> `decode`, `rv_to_u8` -> `encode`

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
```diff
@@ -0,0 +1,70 @@
...
+
+  def decode(_), do: {:error, :invalid_format}
+
+  defp decode_directive(instrs) when is_list(instrs) do
+    instrs
+    |> Enum.reduce_while([], fn
+      [u8, val], acc ->
+        case RVVariable.rv_from_u8(u8) do
+          {:ok, var} ->
+            instr = %RendezvousInstr{rv_variable: var, rv_value: val}
```
noaccOS: this should be done by rendezvousinstr


```suggestion
      instruction, acc ->
        case RendezvousInstr.decode(instruction) do
          {:ok, instruction} ->
```

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  @doc "Encode RendezvousInfo into CBOR-compatible terms."
+  def encode(directives) when is_list(directives) do
+    Enum.map(directives, &encode_directive/1)
+  end
+
+  defp encode_directive(%RendezvousDirective{instructions: instrs}) do
+    Enum.map(instrs, &encode_instruction/1)
```
noaccOS: in general, let's try to avoid abbreviations


```suggestion
  defp encode_directive(%RendezvousDirective{instructions: instructions}) do
    Enum.map(instructions, &encode_instruction/1)
```

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  defp encode_instruction(%RendezvousInstr{rv_variable: var, rv_value: val}) do
+    [RVVariable.rv_to_u8(var), encode_rv_value(val)]
+  end
+
+  defp encode_rv_value(%CBOR.Tag{} = t), do: t
+  defp encode_rv_value(v) when is_binary(v), do: COSE.tag_as_byte(v)
```
noaccOS: these functions should be moved to their own modules

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
```diff
@@ -0,0 +1,70 @@
...
+
+  defp encode_directive(%RendezvousDirective{instructions: instrs}) do
+    Enum.map(instrs, &encode_instruction/1)
+  end
+
+  defp encode_instruction(%RendezvousInstr{rv_variable: var, rv_value: val}) do
+    [RVVariable.rv_to_u8(var), encode_rv_value(val)]
+  end
+
+  defp encode_rv_value(%CBOR.Tag{} = t), do: t
```
noaccOS: I suggest adding the check that it is in fact a binary, tags can have many forms


```suggestion
  defp encode_rv_value(%CBOR.Tag{tag: :bytes} = t), do: t
```

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
```diff
@@ -0,0 +1,70 @@
...
+        {:error, reason} -> {:halt, {:error, reason}}
+      end
+    end)
+    |> case do
+      {:error, reason} -> {:error, reason}
+      list -> {:ok, Enum.reverse(list)}
+    end
+  end
+
+  def decode(_), do: {:error, :invalid_format}
```
noaccOS: let's return standard errors for consistency, even if it would be the rendezvous which should return it. this should be an `:invalid_ownership_voucher` [spec](https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#ECValues)

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
```diff
@@ -0,0 +1,70 @@
...
+          {:ok, var} ->
+            instr = %RendezvousInstr{rv_variable: var, rv_value: val}
+            {:cont, [instr | acc]}
+
+          :error ->
+            {:halt, {:error, {:invalid_rv_variable, u8}}}
+        end
+
+      _, _acc ->
+        {:halt, {:error, :invalid_instruction_format}}
```
noaccOS: in fact, all of these should be `:invalid_ownership_voucher`

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
noaccOS: also, val should be verified to be `%CBOR.Tag{tag: :bytes, value: value}` and we should only use this last `value`

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
noaccOS: we SHOULD never call this in a situation where this is already encoded


```suggestion
```

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info/rendezvous_directive.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Astarte.Pairing.FDO.OwnershipVoucher.RendezvousInfo.RendezvousDirective do
```
noaccOS, line 19: files missing copyright header

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
nedimtokic: Done inside `RendezvousInstr`

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info.ex
```diff
@@ -0,0 +1,53 @@
...
+
+defmodule Astarte.Pairing.FDO.OwnershipVoucher.RendezvousInfo do
+  @moduledoc """
+  Implementation of FDO RendezvousInfo type.
+  """
+
+  alias Astarte.Pairing.FDO.OwnershipVoucher.RendezvousInfo.RendezvousDirective
+  require Logger
+
+  @type t :: [RendezvousDirective.t()]
```
noaccOS: How about having a structure? like we do for `RendezvousDirective`
it makes it easier to pattern match so on

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/rendezvous_info/rendezvous_instr.ex
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def decode([u8_rv_variable, value]) do
+    with {:ok, rv_variable} <- RVVariable.decode(u8_rv_variable),
+         {:ok, rv_value} <- decode_rv_value(value) do
+      {:ok, %__MODULE__{rv_variable: rv_variable, rv_value: rv_value}}
+    end
+  end
+
+  def encode(%__MODULE__{rv_variable: variable, rv_value: value}) do
```
noaccOS: nit: we usually create an alias for the current module instead


```suggestion
  def encode(%RendezvousInstr{rv_variable: variable, rv_value: value}) do
```

AliouneGaye21 (approved):

PR #1706, opened 2025-12-09
chore(pairing): added correct error logs for hello_device

mizzet1:
THis PR adds added correct error logs for hello_device to match fdo fallback controller logic according to FDO spec.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
@@ -47,6 +47,15 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
     :aes_256_ctr
   ]
 
+  @allowed_kex_names [
+    "DHKEXid14",
+    "DHKEXid15",
+    "ASYMKEX2048",
+    "ASYMKEX3072",
+    "ECDH256",
+    "ECDH384"
+  ]
+
   @type cipher ::
           :aes_128_gcm
           | :aes_192_gcm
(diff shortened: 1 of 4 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
```diff
@@ -71,8 +71,14 @@ defmodule Astarte.Pairing.FDO.OwnershipVoucher do
...
-      {:ok, ownership_voucher_cbor} -> decode_cbor(ownership_voucher_cbor)
-      _ -> :error
+      {:ok, ownership_voucher_cbor} ->
+        decode_cbor(ownership_voucher_cbor)
+
+      {:error, :not_found} ->
+        {:error, :resource_not_found}
+
+      _ ->
+        :error
```
noaccOS: `Repo.fetch_one` can only return `{:ok, data} | {:error, reason}` (`reason` defaults to `:not_found`)


```suggestion
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -73,6 +73,15 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
...
     else
+      {:error, :message_body_error} ->
+        {:error, :message_body_error}
+
+      {:error, :resource_not_found} ->
+        {:error, :resource_not_found}
+
+      {:error, :invalid_message} ->
+        {:error, :invalid_message}
+
```
noaccOS: `with` block automatically bubble up problems!
currently, we have an `else` clause but the reason is to log errors more than anything
I think we can safely keep only the existing `else` logic here


```suggestion
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -99,7 +99,8 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
 
       {:ok, hello_device}
     else
-      _ -> :error
+      {:error, :invalid_message} -> {:error, :invalid_message}
+      _ -> {:error, :message_body_error}
```
noaccOS: nit: usually, it's an anti pattern to perform multiple matches on the `else` block of a with
however for how it's structured the code it's not easy to bubble up simple errors, so it's fine to leave it as it is

my preferred solution would be
- separating `decode` to `cbor_decode` (so that the cbor decode failure is handled on its own)
- having an outer function which wraps something like `{:ok, %{max_size: max_size, ...}}`

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -76,32 +93,67 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
...
+  defp parse_hello_device([
+    max_size,
+    %CBOR.Tag{tag: :bytes, value: device_id},
+    %CBOR.Tag{tag: :bytes, value: nonce_hello_device},
+    kex_name,
+    cipher_name,
+    easig_info
+  ]) do
+    case {decode_kex_name(kex_name), SignatureInfo.decode(easig_info), decode_cipher(cipher_name)} do
+      {{:ok, kex_name}, {:ok, easig_info}, {:ok, cipher}} ->
```
noaccOS: it's probably better to use a with block here otherwise the `:invalid_message` from `decode_cipher` won't bubble up to the the `decode` function


```suggestion
    with {:ok, kex_name} <- decode_kex_name(kex_name),
         {:ok, easig_info} <- SignatureInfo.decode(easig_info),
         {:ok cipher} <- decode_cipher(cipher_name) do
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -76,32 +93,67 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
...
-      hello_device =
-        %HelloDevice{
+    with {:ok, message, _rest} <- cbor_decode(cbor_binary),
+         {:ok, hello_device} <- parse_hello_device(message) do
+      {:ok, hello_device}
+    else
+      {:error, :invalid_cbor} -> {:error, :invalid_message}
+      {:error, reason} -> {:error, reason}
+      _ -> {:error, :message_body_error}
+    end
```
noaccOS: we can trust our own private function to return the correct error

```suggestion
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -76,32 +93,67 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
...
+  defp decode_kex_name(kex_name) do
+    kex_atom =
+      case kex_name do
+        "DHKEXid14" -> :dhkexid14
+        "DHKEXid15" -> :dhkexid15
+        "ASYMKEX2048" -> :asymkex2048
+        "ASYMKEX3072" -> :asymkex3072
+        "ECDH256" -> :ecdh256
+        "ECDH384" -> :ecdh384
+        atom when atom in @allowed_kex_names -> atom
```
noaccOS: we can't receive atoms from the cbor decode function, let's not add additional complexity here

```suggestion
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -76,32 +93,67 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
...
+        _ -> nil
```
noaccOS: instead of adding yet another atom with yet another meaning, we can use the error directly 

```suggestion
        _ -> {:error, :invalid_message}
```

nit: if you also change the other branches of this case to return `{:ok, atom()}` instead, you can then write the check with

```elixir
with {:error, _} <- kex_atom do
  "hello device: received #{inspect(kex_name)} as kex_name"
  |> Logger.error()

  kex_atom
end
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
nedimtokic:
```suggestion
      kex_name: :ecdh256,
```

nedimtokic:
```suggestion
    field :kex_name, kex_name()
```

nedimtokic (changes requested):

PR #1710, opened 2025-12-11
feat(pairing): Handle owner service info max size

nedimtokic:
Refactor OwnerServiceInfo message:
- File name was incorrect
- Move message attributes to the its own module
- Move message build logic to the OwnerOnboarding module
- Use message struct for response
- Update tests

Handle owner service info max size:

If  Device service info message is greater then owner service info message max size  then device should start
sending messages in chunks. 

If DeviceServiceInfo message has `flag is_more_service_info: true` we save chunk and wait for next request.
If DeviceServiceInfo has` is_more_service_info: false` we complete service info and register device.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/device_service_info_ready.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/device_service_info_ready.ex
@@ -75,6 +75,8 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfoReady do
 
   defp validate_and_build(nil, size) do
     with :ok <- validate_size(size) do
+      size = normalize_owner_max_size(size)
+
       message =
         %DeviceServiceInfoReady{
           replacement_hmac: nil,
@@ -87,8 +89,9 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfoReady do
 
   defp validate_and_build(hmac, size) do
     with {:ok, hmac} <- Hash.decode(hmac),
-         :ok <- validate_size(size),
-         size <- normalize_owner_max_size(size) do
+         :ok <- validate_size(size) do
+      size = normalize_owner_max_size(size)
+
       message =
         %DeviceServiceInfoReady{
           replacement_hmac: hmac,
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/owner_service_info.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/owner_service_info.ex
@@ -27,6 +27,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.OwnerServiceInfo do
   use TypedStruct
   alias Astarte.Pairing.FDO.Types.ServiceInfo
   alias Astarte.Pairing.FDO.OwnerOnboarding.OwnerServiceInfo
+  alias Astarte.Pairing.Config
 
   typedstruct enforce: true do
     @typedoc "Structure for TO2.OwnerServiceInfo message."
@@ -69,5 +70,44 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.OwnerServiceInfo do
     t
     |> to_cbor_list()
     |> CBOR.encode()
+  end
+
+  def build(realm_name, credentials_secret, encoded_device_id) do
+    service_info = %{
+      "astarte:active" => true,
+      "astarte:realm" => realm_name,
+      "astarte:secret" => credentials_secret,
+      "astarte:baseurl" => "#{Config.base_url!()}",
+      "astarte:deviceid" => encoded_device_id
+    }
+
+    service_info_keys = Map.keys(service_info)
+    service_info_count = Enum.count(service_info_keys)
+
+    # FIXME: this only works for single message service info
+    modules = [service_info_count, service_info_count | service_info_keys]
+
+    service_info
+    |> Map.put("astarte:modules", modules)
+    |> Map.put("astarte:nummodules", service_info_count)
+
+    %OwnerServiceInfo{
+      is_more_service_info: false,
+      is_done: true,
+      service_info: service_info
+    }
+  end
+
+  def empty() do
+    build_empty_owner_service_info()
+    |> OwnerServiceInfo.encode()
+  end
+
+  defp build_empty_owner_service_info() do
+    %OwnerServiceInfo{
+      is_more_service_info: false,
+      is_done: false,
+      service_info: %{}
+    }
   end
 end
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
@@ -42,6 +42,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
     field :svk, struct() | nil
     field :sek, struct() | nil
     field :max_service_info, integer() | nil
+    field :device_service_info, map() | nil
   end
 
   def new(realm_name, hello_device, ownership_voucher, owner_key) do
@@ -91,6 +92,27 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
     with :ok <- Queries.session_add_setup_dv_nonce(realm_name, session.key, setup_dv_nonce) do
       {:ok, %{session | setup_dv_nonce: setup_dv_nonce}}
     end
+  end
+
+  def add_device_service_info(session, realm_name, new_service_info) do
+    service_info = encode_values_to_cbor(new_service_info)
+    session = update_in(session.device_service_info, &Map.merge(&1 || %{}, service_info))
+
+    with :ok <-
+           Queries.session_add_device_service_info(
+             realm_name,
+             session.key,
+             session.device_service_info
+           ) do
+      {:ok, session}
+    end
+  end
+
+  defp encode_values_to_cbor(map) when is_map(map) do
+    Map.new(map, fn
+      {key, value} ->
+        {key, CBOR.encode(value)}
+    end)
   end
 
   def build_session_secret(session, realm_name, owner_key, xb) do
@@ -154,7 +176,8 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
         sevk: sevk,
         svk: svk,
         sek: sek,
-        max_service_info: max_service_info
+        max_service_info: max_service_info,
+        device_service_info: device_service_info
       } = database_session
 
       session = %Session{
@@ -170,7 +193,8 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
         sevk: SessionKey.from_db(sevk),
         svk: SessionKey.from_db(svk),
         sek: SessionKey.from_db(sek),
-        max_service_info: max_service_info
+        max_service_info: max_service_info,
+        device_service_info: device_service_info
       }
 
       {:ok, session}
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
@@ -19,85 +19,56 @@
 defmodule Astarte.Pairing.FDO.ServiceInfo do
   alias Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfo
   alias Astarte.Pairing.FDO.OwnerOnboarding.OwnerServiceInfo
+  alias Astarte.Pairing.FDO.OwnerOnboarding.Session
   alias Astarte.Core.Device
   alias Astarte.Pairing.Engine
-  alias Astarte.Pairing.Config
 
-  # first device message, sending devmod
-  def handle_message_68(
+  # If device has more data to send, save recevied part to the session
+  # and respond with empty OwnerService Info
+  def build_owner_service_info(
         realm_name,
-        _session,
-        %DeviceServiceInfo{is_more_service_info: false, service_info: service_info}
+        session,
+        %DeviceServiceInfo{
+          is_more_service_info: true,
+          service_info: service_info
+        }
+      ) do
+    Session.add_device_service_info(session, realm_name, service_info)
+    resp = OwnerServiceInfo.empty()
+    {:ok, resp}
+  end
+
+  # Device has no more data to send, this function appends recevied part to the previous recevied
+  # parts of service info and proceed sending OwnerService info
+  def build_owner_service_info(
+        realm_name,
+        session,
+        %DeviceServiceInfo{
+          is_more_service_info: false,
+          service_info: service_info
+        }
       ) do
     # TODO: make use of other device service info
-    device_id = generate_device_id(service_info)
-    encoded_device_id = Device.encode_device_id(device_id)
+    {:ok, session} = Session.add_device_service_info(session, realm_name, service_info)
+
+    encoded_device_id =
+      session.device_service_info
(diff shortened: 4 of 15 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -88,81 +93,162 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+        session,
+        %DeviceServiceInfo{is_more_service_info: true, service_info: service_info}
+      ) do
+    updated_session = %{
+      session
+      | service_info: session.service_info ++ service_info
+    }
+
+    Session.add_device_service_info(updated_session)
+    empty_resp = encode_empty_owner_service_info()
```
noaccOS: I would suggest adding this to the OwnerServiceInfo module


```suggestion
    resp = OwnerServiceInfo.empty()
```

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -88,81 +93,162 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+        _realm_name,
+        session,
+        %DeviceServiceInfo{is_more_service_info: true, service_info: service_info}
+      ) do
+    updated_session = %{
+      session
+      | service_info: session.service_info ++ service_info
+    }
+
+    Session.add_device_service_info(updated_session)
```
noaccOS: Service info are maps.
I would suggest hiding the merging logic away in the Session module

```elixir
def add_device_service_info(session, new_service_info) do
  session = update_in(session.service_info, &Map.merge(&1, new_service_info)
  # store update
end
```

so that this module only contains response logic and not session implementation details

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -88,81 +93,162 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
-        %DeviceServiceInfo{is_more_service_info: false, service_info: service_info}
+        session,
+        %DeviceServiceInfo{
+          is_more_service_info: false,
+          service_info: %ServiceInfo{} = device_service_info
+        }
       ) do
     # TODO: make use of other device service info
-    device_id = generate_device_id(service_info)
+    device_service_info = session.service_info ++ device_service_info
```
noaccOS: let's not forget to update the session

```suggestion
    session = Session.add_device_service_info(device_service_info)
```

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -88,81 +93,162 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+    device_id = generate_device_id(device_service_info)
     encoded_device_id = Device.encode_device_id(device_id)
 
     with {:ok, credentials_secret} <- Engine.register_device(realm_name, encoded_device_id) do
```
noaccOS: with the current logic, subsequent device service info messages would all try to register the device
instead, we should have here something like


```suggestion
    with {:ok, next_owner_service_info} <- Session.next_owner_service_info(session) do
```

and in the session we can
- fetch the next service info package if it is already computed
- register the device, build owner service info packages and store them in the session if not

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -88,81 +93,162 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+        owner_service_info
+        |> Map.put("astarte:modules", modules)
+        |> Map.put("astarte:nummodules", service_info_count)
 
-      owner_service_info = %OwnerServiceInfo{
-        is_more_service_info: false,
-        is_done: true,
-        service_info: service_info
-      }
+      max_size = session.max_service_info || 1300
```
noaccOS: there are two service info values, let's make sure that names are explicit


```suggestion
      max_size = session.max_owner_service_info_size
```

i'd also say that performing the check on insertion instead than on read is cleaner as we're guaranteed to have an integer here

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -88,81 +93,162 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+    %OwnerServiceInfo{
+      is_more_service_info: false,
+      is_done: false,
+      service_info: %{}
+    }
+  end
+
+  defp chunk_service_info(service_info, max_chunk_size) do
+    entries =
+      Enum.map(service_info, fn {k, v} -> [k, v] end)
```
noaccOS:
```suggestion
    entries = ServiceInfo.encode_map(service_info)
```

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -88,81 +93,162 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+
+    Enum.reduce(entries, {[], [], 0}, fn entry, {chunks, current_chunk, current_size} ->
```
noaccOS: you can also use `Enum.chunk_while` to at least not having to track the chunks

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/device_service_info_ready.ex
```diff
@@ -85,7 +87,8 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfoReady do
 
   defp validate_and_build(hmac, size) do
     with {:ok, hmac} <- Hash.decode(hmac),
-         :ok <- validate_size(size) do
+         :ok <- validate_size(size),
+         size <- normalize_owner_max_size(size) do
```
noaccOS: nit: plain matches should go outside the `with` blocks (or use a `=` if they need to be placed within other instructions)

```suggestion
         :ok <- validate_size(size) do
       size = normalize_owner_max_size(size)
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/owner_service_info.ex
```diff
@@ -70,4 +71,49 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.OwnerServiceInfo do
...
+      is_more_service_info: false,
+      is_done: false,
+      service_info: %{}
+    }
+  end
+
+  def to_db(nil), do: nil
+  def to_db(key), do: :erlang.term_to_binary(key)
+  def from_db(nil), do: nil
+  def from_db(key), do: :erlang.binary_to_term(key)
```
noaccOS: since we're not running anymore I would like to have a cleaner solution than storing encoded binaries in the database we then have to parse ourselves

if you need a single structure, you can use a [udt](https://hexdocs.pm/exandra/0.16.0/Exandra.html#module-user-defined-types-udts) or even just a [map](https://hexdocs.pm/exandra/0.16.0/Exandra.html#module-maps), but we can probably do better by separating the single fields

noaccOS (changes requested):
A long needed refactor, thanks :maple_leaf:

apps/astarte_pairing/lib/astarte_pairing/fdo/types/service_info.ex
```diff
@@ -3,19 +3,25 @@ defmodule Astarte.Pairing.FDO.Types.ServiceInfo do
...
   end
 
   def decode(service_info) do
     with [key, value] <- service_info,
          true <- is_binary(key),
-         %CBOR.Tag{tag: :bytes, value: value} <- value,
-         {:ok, value, _} <- CBOR.decode(value) do
+         %CBOR.Tag{tag: :bytes, value: cbor_value} <- value,
+         {:ok, value, _} <- CBOR.decode(cbor_value) do
+      [module, key] = String.split(key, ":", parts: 2)
```
noaccOS: as this is a string we receive from the device (external data), we should not make assertions on the validity of such data.
for this reason, we should place this match in the `with` clauses `[module, key] <- String.split(...)`

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
```diff
@@ -93,6 +95,27 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
     end
   end
 
+  def add_device_service_info(session, realm_name, new_service_info) do
+    service_info_map = service_info_to_map(new_service_info)
```
noaccOS: we should receive a map here. if it isn't so, let's make the change in the deviceserviceinfo module to use decode_map

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -310,6 +310,23 @@ defmodule Astarte.Pairing.Queries do
...
+    Repo.query!(
+      """
+      UPDATE #{keyspace}.to2_sessions
+      SET device_service_info = ?
+      WHERE session_key = ?
+      """,
+      [service_info, session_key]
+    )
+
+    :ok
```
noaccOS: we already have an helper function to update sessions

```suggestion
    updates = [device_service_info: service_info]
    update_session(realm_name, session_key, updates)
```

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/device_service_info_test.exs
```diff
@@ -26,44 +26,57 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfoTest do
...
-        "astarte:baseurl" => "http://api.astarte.localhost",
-        "astarte:deviceid" => "2TBn-jNESuuHamE2Zo1anA",
-        "astarte:modules" => [1, 0, "astarte_interface_1", "astarte_interface_2"],
-        "astarte:nummodules" => 1,
-        "astarte:secret" => "super_secret_credential"
-      }
-
-      assert msg.service_info == expected_map
-    end
+    # test "correctly decodes a Astarte ServiceInfo payload " do
```
noaccOS: should this test work now?

libs/astarte_data_access/lib/astarte_data_access/fdo/to2_session.ex
```diff
@@ -53,6 +53,7 @@ defmodule Astarte.DataAccess.FDO.TO2Session do
     field :sevk, :binary
     field :svk, :binary
     field :sek, :binary
-    field :max_service_info, :integer 
+    field :max_service_info, :integer
+    field :device_service_info, :map
```
noaccOS: `:map` is represented as a [json string](https://hexdocs.pm/exandra/0.16.0/Exandra.html#module-maps) in the database 
here we have a scylladb map, so we want


```suggestion
    field :device_service_info, Exandra.Map, key: Exandra.Tuple, types: [:string, :string], value: :binary
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
```diff
@@ -93,6 +95,27 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
...
+           ) do
+      {:ok, session}
+    end
+  end
+
+  defp service_info_to_map(service_info_list) when is_list(service_info_list) do
+    service_info_list
+    |> Enum.reduce(%{}, fn %ServiceInfo{module: module, key: key, cbor_value: cbor}, acc ->
+      Map.put(acc, {module, key}, cbor)
+    end)
```
noaccOS: (this function shouldn't be needed, but I'll comment for future reference)

If we're building a map from each entry in enum, using a `reduce` is the least legible way to do it

some alternatives:

```elixir
# Map.new is like the reduce implementation but it's more explicit in what it's doing
service_info
|> Map.new(fn %ServiceInfo{module: module, key: key, cbor_value: cbor} ->
  key = {module, key}
  {key, cbor}
end)

# you may also use a for loop with the same syntax, although it's situational
for %ServiceInfo{module: module, key: key, cbor_value: cbor} <- service_info_list, into: %{} do
  key = {module, key}
  {key, cbor}
end
```

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/device_service_info_test.exs
```diff
@@ -8,7 +8,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfoTest do
       payload = [false, []]
       assert {:ok, %DeviceServiceInfo{} = msg} = DeviceServiceInfo.decode(payload)
       assert msg.is_more_service_info == false
-      assert msg.service_info == %{}
+      assert msg.service_info == []
```
noaccOS: from my previous comment, this should indeed be a map

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
nedimtokic: @noaccOS 
If we change `decode_map` to return `%{{module, key} => cbor_value} `then we need to decode value when we need it ?

noaccOS: As a start, let's keep the current implementation of service info decode/decode_map structure: we can incorporate the switch to `key` fro just `key` to `{module, key}`, but let's always return only the decoded value.
If we later find that multiple decoding/encodings are a performance bottleneck (I highly doubt it) then we can re-implement your optimization of saving the cbor_value directly, but for now I'll just say *insert premature optimization quote here*

tldr: let's keep decode_map a `[encoded_service_info] -> %{{module, key} => value}` (or maybe `%{module => %{key, value}}`? it probably depends on how we want to use these value afterwards, for this reason I'd keep the tuple key for now)

nedimtokic: In this case we need to CBOR encode value before saving to DB?

noaccOS: yes, internally we use the decoded value, then for the external world (device, database) we use encoded values
in this instance, it just so happens that we decide to cbor encode in both instances

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/device_service_info_test.exs
nedimtokic: Updated.

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
nedimtokic: Added function to encode values and store it in DB.

apps/astarte_pairing/lib/astarte_pairing/fdo/types/service_info.ex
```diff
@@ -36,8 +39,8 @@ defmodule Astarte.Pairing.FDO.Types.ServiceInfo do
       {:ok, list} ->
         map =
           Map.new(list, fn {:ok, value} ->
-            %ServiceInfo{key: key, value: value} = value
-            {key, value}
+            %ServiceInfo{module: module, key: key, value: value} = value
+            {{module, key}, %{value: value}}
```
noaccOS: why was the value changed to a map?

nedimtokic: Leftover, fixed now.

noaccOS (approved):
Good job 🔥

PR #1711, opened 2025-12-11
fix: crashes on new scylladb versions

noaccOS:
somewhere along the lines scylladb changed the default for 
NetworkReplicationStrategy to use tablets. these are incompatible with
some astarte queries as we need lightweight transactions for some
operations (such as device deletion)

for now, always disable tablets so that the old logic still works.
this is accomplished by always trying to create keyspaces and explicitly
disabling tablets, and in case we're on cassandra or an old version of
scylladb, we catch the failure and try again without specifying an
option (as tablets don't exist at all in those databases)

update scylladb to the new, source available version to align it to the
production environment

### DIFF
--- a/apps/astarte_housekeeping/test/astarte_housekeeping/realms/realms_test.exs
+++ b/apps/astarte_housekeeping/test/astarte_housekeeping/realms/realms_test.exs
@@ -25,13 +25,14 @@ defmodule Astarte.Housekeeping.RealmsTest do
 
   alias Astarte.Core.Generators.Realm, as: GeneratorsRealm
   alias Astarte.DataAccess.Repo
+  alias Astarte.Housekeeping.AMQP
+  alias Astarte.Housekeeping.AMQP.Vhost
   alias Astarte.Housekeeping.Config
+  alias Astarte.Housekeeping.Helpers.Database
   alias Astarte.Housekeeping.Helpers.Database
   alias Astarte.Housekeeping.Realms
   alias Astarte.Housekeeping.Realms.Queries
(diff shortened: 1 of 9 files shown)


nedimtokic (changes requested):
Seems that we are not able to use command `enable` tablets.
Did you tried to configure scylla with tablets disabled :
https://opensource.docs.scylladb.com/stable/reference/configuration-parameters.html#confval-enable_tablets

apps/astarte_housekeeping/test/astarte_housekeeping/realms/realms_test.exs
```diff
@@ -211,6 +212,21 @@ defmodule Astarte.Housekeeping.RealmsTest do
...
       assert {:error, %Ecto.Changeset{}} = Realms.create_realm(@empty_pubkey_attrs)
     end
+
+    @tag :regression
+    test "creates keyspaces where lightweight transactions work", %{realm_name: realm_name} do
+      attrs = %{
+        realm_name: realm_name,
+        jwt_public_key_pem: pubkey(),
+        replication_class: "NetworkTopologyStrategy",
+        datacenter_replication_factors: %{replication_factor: 1}
```
frossq: Here there needs to be a map with the key(s) being the name(s) of the datacenter(s), e.g.
```
                replication_class: "NetworkTopologyStrategy",
                datacenter_replication_factors: %{"datacenter1" => 1}
```

frossq (approved):
:fire:

lusergit (approved):
🐊

PR #1714, opened 2025-12-15
chore(pairing): added correct error logs for device_service_info

mizzet1:
THis PR adds added correct error logs for device_service_info to match fdo fallback controller logic according to FDO spec.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
@@ -160,7 +160,7 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
 
   # end
   def handle_message_68(_realm_name, _session_key, _) do
-    {:error, :invalid_payload}
+    {:error, :message_body_error}
   end
 
   defp generate_device_id(%{"devmod:sn" => sn}), do: UUID.uuid5(:oid, sn, :raw)


apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -84,7 +84,7 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
       {:ok, msg_67_payload}
     else
       _ ->
-        {:error, :failed_66}
+        {:error, :invalid_message}
```
nedimtokic: This function is called after successful decoding of message 66, so I don't expect this to return such an error.
Instead, I would suggest that decoding of message 66 and `handle_msg_66` be combined into one function in the OwnerOnboarding module, as we did for the previous endpoints.

If we want to be consistent, then we should move `handle_msg_66` to the OwnerServiceInfoReady module and prepare this response there. Currently, the OwnerServiceInfoReady module is not used anywhere.

In my opinion, `handle_msg_68` belongs to the OwnerServiceInfo module.
@noaccOS

noaccOS: I'll start by saying that this is not a refactor pr but its scope is message errors
I do agree that a refactor is needed (although I like the current approach of many other modules of keeping encode/decoding and not much more functionality in the message module, and glue application logic outside (usually in owner_onboarding), so I'd keep decoding and handle separate

noaccOS: The table in the [spec](https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#ECValues) does not report invalid_message as a valid error code for device_service_info_ready: we're not failing a verification from the device, if something bad happened it's probably our internal logic error and so returning 500 is actually appropriate here

PR #1715, opened 2025-12-15
ci: optimize build caching with matrix strategy

lucamarchiori:
The previous build approach utilizing `docker/bake-action` with `docker-compose` proved ineffective for caching in the CI environment. The monolithic build context often resulted in frequent cache misses, causing full rebuilds of the entire stack even when only isolated changes were made. This yielded no tangible performance benefits and increased CI duration.

This commit completely refactors the `astarte-build` workflow to use a job matrix strategy.
- **Parallel Builds:** Each Astarte application is now built in its own isolated job.
- **Granular Caching:** Caching is now explicitly scoped per-application and per-branch (using a slugified branch name). This prevents cache thrashing between branches and ensures that a change in one microservice does not invalidate the build cache for others.

Additionally, the `astarte-e2e-build` workflow has been updated to adopt the same branch-scoped caching strategy to ensure consistent performance across all build pipelines.

### DIFF
--- a/.github/workflows/astarte-e2e-build-workflow.yaml
+++ b/.github/workflows/astarte-e2e-build-workflow.yaml
@@ -14,6 +14,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: docker/setup-buildx-action@v3
+      - name: Build branch slug env variable (job-scoped)
+        run: |
+          # Slugify branch/tag name for Docker tag safety and consistency
+          echo "BRANCH_SLUG=$(echo "${GITHUB_REF_NAME}" | sha1sum | cut -f 1 -d ' ')" >> $GITHUB_ENV
       - name: Build Astarte E2E
         uses: docker/build-push-action@v6
         with:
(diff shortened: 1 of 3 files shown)


noaccOS (approved):

.github/workflows/astarte-e2e-build-workflow.yaml
```diff
@@ -14,15 +14,22 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: docker/setup-buildx-action@v3
+      - name: Build repo name env variable (job-scoped)
+        run: |
+          echo "REPO_NAME=${GITHUB_REPOSITORY,,}" >> $GITHUB_ENV
```
noaccOS: not sure where this is being used?

.github/workflows/astarte-e2e-build-workflow.yaml
```diff
@@ -14,15 +14,22 @@ jobs:
...
+      - name: Build branch slug env variable (job-scoped)
+        run: |
+          # Slugify branch/tag name for Docker tag safety and consistency
+          echo "BRANCH_SLUG=$(echo "${GITHUB_REF_NAME}" | tr '[:upper:]' '[:lower:]' | sed 's/[^a-z0-9_.-]/-/g')" >> $GITHUB_ENV
```
noaccOS: this may create quite a few conflicts (because people don't use jj)
how about using a sha instead?


```suggestion
          echo "BRANCH_SLUG=$(echo "${GITHUB_REF_NAME}" | sha1sum | cut -f 1 -d ' ')" >> $GITHUB_ENV
```

.github/workflows/astarte-e2e-build-workflow.yaml
lucamarchiori: Left by mistake after some trials with GHCR

PR #1717, opened 2025-12-15
build(docker): optimize docker build

noaccOS:
clean up docker build process by
- only keeping relevant files in the context by adding more files to the dockerignore
- remove unnecessary files from the container
- increasing cacheability to avoid re-downloading and compiling all dependencies when a shared dependency is updated
- align builder and runner container

on my machine, fixes build on docker 29, which failed due to
- keys from docker-compose-initializer being included in the build context
- libs having a wrong relative location to the context


nedimtokic (approved):

AliouneGaye21 (approved):

lusergit (approved):
👍

PR #1726, opened 2025-12-17
support ECDSA SECP384R1 

AliouneGaye21:

This PR introduces High Security FIDO Device Onboarding (FDO) support by enabling the use of the NIST P-384 curve (SECP384R1) and ES384 signing algorithm.
* **Dynamic Crypto Selection:** `OwnerOnboarding` now dynamically detects the owner's key type (P-256 or P-384) and selects the appropriate signing algorithm (ES256 or ES384) and hash function (SHA-256 or SHA-384).
* **Key Exchange Strategy:** Introduced `KeyExchangeStrategy` module to implement FDO Spec 3.6.5 logic. It selects the correct suite (ECDH256 or ECDH384) based on the intersection of Device and Owner capabilities.
* **Session & KDF:** Updated `SessionKey` and `Session` modules to support SHA-384 based Key Derivation Functions (KDF) and variable AES key lengths.
* **COSE Signing:** `ProveOVHdr` now correctly encodes the signing algorithm in the COSE protected header based on the selected suite.
* **Testing:** Added  unit tests for P-384 flows, including `KeyExchangeStrategy` selection logic and full `hello_device` integration tests with mocked P-384 crypto material.

Fixes #

* [ ] Yes
* [x] No

```docs

```


nedimtokic (changes requested):
Fix file names:
- key_exchange_stategy.ex - > key_exchange_strategy.ex
- kex_exchange_strategy_test.exs ->  key_exchange_strategy_test.exs

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/key_exchange_stategy.ex
```diff
@@ -0,0 +1,56 @@
...
+        {:ok, @ecdh256}
+
+      {:es384, %ECC{crv: :p384}} ->
+        {:ok, @ecdh384}
+
+      {:es384, %ECC{crv: :p256}} ->
+        {:ok, @ecdh256}
+
+      {:es256, %ECC{crv: :p384}} ->
+        {:ok, @ecdh256}
```
frossq: `{:ok, @ecdh384}`

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -53,7 +53,10 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
          {:ok, session} <-
```
frossq, line 61: Probably we need to select the algorithm to be used before creating the session

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
```diff
@@ -43,7 +47,8 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SessionKey do
       key.y::binary, blen_r::integer-unsigned-size(16), random::binary>>
   end
 
-  def compute_shared_secret("ECDH256", %ECC{} = owner_key, owner_random, xb) do
+  def compute_shared_secret(suite, %ECC{} = owner_key, owner_random, xb)
+      when suite in ["ECDH256", "ECDH384"] do
     {device_random, device_public} = parse_xb_ecdh(xb)
```
frossq, line 52: We also need to consider the different xb formats in `parse_xb_ecdh`

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
```diff
@@ -0,0 +1,196 @@
...
+               OwnerOnboarding.hello_device(realm_name, cbor_hello)
+
+      assert session_key == fake_session_key
+      assert {:ok, sign1_msg} = Sign1.decode_cbor(resp_binary)
+
+      assert sign1_msg.phdr.alg == :es384
+    end
+  end
+
+  defp generate_p384_data_and_pem do
```
frossq: Good candidate to be put in Helpers.FDO module and allow to generate custom vouchers for tests

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/key_exchange_stategy.ex
AliouneGaye21: Good take

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
AliouneGaye21: You're right. I have already implemented the module for making choice. Just neeed to use it here

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
AliouneGaye21: Actually the current implementation of parse_xb_ecdh already handles different key sizes dynamically by reading the length prefixes (blen_x, blen_y). It inherently supports both P-256 and P-384 formats.

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
AliouneGaye21: Nice to have

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
```diff
@@ -34,6 +34,10 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SessionKey do
     {:ok, random, xa}
   end
 
+  def new(suite, _) do
+    {:error, {:unsupported_kex_suite, suite}}
```
noaccOS: please only return valid errors for the fallback controller. I think this should be an `{:error, :invalid_message}` (INVALID_MESSAGE_ERROR)

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
```diff
@@ -63,27 +68,47 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SessionKey do
...
+      :crypto.compute_key(:ecdh, device_public, owner_key.d, curve)
 
     <<shared_secret::binary, device_random::binary, owner_random::binary>>
   end
 
-  def derive_key(:aes_256_gcm, shared_secret, _owner_random) do
+  def derive_key(kex_suite_name, cipher_name, shared_secret, _owner_random) do
     context_random = <<>>
 
+    {kdf_hash_alg, kdf_len} =
```
noaccOS: `derive_sevk` is already the generic function to generate a sevk based on the given parameters. this implementation moves the complexity to 3 different places (derive key function call, `kex_suite_name` match and `get_key_length_bits` call). I suggest duplicating the original `derive_key` function so that we only have one level of generics and it's easier to follow

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
```diff
@@ -63,27 +68,47 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SessionKey do
...
+      key_length_bits,
+      kdf_len * 8
     )
   end
 
+  defp get_key_length_bits(cipher) do
+    case cipher do
+      c when c in [:aes_128_gcm, :aes_128_ctr, :aes_128_cbc] -> 128
+      c when c in [:aes_192_gcm] -> 192
+      _ -> 256
```
noaccOS: these defaults are sources of bugs, in general it's better to be explicit

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
```diff
@@ -63,27 +68,47 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SessionKey do
...
+      cipher_name,
       :hmac,
-      :sha256,
+      kdf_hash_alg,
       shared_secret,
       context_random,
-      256,
-      256
+      key_length_bits,
+      kdf_len * 8
```
noaccOS: where did you find that the output length depends on the kex suite name?

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -52,10 +53,21 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
          {:ok, ownership_voucher} <- OwnershipVoucher.fetch(realm_name, device_id),
          {:ok, owner_private_key} <- fetch_owner_private_key(realm_name, device_id),
          {:ok, pub_key} <- OwnershipVoucher.owner_public_key(ownership_voucher),
+         {:ok, selected_alg} <-
+           KeyExchangeStrategy.select(hello_device.easig_info, owner_private_key),
```
noaccOS: we should not select an algorithm but instead we may validate that the one included in hello device is valid. this implementation is ignoring the device's choice

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -52,10 +53,21 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
...
+             selected_alg,
+             hello_device,
+             ownership_voucher,
+             owner_private_key
+           ) do
       num_ov_entries = Enum.count(ownership_voucher.entries)
-      hello_device_hash = Hash.new(:sha256, cbor_hello_device)
+
+      {signing_alg, hash_alg} = get_crypto_params(owner_private_key)
+      hello_device_hash = Hash.new(hash_alg, cbor_hello_device)
```
noaccOS: where did you find that the hello device hash algorithm depends on the owner private key?

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -285,4 +303,14 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
...
+
+  defp get_crypto_params(%COSE.Keys.ECC{crv: crv}) do
+    case crv do
+      :p256 -> {:es256, :sha256}
+      :secp256r1 -> {:es256, :sha256}
+      :p384 -> {:es384, :sha384}
+      :secp384r1 -> {:es384, :sha384}
+      _ -> {:es256, :sha256}
+    end
+  end
```
noaccOS, line 313: this not only contains invalid entries (why are there curve entries mixed in??), but the first tuple element is just `key.alg`

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
```diff
@@ -0,0 +1,152 @@
...
+
+    test "High Security P-384: responds with ES384 signed ProveOVHdr", %{realm: realm_name} do
+      {voucher_p384, priv_key_pem} = generate_p384_data_and_pem()
+      device_id = voucher_p384.header.guid
+
+      insert_voucher(realm_name, priv_key_pem, <<>>, device_id)
+
+      Mimic.expect(OwnershipVoucher, :fetch, fn ^realm_name, ^device_id ->
+        {:ok, voucher_p384}
+      end)
```
noaccOS: can't you insert the valid voucher instead of using mocks?

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
```diff
@@ -0,0 +1,152 @@
...
+               OwnerOnboarding.hello_device(realm_name, cbor_hello)
+
+      assert is_binary(session_key)
+
+      assert {:ok, sign1_msg} = Sign1.decode_cbor(resp_binary)
+
+      assert sign1_msg.phdr.alg == :es256
+    end
+
+    test "High Security P-384: responds with ES384 signed ProveOVHdr", %{realm: realm_name} do
```
noaccOS: (apart from the fact that I'm not sure what High Security means here but) 
this test is on the track to become the next simple flow.
it's overly complicated because it's hard to pinpoint all the stuff you need from the getgo, but I suggest adding a setup with
- two devices: one uses p256 the other p384
- each test can use its device `%{device_p256: device_id} = context`
- valid hello device messages do not need 3 different mocks ;3

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/session_test.exs
```diff
@@ -120,4 +122,50 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SessionTest do
       assert alg == :aes_256_gcm
     end
   end
+
+  describe "derive_key/2 with P-384 (ECDH384)" do
+    setup %{realm: realm_name} do
+      p384_device_key = COSE.Keys.ECC.generate(:es384)
+      p384_owner_key = COSE.Keys.ECC.generate(:es384)
+
+      hello_device = %{HelloDevice.generate() | kex_name: "ECDH384", easig_info: :es384}
```
noaccOS: maybe it's time to give HelloDevice.generate an opts parameter

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/session_test.exs
```diff
@@ -120,4 +122,50 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SessionTest do
...
+          kex_suite_name: "ECDH384",
+          cipher_suite_name: :aes_256_gcm,
+          owner_random: owner_random,
+          prove_dv_nonce: :crypto.strong_rand_bytes(16)
+        }
+
+      Astarte.Pairing.Queries.store_session(realm_name, session_key, dummy_session_params)
+
+      {:ok, session_with_secret} =
+        Session.build_session_secret(session, realm_name, p384_owner_key, xb)
```
noaccOS: wouldn't it be easier to use Session.new |> Session.build_session_secret? why duplicate the logic in tests?

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +170,68 @@ defmodule Astarte.Helpers.FDO do
...
+      nil,
+      "p384-device",
+      %CBOR.Tag{tag: :bytes, value: pub_key_binary},
+      nil
+    ]
+
+    valid_cbor_header = CBOR.encode(fake_header_list)
+
+    dummy_hmac = :crypto.strong_rand_bytes(48)
+    valid_cbor_hmac = CBOR.encode(%CBOR.Tag{tag: :bytes, value: dummy_hmac})
```
noaccOS: an hmac is not a binary it's a `Hash`

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +170,68 @@ defmodule Astarte.Helpers.FDO do
...
+
+    guid_raw = UUID.uuid4(:raw)
+
+    header = %OwnershipVoucher.Header{
+      guid: guid_raw,
+      device_info: "p384-device",
+      public_key: %PublicKey{type: :secp384r1, encoding: :x509, body: pub_key_binary}
+    }
+
+    fake_header_list = [
```
noaccOS: avoid calling all data "fake" "dummy" or whatever
let's try to use data as close to real as they can be, and avoid using adjectives which don't add meaning

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +170,68 @@ defmodule Astarte.Helpers.FDO do
...
+        type: :secp384r1,
+        encoding: :x509,
+        body: pub_key_binary
+      }
+    }
+
+    %OwnershipVoucher{
+      header: header,
+      hmac: nil,
+      entries: [],
```
noaccOS: an ownership voucher without entries will crash in parts of to2
let's try to always use valid data throughout, otherwise we'll need to revisit these tests each time we make adjustments to the code

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
AliouneGaye21: The dependency comes from the "Cipher Suite Names and Meanings" table in the FDO specification. Spec 4.4

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
AliouneGaye21: Yes, actually this implementation in based on the map on on the spec 3.6.5. The naming of the module may be validate. Wich is actually what it is doing based on the device attestation and the owner attestation. In the mapping when they don't present the same type we have not recommended algorithms, but they are included

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
AliouneGaye21: Image
From the spec 5.5.3. TO2.ProveOVHdr

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
AliouneGaye21: Thanks for the catch on key.alg, you are right. Using the curve aliases there was messy and key.alg is definitely better. I've refactored get_crypto_params to match directly on the algorithm field.

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
noaccOS: It means that the hash is contained inside a `Sign1` message which is signed using the owner's signature, it does not contain any reference to the algorithm used for the hashing

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/key_exchange_strategy.ex
```diff
@@ -0,0 +1,54 @@
...
+        {:ok, @ecdh256}
+
+      {@ecdh384, %ECC{crv: crv}} when crv in [:p384, :secp384r1] ->
+        {:ok, @ecdh384}
+
+      {requested, %ECC{crv: actual_crv}} when requested in @supported_suites ->
+        {:error, {:kex_suite_mismatch, requested: requested, owner_curve: actual_crv}}
+
+      _ ->
+        {:error, :invalid_kex_validation_params}
```
noaccOS:
```suggestion
      _ ->
        {:error, :invalid_message}
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -52,10 +53,20 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
          {:ok, ownership_voucher} <- OwnershipVoucher.fetch(realm_name, device_id),
          {:ok, owner_private_key} <- fetch_owner_private_key(realm_name, device_id),
          {:ok, pub_key} <- OwnershipVoucher.owner_public_key(ownership_voucher),
+         {:ok, _valid_alg} <-
+           KeyExchangeStrategy.validate(hello_device.kex_name, owner_private_key),
```
noaccOS: we're not using the returned value, I suggest changing it to a simple check

```suggestion
         :ok <- KeyExchangeStrategy.validate(hello_device.kex_name, owner_private_key),
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -285,4 +302,12 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
   defp build_done2_message(setup_dv_nonce) do
     %Done2Payload{:nonce_to2_setup_dv => setup_dv_nonce} |> Done2Payload.encode()
   end
+
+  defp get_crypto_params(%COSE.Keys.ECC{alg: alg}) do
+    case alg do
+      :es256 -> {:es256, :sha256}
+      :es384 -> {:es384, :sha384}
+      _ -> {:error, :unsupported_owner_key_algorithm}
```
noaccOS: if one of the branches returns `{:error, reason}`, the other should return `{:ok, value}`, otherwise the match on line 67 will succeed and `:error` will be treated as a valid signing algorithm

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
```diff
@@ -0,0 +1,158 @@
...
+  import Astarte.Helpers.FDO
+
+  setup_all do
+    realm_name = "testrealm"
+
+    Database.setup!(realm_name)
+
+    Mimic.copy(OwnershipVoucher)
+    Mimic.copy(Astarte.Pairing.FDO.OwnerOnboarding.Session)
+    Mimic.copy(Astarte.Pairing.FDO.OwnerOnboarding.DeviceAttestation)
```
noaccOS: these should be put in `test_helper.exs`

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,179 @@ defmodule Astarte.Helpers.FDO do
...
+    # Use pkix_sign to encode and sign in one step 
+    p384_cert_der = :public_key.pkix_sign(tbs_cert, device_priv_key)
+
+    voucher = %OwnershipVoucher{
+      header: header,
+      hmac: hmac_struct,
+      entries: [decoded_entry_tag],
+      protocol_version: 101,
+      cbor_header: valid_cbor_header,
+      cbor_hmac: valid_cbor_hmac,
```
noaccOS: these two are not needed (they are used as shortcuts for certain functions), and even if you do need them they should just be `Header.encode(header)` (though this doesn't exist yet) and `Hash.encode(hmac)`, you don't need to make declarations for both. Let's avoid adding even more 🍝 

```suggestion
```

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,179 @@ defmodule Astarte.Helpers.FDO do
...
+      %CBOR.Tag{tag: :bytes, value: pub_key_der},
+      %CBOR.Tag{tag: :bytes, value: cert_chain_hash_cbor}
+    ]
+
+    valid_cbor_header = CBOR.encode(header_cbor_structure)
+
+    # HMAC
+    random_hmac_bytes = :crypto.strong_rand_bytes(48)
+    hmac_struct = %Hash{type: :hmac_sha384, hash: random_hmac_bytes}
+    valid_cbor_hmac = CBOR.encode([-6, %CBOR.Tag{tag: :bytes, value: random_hmac_bytes}])
```
noaccOS: it's not important as this should be removed anyway, but this is incorrect: random_hmac_bytes is not a binary so tagging as bytes is incorrect

the proper way would be `valid_cbor_hmac = Hash.encode_cbor(hmac_struct)`

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,179 @@ defmodule Astarte.Helpers.FDO do
...
+       {:SignatureAlgorithm, {1, 2, 840, 10045, 4, 3, 3}, :asn1_NOVALUE}, subject, validity,
+       subject, spki, :asn1_NOVALUE, :asn1_NOVALUE, :asn1_NOVALUE}
+
+    # Use pkix_sign to encode and sign in one step 
+    p384_cert_der = :public_key.pkix_sign(tbs_cert, device_priv_key)
+
+    voucher = %OwnershipVoucher{
+      header: header,
+      hmac: hmac_struct,
+      entries: [decoded_entry_tag],
```
noaccOS: this is supposed to be a binary not a cbor tag 18?

> i.e., for OVDevCertChain = X5CHAIN = [ bstr[cert1] ... bstr[certN] ],
OVDevCertChainHash = Hash[cert1 || ... || certN ]

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,179 @@ defmodule Astarte.Helpers.FDO do
...
+       {:PublicKeyAlgorithm, {1, 2, 840, 10045, 2, 1}, {:namedCurve, oid_p384}},
+       {:ECPoint, device_pub_key_point}}
+
+    tbs_cert =
+      {:OTPTBSCertificate, :v3, 123_456,
+       {:SignatureAlgorithm, {1, 2, 840, 10045, 4, 3, 3}, :asn1_NOVALUE}, subject, validity,
+       subject, spki, :asn1_NOVALUE, :asn1_NOVALUE, :asn1_NOVALUE}
+
+    # Use pkix_sign to encode and sign in one step 
+    p384_cert_der = :public_key.pkix_sign(tbs_cert, device_priv_key)
```
noaccOS: Let's _divide et impera_, let's put this in its own private function to avoid cluttering

```elixir
p384_cert_der = gen_cert(device_priv_key)
```

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,179 @@ defmodule Astarte.Helpers.FDO do
...
+    hmac_struct = %Hash{type: :hmac_sha384, hash: random_hmac_bytes}
+    valid_cbor_hmac = CBOR.encode([-6, %CBOR.Tag{tag: :bytes, value: random_hmac_bytes}])
+
+    # Entry with Signature
+    protected_header_bin = CBOR.encode(%{1 => -35})
+    unprotected_header_map = %{}
+    sig_structure = ["Signature1", protected_header_bin, <<>>, entry_payload]
+
+    der_signature = :public_key.sign(CBOR.encode(sig_structure), :sha384, priv_key_record)
+    raw_signature = der_to_raw_p384(der_signature)
```
noaccOS: we have sign1 messages support in the cose library, let's reuse existing functions instead of doing all the logic from the ground up every time

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,179 @@ defmodule Astarte.Helpers.FDO do
...
+    pem = :public_key.pem_encode([pem_entry])
+
+    # Encode Owner Public Key to X.509 DER for Header
+    pub_key_asn1 = {{:ECPoint, pub_key_raw_point}, {:namedCurve, oid_p384}}
+
+    {:SubjectPublicKeyInfo, pub_key_der, _} =
+      :public_key.pem_entry_encode(:SubjectPublicKeyInfo, pub_key_asn1)
+
+    # Prepare Entry Payload (COSE Key)
+    <<0x04, x_coord::binary-size(48), y_coord::binary-size(48)>> = pub_key_raw_point
```
noaccOS: let's use COSE.Keys instead

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,179 @@ defmodule Astarte.Helpers.FDO do
...
+      -3 => y_coord
+    }
+
+    entry_payload = CBOR.encode(cose_key_map)
+
+    # Create Header
+    guid_raw = UUID.uuid4(:raw)
+
+    cert_chain_hash_bytes = :crypto.strong_rand_bytes(48)
+    cert_chain_hash_struct = %Hash{type: :sha384, hash: cert_chain_hash_bytes}
```
noaccOS: this should be the hash of the certificate chain: once we have the cert_chain, we can

```elixir
cert_chain_hash = Hash.new(:sha256, Enum.join(cert_chain))
```

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
```diff
@@ -0,0 +1,158 @@
...
+  describe "hello_device/2" do
+    test "P-256 Flow: negotiates ECDH256 and ES256", %{realm: realm, p256: ctx} do
+      Mimic.stub(OwnershipVoucher, :fetch, fn ^realm, id ->
+        assert id == ctx.id
+        {:ok, ctx.voucher}
+      end)
+
+      Mimic.stub(OwnershipVoucher, :owner_public_key, fn _ ->
+        {:ok, [10, 1, %CBOR.Tag{tag: :bytes, value: <<1, 2, 3>>}]}
+      end)
```
noaccOS: let's have actual fetches here instead of using stubs

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    Mimic.stub(Astarte.Pairing.FDO.OwnerOnboarding.Session, :new, fn _realm, hello, _ov, _key ->
+      assert hello.kex_name == expected_alg
+      {:ok, session}
+    end)
+
+    Mimic.stub(Astarte.Pairing.FDO.OwnerOnboarding.DeviceAttestation, :eb_sig_info, fn _ ->
+      expected_sig
+    end)
```
noaccOS: why do we need these stubs?

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
```diff
@@ -0,0 +1,158 @@
...
+
+    {key, session}
+  end
+
+  defp encode_hello_device_manually(msg) do
```
noaccOS: let's add an encode function to hello_device instead

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -163,4 +165,30 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
         {:error, :invalid_message}
     end
   end
+
+  def encode_hello_device(msg) do
```
noaccOS: don't repeat the module name in the function names

`HelloDevice.encode_hello_device(hello_device)` -> `HelloDevice.encode(hello_device)`


```suggestion
  def encode(hello_device) do
```

noaccOS: `msg` as a parameter name is generic and does not give information of what the parameter should be. in this case it's a `t()` so I suggest `hello_device`

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -163,4 +165,30 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
...
+    sig_info_cbor =
+      case msg.easig_info do
+        :es384 -> [-35, %CBOR.Tag{tag: :bytes, value: <<>>}]
+        :es256 -> [-7, %CBOR.Tag{tag: :bytes, value: <<>>}]
+        val -> val
+      end
```
noaccOS: 🍝 

```suggestion
    encoded_sig_info = SignatureInfo.encode(msg.easig_info)
```

the `_cbor` suffix is also not correct as what we have is not an encoded cbor (ie the result of `CBOR.encode(x)`)

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -163,4 +165,30 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
         {:error, :invalid_message}
     end
   end
+
+  def encode_hello_device(msg) do
+    sig_info_cbor =
+      case msg.easig_info do
+        :es384 -> [-35, %CBOR.Tag{tag: :bytes, value: <<>>}]
+        :es256 -> [-7, %CBOR.Tag{tag: :bytes, value: <<>>}]
+        val -> val
```
noaccOS: this is a dangerous pattern to adopt in general, as by allowing passthrough of values it's easy to miss mistakes. these default branches should not be included unless there is strong motivation to include them in your specific instance (such as you're making generic code to be reused by other modules)

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -163,4 +165,30 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
...
+      end
+
+    cipher_id =
+      case msg.cipher_name do
+        :aes_256_gcm -> 3
+        :aes_128_gcm -> 1
+        other -> other
+      end
```
noaccOS: in our decode function we're calling the `COSE` library so I'd say we can do the same here


```suggestion
    cipher_id = COSE.algorithm(msg.cipher_name)
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
```diff
@@ -163,4 +165,30 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice do
...
+
+    [
+      msg.max_size,
+      %CBOR.Tag{tag: :bytes, value: msg.device_id},
+      %CBOR.Tag{tag: :bytes, value: msg.nonce},
+      msg.kex_name,
+      cipher_id,
+      sig_info_cbor
+    ]
+    |> CBOR.encode()
```
noaccOS: usually we split the `encode` (sometimes called `to_cbor_list`) and the `cbor_encode` functions in case we only need a step of the whole process

here I suggest having everything but the last line be the `encode` function and then having

```elixir
def cbor_encode(hello_device), do: encode(hello_device) |> CBOR.encode()
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/key_exchange_strategy.ex
```diff
@@ -0,0 +1,50 @@
...
+  Validates the Device's Key Exchange choice against the Owner's Key.
+
+  ## Parameters
+  - `device_kex_name`: The string identifying the suite chosen by the device (e.g., "ECDH256").
+  - `owner_key`: The COSE Key struct representing the Owner's private key.
+  """
+  @spec validate(String.t(), struct()) :: {:ok, String.t()} | {:error, term()}
+  def validate(device_kex_name, owner_key) do
+    case {device_kex_name, owner_key} do
+      {@ecdh256, %ECC{crv: crv}} when crv in [:p256, :secp256r1] ->
```
noaccOS: if the `crv` can be either value, there is a bug in `cose`


```suggestion
      {@ecdh256, %ECC{crv: :p256}} ->
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/key_exchange_strategy.ex
```diff
@@ -0,0 +1,50 @@
...
+        {:ok, @ecdh256}
+
+      {@ecdh384, %ECC{crv: crv}} when crv in [:p384, :secp384r1] ->
```
noaccOS:
```suggestion
      {@ecdh384, %ECC{crv: :p384}} ->
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/key_exchange_strategy.ex
```diff
@@ -0,0 +1,50 @@
...
+
+  ## Parameters
+  - `device_kex_name`: The string identifying the suite chosen by the device (e.g., "ECDH256").
+  - `owner_key`: The COSE Key struct representing the Owner's private key.
+  """
+  @spec validate(String.t(), struct()) :: {:ok, String.t()} | {:error, term()}
+  def validate(device_kex_name, owner_key) do
+    case {device_kex_name, owner_key} do
+      {@ecdh256, %ECC{crv: crv}} when crv in [:p256, :secp256r1] ->
+        {:ok, @ecdh256}
```
noaccOS: why are we returning the `kex_name` in case of success?




```suggestion
        :ok
```

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,255 @@ defmodule Astarte.Helpers.FDO do
...
+  def generate_voucher_data_and_pem(opts \\ []) do
+    curve = Keyword.get(opts, :curve, :p384)
+    encoding = Keyword.get(opts, :encoding, :x5chain)
+
+    # Curve parameters
+    {oid, hash_alg, sig_alg, fdo_alg_id, pub_key_type} = get_curve_params(curve)
+
+    # Device key
+    device_priv_key = :public_key.generate_key({:namedCurve, oid})
+    {:ECPrivateKey, _, _, _, device_pub_key_point, _} = device_priv_key
```
noaccOS: can this function be restructured to use COSE.Keys?


```suggestion
    device_priv_key = COSE.Keys.ECC.generate(curve)
```

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,255 @@ defmodule Astarte.Helpers.FDO do
...
+
+    entry_tag = %CBOR.Tag{tag: 18, value: entry_structure}
+
+    voucher = %OwnershipVoucher{
+      header: header_struct,
+      hmac: hmac_struct,
+      entries: [entry_tag],
+      protocol_version: 101,
+      cbor_header: valid_cbor_header,
+      cbor_hmac: valid_cbor_hmac_list,
```
noaccOS: not all comments from my previous review have been addressed

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
```diff
@@ -0,0 +1,139 @@
...
+      p256_chain: %{id: id_p256_chain},
+      p384_x509: %{id: id_p384_x509},
+      p384_chain: %{id: id_p384_chain}
+    }
+  end
+
+  describe "hello_device/2" do
+    test "P-256 Flow: negotiates ECDH256 and ES256", %{realm_name: realm_name, p256_x509: ctx} do
+      hello_msg = generate_hello_msg(ctx.id, "ECDH256", :es256)
+      cbor_hello = HelloDevice.encode_hello_device(hello_msg)
```
noaccOS: can we get this hello device directly from the setup? I think the actual test starts at the lines below (same for all other tests)

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
```diff
@@ -0,0 +1,139 @@
...
+    cbor_hello = HelloDevice.encode_hello_device(hello_msg)
+
+    assert {:ok, session_key, resp_binary} = OwnerOnboarding.hello_device(realm_name, cbor_hello)
+
+    assert is_binary(session_key)
+    assert {:ok, sign1_msg} = Sign1.decode_cbor(resp_binary)
+    assert sign1_msg.phdr.alg == :es384
+  end
+
+  defp encode_voucher_to_cbor(voucher) do
```
noaccOS: let's add an `encode` function to the ownershipvoucher module instead

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
```diff
@@ -0,0 +1,139 @@
...
+    CBOR.encode(cbor_list)
+  end
+
+  defp generate_hello_msg(device_id, kex_name, sig_info) do
+    HelloDevice.generate(
+      device_id: device_id,
+      kex_name: kex_name,
+      easig_info: sig_info
+    )
+  end
```
noaccOS: Is this wrapper actually adding anything over the base `HelloDevice.generate` function? I think calling the base function directly would be more explicit without loss in readability

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,255 @@ defmodule Astarte.Helpers.FDO do
...
+      3 => cose_alg,
+      -2 => x,
+      -3 => y
+    }
+
+    entry_list = [<<>>, <<>>, <<>>, cose_key_map]
+    CBOR.encode(entry_list)
+  end
+
+  defp encode_pubkey_for_cbor(type, encoding, body) do
```
noaccOS: How is this different to `PublicKey.encode`?

noaccOS (changes requested):

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +171,255 @@ defmodule Astarte.Helpers.FDO do
...
+  def generate_p256_x509_data_and_pem do
+    generate_voucher_data_and_pem(curve: :p256, encoding: :x509)
+  end
+
+  # Generic generator for all suported curve e encoding
+  # Options:
+  #   :curve -> :p256 o :p384 (default :p384)
+  #   :encoding -> :x509 o :x5chain (default :x5chain)
+  def generate_voucher_data_and_pem(opts \\ []) do
+    curve = Keyword.get(opts, :curve, :p384)
```
frossq: Probably better to use p256 as default to be aligned with all the defaults of the FDO implementation

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/key_exchange_strategy.ex
AliouneGaye21: Yes actually the curve are onlu p256, p384 ecc...

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
AliouneGaye21: Thanks for the clarification. while the spec mandates the signature algorithm based on the owner's key curve, the hash algorithm for `HelloDeviceHash` is tied to the negotiated Cipher Suite (KEX) rather than the signing key itself.
I have refactored the logic to decouple these two operations.

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/session_test.exs
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/test/support/helpers/fdo.ex
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
noaccOS: As per the table in 3.6 I would say eg aes_128_gcm should have output length 128 instead, I don't think the 4.4 table is relevant here

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
noaccOS: thats-a-nice-argument-senator-why-dont-you-back-it-up-with-a-source

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/key_exchange_strategy.ex
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding/owner_onboarding_test.exs
AliouneGaye21: Now we are colling directly the generate from hello_decice. I have aaddesd this funzction because of the optional parameters. 

- [ ] Adressed

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/hello_device.ex
AliouneGaye21: - [ ] Adressed

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -169,4 +172,229 @@ defmodule Astarte.Helpers.FDO do
...
+    guid_raw = UUID.uuid4(:raw)
+
+    chain_data_to_hash =
+      case pub_key_body do
+        list when is_list(list) -> Enum.join(list)
+        bin -> bin
+      end
+
+    cert_chain_hash_bytes = :crypto.hash(hash_alg, chain_data_to_hash)
+    cert_chain_hash_struct = Hash.new(hash_alg, cert_chain_hash_bytes)
```
noaccOS:
```suggestion
    cert_chain_hash_struct = Hash.new(hash_alg, chain_data_to_hash)
```

`Hash.new` already performs the hashing

noaccOS (changes requested):
We'll revisit the ownership voucher generation later, for now other my comment it's good enough

Good job!

PR #1727, opened 2025-12-17
feat(pairing): add support for DHKEX shared secret derivation algorithms in TO2

frossq:
Implement DHKEX shared key derivation algorithms for TO2 session.

Add tests for TO2 shared secret derivation procedures; covered algorithms:
- ECDH256
- ECDH384
- DHKEX group 14
- DHKEX group 15

* [X] I have added or ran the appropriate tests

Fixes #

Depends on cose PR https://github.com/secomind/cose-elixir/pull/14 (to be merged)

* [ ] Yes
* [X] No

```docs

```


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
```diff
@@ -21,6 +21,38 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SessionKey do
...
+                   "EF9519B3CD3A431B302B0A6DF25F14374FE1356D6D51C245" <>
+                   "E485B576625E7EC6F44C42E9A637ED6B0BFF5CB6F406B7ED" <>
+                   "EE386BFB5A899FA5AE9F24117C4B1FE649286651ECE45B3D" <>
+                   "C2007CB8A163BF0598DA48361C55D39A69163FA8FD24CF5F" <>
+                   "83655D23DCA3AD961C62F356208552BB9ED529077096966D" <>
+                   "670C354E4ABC9804F1746C08CA18217C32905E462E36CE3B" <>
+                   "E39E772C180E86039B2783A2EC07A28FB5C55DF06F4C52C9" <>
+                   "DE2BCBF6955817183995497CEA956AE515D2261898FA0510" <>
+                   "15728E5A8AACAA68FFFFFFFFFFFFFFFF")
+                |> String.to_integer(16)
```
Annopaolo: Nit (or maybe, even better, just the integer)
```suggestion
  @dhkex_p_2048 """
  FFFFFFFFFFFFFFFFC90FDAA22168C234C4C6628B80DC1CD1
  29024E088A67CC74020BBEA63B139B22514A08798E3404DD
  EF9519B3CD3A431B302B0A6DF25F14374FE1356D6D51C245
  E485B576625E7EC6F44C42E9A637ED6B0BFF5CB6F406B7ED
  EE386BFB5A899FA5AE9F24117C4B1FE649286651ECE45B3D
  C2007CB8A163BF0598DA48361C55D39A69163FA8FD24CF5F
  83655D23DCA3AD961C62F356208552BB9ED529077096966D
  670C354E4ABC9804F1746C08CA18217C32905E462E36CE3B
  E39E772C180E86039B2783A2EC07A28FB5C55DF06F4C52C9
  DE2BCBF6955817183995497CEA956AE515D2261898FA0510
  15728E5A8AACAA68FFFFFFFFFFFFFFFF
  """
  |>  String.to_integer(16)              
```

Annopaolo (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
```diff
@@ -178,6 +250,22 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.SessionKey do
...
+  def get_dhkex_p(dhkex_group) do
+    case dhkex_group do
+      14 ->
+        @dhkex_p_2048
+
+      15 ->
+        @dhkex_p_3072
+
+      _ ->
+        :unsupported
```
noaccOS: nit: as we're using this in tests, I suggest being explicit

```suggestion
        raise "unsupported dhkex #{inspect(dhkex_group)}"
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
noaccOS: as an alternative, as you can't pipe on module attributes without wrapping the experssion in parentheses, you may also

```elixir
@dhkex_p_2048_str """
FFFF
....
15728
"""

@dhkex_p_2048 String.to_integer(@dhkex_p_2048_str, 16)
```

apps/astarte_pairing/test/support/cases/fdo_session.ex
```diff
@@ -34,6 +34,7 @@ defmodule Astarte.Cases.FDOSession do
   alias Astarte.Pairing.FDO.OwnerOnboarding.HelloDevice
   alias Astarte.Pairing.FDO.OwnerOnboarding.Session
   alias Astarte.Pairing.FDO.OwnerOnboarding.SessionKey
+  alias COSE.Keys.ECC, as: ECC
```
noaccOS: nit

```suggestion
  alias COSE.Keys.ECC
```

apps/astarte_pairing/test/support/cases/fdo_session.ex
```diff
@@ -85,6 +86,123 @@ defmodule Astarte.Cases.FDOSession do
...
+        owner_key = ECC.generate(:es256)
+        device_key = ECC.generate(:es256)
+        {:ok, owner_rand, xa} = SessionKey.new("ECDH256", owner_key)
+        {:ok, device_rand, xb} = SessionKey.new("ECDH256", device_key)
+
+        custom_session =
+          session_fixture(original_session, %{
+            kex_suite_name: "ECDH256",
+            owner_random: owner_rand
+          })
```
noaccOS: this is a bad pattern because you take the original struct, which is built in a way which ensures the built session is consistent with itself and with the hello device message, and you're cherry picking fields

the result is not guaranteed to be a valid struct

a better alternative would be making use of the kex_name parameter in the original hello_device build, so that we know everything else follows and is consistent. we also should not need to manually re-generate the owner random this way (if the protocol works correctly)

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_key.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/cases/fdo_session.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/cases/fdo_session.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/cases/fdo_session.ex
```diff
@@ -85,6 +86,137 @@ defmodule Astarte.Cases.FDOSession do
...
+        %{}
+
+      "ECDH256" ->
+        owner_key = ECC.generate(:es256)
+        device_key = ECC.generate(:es256)
+        {:ok, device_rand, xb} = SessionKey.new("ECDH256", device_key)
+
+        # generate a new consistent session starting from a HelloDevice msg requesting ECDH256
+        hello_device_ecdh256 =
+          HelloDevice.generate(kex_name: "ECDH256")
```
noaccOS: can't we just change the hello device on line 50? why do we need to re-declare it?

noaccOS (changes requested):

apps/astarte_pairing/test/support/cases/fdo_session.ex
frossq: We can :+1:  I just recreated it for (my own) clarity

frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/cases/fdo_session.ex
```diff
@@ -44,47 +49,176 @@ defmodule Astarte.Cases.FDOSession do
...
 
-  setup_all %{realm_name: realm_name} do
-    device_id = sample_device_guid()
-    hello_device = %{HelloDevice.generate() | device_id: device_id}
-    ownership_voucher = sample_ownership_voucher()
-    owner_key = sample_extracted_private_key()
+  @allowed_owner_key_tag_values ["EC256", "EC384", "RSA2048", "RSA3072"]
+  @allowed_kex_name_tag_values ["ECDH256", "ECDH384", "DHKEXid14", "DHKEXid15"]
+
+  setup %{realm_name: realm_name, astarte_instance_id: astarte_instance_id} = context do
```
noaccOS: nit: prefer extracting from variables in the bodies instead of the definitions. this applies generally to functions but can be broadly applied to cases such as this `setup` block

apps/astarte_pairing/test/support/cases/fdo_session.ex
```diff
@@ -44,47 +49,176 @@ defmodule Astarte.Cases.FDOSession do
...
+      case key_type do
+        kt when is_nil(kt) or kt == "EC256" ->
+          # no key modifier applied: use the default context (EC256 owner/device key + ECDH256 KEX)
+          default_fdo_context_map()
+
+        custom_key_type ->
+          custom_fdo_context_modifier(custom_key_type, kex_name)
+      end
+
+    context = Map.merge(context, fdo_context)
```
noaccOS: if we need to merge two contexts, it can probably be made cleaner by having two subsequent setup blocks

apps/astarte_pairing/test/support/cases/fdo_session.ex
```diff
@@ -44,47 +49,176 @@ defmodule Astarte.Cases.FDOSession do
...
-    hello_device = %{HelloDevice.generate() | device_id: device_id}
-    ownership_voucher = sample_ownership_voucher()
-    owner_key = sample_extracted_private_key()
+  @allowed_owner_key_tag_values ["EC256", "EC384", "RSA2048", "RSA3072"]
+  @allowed_kex_name_tag_values ["ECDH256", "ECDH384", "DHKEXid14", "DHKEXid15"]
+
+  setup %{realm_name: realm_name, astarte_instance_id: astarte_instance_id} = context do
+    # a different test setup may be required to use non-default keys and algorithms
+    # insert here the condition to select a custom setup path and the related logic
+    key_type = Map.get(context, :owner_key)
```
noaccOS: if we're treating something as the "default", use that deafult instead of `nil`


```suggestion
    key_type = Map.get(context, :owner_key, "EC256")
```

this way you have 1 less possible value to handle

apps/astarte_pairing/test/support/cases/fdo_session.ex
```diff
@@ -44,47 +49,176 @@ defmodule Astarte.Cases.FDOSession do
...
+
+    {:ok, session} =
+      Session.build_session_secret(session, realm_name, context.owner_key, context.xb)
+
+    {:ok, session} = Session.derive_key(session, realm_name)
+
+    on_exit(fn ->
+      setup_database_access(astarte_instance_id)
+      delete_session(realm_name, session.key)
+    end)
```
noaccOS, line 122: let's put this just below the `Session.new` so that if we have a crash in `build_session_secret` we still clean up on exit

apps/astarte_pairing/test/support/cases/fdo_session.ex
```diff
@@ -44,47 +49,176 @@ defmodule Astarte.Cases.FDOSession do
...
+
+    Map.merge(context, %{session: session})
```
noaccOS: each setup block should return the fields it has changed, this is not obvious by looking at this line as we have a context which is already a merge of the original context and new entries plus this session

if we have two distinct setup this last one can just return `%{session: session}`

apps/astarte_pairing/test/support/cases/fdo_session.ex
```diff
@@ -44,47 +49,176 @@ defmodule Astarte.Cases.FDOSession do
...
-  setup context do
-    %{
-      astarte_instance_id: astarte_instance_id,
-      hello_device: hello_device,
-      ownership_voucher: ownership_voucher,
-      realm_name: realm_name,
-      owner_key: owner_key,
-      xb: xb
-    } = context
+  defp custom_fdo_context_modifier(key_type, kex_name) do
```
noaccOS: I'm not sure this is the best place to split the logic
For one, I don't see why the `default_fdo_context` is separate, as we're still performing further specialization below
Then, all paths look exactly the same (except for the key type and other details), would it not make sense to have generic functions just for specific instructions?

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -174,36 +174,49 @@ defmodule Astarte.Helpers.FDO do
...
-  #   :curve -> :p256 o :p384 (default :p256)
-  #   :encoding -> :x509 o :x5chain (default :x509)
+  #   :curve -> :p256 or :p384 (default :p256)
+  #   :encoding -> :x509 or :x5chain (default :x509)
+  #   :device_key -> an EC256 or EC384 device key used for attestation inside the voucher;
+  #                  if empty, a new key will be generated on the fly according to the curve type
   def generate_voucher_data_and_pem(opts \\ []) do
     curve = Keyword.get(opts, :curve, :p256)
     encoding = Keyword.get(opts, :encoding, :x509)
+    device_priv_key = Keyword.get(opts, :device_key, nil)
```
noaccOS: nit, will need to be rewritten using generators in the long run

```suggestion
    device_priv_key = Keyword.get_lazy(opts, :device_key, fn -> ECC.generate(cose_alg) end)
```

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -174,36 +174,49 @@ defmodule Astarte.Helpers.FDO do
...
+      case device_priv_key do
+        nil ->
+          ECC.generate(cose_alg)
+
+        %ECC{alg: ^cose_alg} ->
+          device_priv_key
+
+        _ ->
+          raise "mismatch between device key type #{device_priv_key.alg} and curve #{curve}"
+      end
```
noaccOS: nit, a bit verbose

```suggestion
    %ECC{alg: ^cose_alg} = device_priv_key
```

apps/astarte_pairing/test/support/cases/fdo_session.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/cases/fdo_session.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/cases/fdo_session.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/cases/fdo_session.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/cases/fdo_session.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/cases/fdo_session.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/helpers/fdo.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/helpers/fdo.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/support/helpers/fdo.ex
```diff
@@ -174,26 +174,28 @@ defmodule Astarte.Helpers.FDO do
     |> Repo.insert(prefix: Realm.keyspace_name(realm_name))
   end
 
-  def generate_p384_x5chain_data_and_pem do
+  def generate_p384_x5chain_data_and_pem() do
```
noaccOS, line 177: credo will hate you for this

noaccOS (approved):

apps/astarte_pairing/test/support/helpers/fdo.ex
frossq: :pray:

PR #1728, opened 2025-12-18
chore(fdo): the owner public key is encoded only when sent to the device

mizzet1:
With this PR the owner public key is handled internally with its decoded struct and encoded only when is sent to the device.


frossq (approved):
:+1:

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -50,6 +50,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
          {:ok, ownership_voucher} <- OwnershipVoucher.fetch(realm_name, device_id),
          {:ok, owner_private_key} <- fetch_owner_private_key(realm_name, device_id),
          {:ok, pub_key} <- OwnershipVoucher.owner_public_key(ownership_voucher),
+         encoded_pub_key = PublicKey.encode(pub_key),
```
noaccOS: this can be moved inside the body of the `with` block, in order to limit the conditions (ie: the failure points) to the bare minimum

PR #1731, opened 2025-12-22
feat(rm): trigger installation and deletion notification

noaccOS:
realm management -> realm management notifications
depends on #1729


apps/astarte_realm_management/test/support/helpers/triggers.ex
```diff
@@ -24,6 +24,14 @@ defmodule Astarte.Helpers.Triggers do
     Cache.reset_realm_cache(realm_name)
   end
 
+  def rpc_trigger_client do
+    Astarte.RealmManagement.Supervisor
+    |> Supervisor.which_children()
+    |> Enum.find_value(fn {id, pid, _type, _mod} ->
+      id == Astarte.RPC.Triggers.Client && pid
```
Annopaolo, line 31: Imho it's better to make this comparison more explicit:
```suggestion
      id == Astarte.RPC.Triggers.Client and pid != nil
```

Annopaolo (approved):

apps/astarte_realm_management/test/support/helpers/triggers.ex
noaccOS: Your version is not equivalent, as the function it's being used in is `find_value`: we're returning the `pid` for the child with id `Astarte.RPC.Triggers.Client`

AliouneGaye21 (approved):

PR #1733, opened 2025-12-22
chore(fdo): replacing stubs with real calls on controller_test

mizzet1:
With this PR the stubs in fdo_onboarding_controller_test have been replaced with real calls, using a mock session. (in particular the Session.decrypt_and_verify and encrypt_and_verify functions).

Asserts that returned error 500 by default where also changed to return the correct FDO error code 100 (message_body_error) where needed.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
@@ -36,35 +36,36 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
         extra_payload_claims \\ %{},
         extra_unprotected_header_claims \\ %{}
       ) do
-    # COSE.Messages.Sign1.verify_decode returns :error for both decode failures and verification failures
-    # This extra check distinguish between error 100 and 101 according to FDO Spec
+    with {:ok, decoded_eatoken} <- decode_eatoken(eatoken_cbor),
+         {:ok, :verified} <- verify_signature(decoded_eatoken, device_public_key),
+         {:ok, payload} <- verify_payload(decoded_eatoken.payload, extra_payload_claims) do
+      unprotected_headers =
+        translate_unprotected_headers(
+          decoded_eatoken.uhdr,
+          extra_unprotected_header_claims
+        )
+
+      {:ok, %{decoded_eatoken | payload: payload, uhdr: unprotected_headers}}
+    end
+  end
+
+  defp decode_eatoken(eatoken_cbor) do
     case Sign1.decode_cbor(eatoken_cbor) do
-      {:ok, decoded} ->
-        # CBOR structure is valid, now verify the signature
-        if Sign1.verify(decoded, device_public_key) do
-          with {:ok, payload} <- verify_payload(decoded.payload, extra_payload_claims) do
-            unprotected_headers =
-              translate_unprotected_headers(
-                decoded.uhdr,
-                extra_unprotected_header_claims
-              )
-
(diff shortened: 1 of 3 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -36,14 +36,31 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
...
-      unprotected_headers =
-        translate_unprotected_headers(eatoken_message.uhdr, extra_unprotected_header_claims)
+    # COSE.Messages.Sign1.verify_decode returns :error for both decode failures and verification failures
+    # This extra check distinguish between error 100 and 101 according to FDO Spec
+    case Sign1.decode_cbor(eatoken_cbor) do
+      {:ok, _decoded} ->
+        # CBOR structure is valid
+        case Sign1.verify_decode(eatoken_cbor, device_public_key) do
+          {:ok, eatoken_message} ->
+            with {:ok, payload} <- verify_payload(eatoken_message.payload, extra_payload_claims) do
```
noaccOS: you don't need to decode it twice

```suggestion
      {:ok, eatoken_message} ->
        # CBOR structure is valid
        with true <- Sign1.verify(eatoken_message, device_public_key),
             {:ok, payload} <- verify_payload(eatoken_message.payload, extra_payload_claims) do
```

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -64,6 +61,50 @@ defmodule Astarte.PairingWeb.FDOOnboardingControllerTest do
...
+    on_exit(fn ->
+      setup_database_access(astarte_instance_id)
+      Astarte.Helpers.Database.delete_session(realm_name, session.key)
+    end)
+
+    {:ok, session} = Session.build_session_secret(session, realm_name, owner_key, xb)
+    {:ok, session} = Session.derive_key(session, realm_name)
+
+    %{session: session}
+  end
```
noaccOS: this is the third test file with the same exact setup
while this is not your fault, could you please add a new `test/support/cases/fdo_session.ex` which runs this setup? then from the test files, it would be a matter of `use Astarte.Pairing.Cases.FDOSession`

frossq (approved):
:fire:

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -36,14 +36,29 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
...
-         {:ok, payload} <- verify_payload(eatoken_message.payload, extra_payload_claims) do
-      unprotected_headers =
-        translate_unprotected_headers(eatoken_message.uhdr, extra_unprotected_header_claims)
+    # COSE.Messages.Sign1.verify_decode returns :error for both decode failures and verification failures
+    # This extra check distinguish between error 100 and 101 according to FDO Spec
+    case Sign1.decode_cbor(eatoken_cbor) do
+      {:ok, decoded} ->
+        # CBOR structure is valid, now verify the signature
+        if Sign1.verify(decoded, device_public_key) do
+          with {:ok, payload} <- verify_payload(decoded.payload, extra_payload_claims) do
```
Annopaolo: `with` inside `if` inside `case` (or any other deeply needed code structure) make it difficult to follow the happy path. I would suggest making three `with`-friendly functions:
- one for decoding 
- one for signature verification
- one for payload verification 
Each function would handle its specific error path, then use them all in a single `with` block here

PR #1738, opened 2025-12-23
chore(fdo): device should be deleted if fdo process didn’t complete successfully

mizzet1:
**Problem** 

In the current implementation, the device is permanently inserted into the database when the server processes Message 68 (DeviceServiceInfo).

However, the onboarding process is not actually finished until the device receives Message 70 (Done). If a failure occurs in the small window between Message 68 and Message 70, we end up in an inconsistent state, where:

- The Device believes it failed and will eventually retry.
- The Database believes the Device is successfully registered.

**Solution**

Register the device with a ttl and remove the ttl only if we succesfully finish the to2 protocol.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
@@ -249,13 +249,14 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
     end
   end
 
-  def done(to2_session, body) do
+  def done(realm_name, to2_session, body) do
     # retrieve nonce NonceTO2ProveDv from session and check against incoming nonce from device
     # if match -> retrieve NonceTO2SetupDv from session and send back to device
     with {:ok, %DonePayload{nonce_to2_prove_dv: prove_dv_nonce_challenge}} <-
            DonePayload.decode(body),
          :ok <-
-           check_prove_dv_nonces_equality(prove_dv_nonce_challenge, to2_session.prove_dv_nonce) do
+           check_prove_dv_nonces_equality(prove_dv_nonce_challenge, to2_session.prove_dv_nonce),
+         {:ok, _device} <- Queries.remove_device_ttl(realm_name, to2_session.device_id) do
       done2_message = build_done2_message(to2_session.setup_dv_nonce)
       {:ok, done2_message}
     end
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
@@ -66,8 +66,12 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
       ) do
     with {:ok, session} <-
            Session.add_device_service_info(session, realm_name, service_info) do
-      encoded_device_id = generate_encoded_device_id(session.device_service_info)
-      build_and_send_owner_service_info(session, realm_name, encoded_device_id)
+      device_id = generate_device_id(session.device_service_info)
+
+      with {:ok, _session} <- Session.add_device_id(session, realm_name, device_id) do
+        encoded_device_id = Device.encode_device_id(device_id)
+        build_and_send_owner_service_info(session, realm_name, encoded_device_id)
+      end
     end
   end
 
@@ -78,15 +82,9 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
     end
   end
 
-  defp generate_encoded_device_id(device_service_info) do
-    device_service_info
-    |> generate_device_id()
-    |> Device.encode_device_id()
-  end
-
   defp build_and_send_owner_service_info(session, realm_name, encoded_device_id) do
     with {:ok, credentials_secret} <-
-           Engine.register_device(realm_name, encoded_device_id) do
+           Engine.register_device(realm_name, encoded_device_id, unconfirmed: true) do
       owner_service_info =
         OwnerServiceInfo.build(realm_name, credentials_secret, encoded_device_id)
 
@@ -132,5 +130,6 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
   end
 
   defp generate_device_id(%{{"devmod", "sn"} => %{value: sn}}), do: UUID.uuid5(:oid, sn, :raw)
+
   defp generate_device_id(_), do: Device.random_device_id()
 end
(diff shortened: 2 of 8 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -244,15 +244,34 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
...
+
+          error
+      end
+    else
+      {:error, reason} = error ->
+        Logger.error(
+          "Done message processing failed for device #{inspect(to2_session.device_id)}: #{inspect(reason)}"
+        )
+
+        error
```
noaccOS: let's try to avoid adding too many nested blocks

```suggestion
           check_prove_dv_nonces_equality(prove_dv_nonce_challenge, to2_session.prove_dv_nonce),
           {:ok, _device} <- Queries.remove_device_ttl(realm_name, to2_session.device_id) do
             done2_message = build_done2_message(to2_session.setup_dv_nonce)
             {:ok, done2_message}
           end
```

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -51,12 +51,11 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
     # TODO: make use of other device service info
     {:ok, session} = Session.add_device_service_info(session, realm_name, service_info)
 
-    encoded_device_id =
-      session.device_service_info
-      |> generate_device_id()
-      |> Device.encode_device_id()
+    device_id = session.device_id
+    encoded_device_id = Device.encode_device_id(device_id)
```
noaccOS: so wait, where are we setting the device_id now?
I think this should still be the place which generates it, as we need to make sure we've received the serial number (or that we won't receive one) before generating the device_id

I think we should change the following with block to update the session instead

```elixir
with {:ok, session} <- Session.add_device_id(session, realm_name, device_id),
     {:ok, credentials_secret} <- Engine.register_device(...)
```

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -189,6 +189,22 @@ defmodule Astarte.Pairing.Queries do
     |> Repo.update(prefix: keyspace_name, consistency: consistency)
   end
 
+  def remove_device_ttl(realm_name, device_id) do
+    keyspace_name = Realm.keyspace_name(realm_name)
+    consistency = Consistency.device_info(:write)
+
+    with {:ok, device} <- fetch_device(realm_name, device_id) do
+      device
+      |> Ecto.Changeset.change()
```
noaccOS: we can add the struct directly we don't need a changeset

```suggestion
```

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -189,6 +189,22 @@ defmodule Astarte.Pairing.Queries do
...
+      |> Repo.insert(
+        prefix: keyspace_name,
+        consistency: consistency,
+        ttl: 0,
```
noaccOS: this is the default of the device table, it can be omitted

```suggestion
```

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -189,6 +189,22 @@ defmodule Astarte.Pairing.Queries do
...
+        on_conflict: :replace_all
```
noaccOS: this has no effect for scylladb as all inserts are upserts
you may need to add `allow_stale: true` if you get some error from scylladb but you probably don't even need that option

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/done_test.exs
```diff
@@ -17,37 +17,120 @@
...
+      from d in DeviceDB,
+        where: d.device_id == ^device_id,
+        select: fragment("TTL(?)", d.first_registration)
+
+    result = Repo.one(query, prefix: keyspace)
+
+    case result do
+      nil -> {:ok, nil}
+      ttl when is_integer(ttl) -> {:ok, ttl}
+    end
```
noaccOS: this is just wrapping the result in a result tuple, which doesn't add information but adds complexity. it can just be omitted

```suggestion
    Repo.one(query, prefix: keyspace)
```

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/done_test.exs
```diff
@@ -17,37 +17,120 @@
...
+      realm: realm_name,
+      session: session
+    } do
+      done_msg = [%CBOR.Tag{tag: :bytes, value: session.prove_dv_nonce}]
+
+      {:ok, device_before} = Queries.fetch_device(realm_name, session.device_id)
+      assert device_before.device_id == session.device_id
+
+      {:ok, ttl_before} = get_device_ttl(realm_name, session.device_id)
+      assert ttl_before > 0
```
noaccOS: this check isn't complete

```elixir
iex(5)> nil > 0
true
```

noaccOS (changes requested):

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/done_test.exs
```diff
@@ -17,37 +17,115 @@
 #
 
 defmodule Astarte.Pairing.FDO.Onboarding.DoneTest do
-  use ExUnit.Case, async: true
+  use Astarte.Cases.Data, async: false
+  use Astarte.Cases.Device
```
noaccOS: is this needed if we're using the device_id from the session?

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
mizzet1: For this I reverted back to how it was before and update the session instead as suggested.

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/done_test.exs
mizzet1: I removed Astarte.Cases.Device because the tests use Queries.register_device in the setup and left Astarte.Cases.Data because is needed to retrieve the realm_name for the session.

PR #1739, opened 2025-12-24
refactor(fdo): move eatoken encode logic to eatoken module

mizzet1:
This PR  moves the custom EAToken encode logic in ProveDevice in a dedicated module.

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
@@ -17,18 +17,76 @@
 #
 
 defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
+  @moduledoc """
+  Functions for encoding and decoding Entity Attestation Tokens (EAT) used in FDO.
+
+  EAT tokens are the COSE_Sign1 objects that contain attestation claims about a device.
+  """
   alias COSE.Messages.Sign1
 
+  # UEID prefix byte as per FDO specification
+  @eat_random <<1>>
+
+  # IANA & FDO EAT Claim mappings (Appendix E & Spec 3.3.6)
   @known_payload_claims %{
+    # EAT-NONCE: Nonce claim for freshness
     10 => :nonce,
+    # UEID: Universal Entity ID claim
     256 => :ueid,
+    # EAT-FDO: FDO Claim wrapping the FDO payload
     -257 => :fdo
   }
 
   @known_unprotected_header_claims %{
-    -258 => :maroeprefix,
-    -259 => :euphnonce
+    -258 => :maroeprefix
   }
+
+  @doc """
+  Builds a UEID (Universal Entity ID) by prepending the FDO-specified prefix byte
+  to the given GUID.
+
+  """
+  def build_ueid(guid), do: @eat_random <> guid
+
+  def parse_ueid(ueid) do
+    case ueid do
+      <<@eat_random, guid::binary-size(16)>> -> {:ok, guid}
+      _ -> {:error, :invalid_ueid}
+    end
+  end
+
+  @doc """
+  Encodes and signs an EAToken with the given payload claims, unprotected header claims,
+  and private key.
+
+  """
+  def encode_sign(
+        payload_claims,
+        uhdr_claims,
+        priv_key,
+        extra_payload_claims \\ %{},
+        extra_unprotected_header_claims \\ %{}
+      ) do
+    eat_payload =
+      translate_claims_encode(
+        payload_claims,
+        @known_payload_claims,
+        extra_payload_claims
+      )
+
+    eat_uhdr =
+      translate_claims_encode(
+        uhdr_claims,
+        @known_unprotected_header_claims,
+        extra_unprotected_header_claims
+      )
+
+    eat_cbor_payload = CBOR.encode(eat_payload)
+    phdr = %{alg: :es256}
+
+    Sign1.build(eat_cbor_payload, phdr, eat_uhdr)
+    |> Sign1.sign_encode_cbor(priv_key)
+  end
 
   def verify_decode_cbor(
         eatoken_cbor,
@@ -84,6 +142,16 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
     |> Map.new(fn {claim_id, value} ->
       claim = Map.get(known_claims, claim_id, claim_id)
       {claim, value}
+    end)
+  end
+
+  defp translate_claims_encode(claims, known_claims, extra_claims) do
+    all_claims = Map.merge(known_claims, extra_claims)
+    atom_to_int = for {int, atom} <- all_claims, into: %{}, do: {atom, int}
+
+    Map.new(claims, fn {key, value} ->
+      int_key = Map.get(atom_to_int, key, key)
+      {int_key, value}
     end)
   end
 end
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/prove_device.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/prove_device.ex
@@ -33,22 +33,14 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.ProveDevice do
   # xB keys generation
   alias Astarte.Pairing.FDO.OwnerOnboarding.SessionKey
   alias COSE.Keys.ECC
-  alias COSE.Messages.Sign1
 
-  # Alias per la libreria COSE (da adattare in base alla libreria effettiva in uso su Astarte)
+  # EAToken encoding and decoding
   alias Astarte.Pairing.FDO.OwnerOnboarding.EAToken
 
-  # --- IANA & FDO Constants (Appendix E & Spec 3.3.6) ---
-
-  # EAT-FDO: FDO-specific claim wrapping the FDO payload (Label -257) [cite: 3456]
-  @eat_fdo_label -257
-  @eat_nonce_label 10
-  @eat_ueid_label 256
-  @euph_nonce -259
-
-  @eat_random <<1>>
-
   @nonce_binary_size 16
+
+  # EUPHNonce: Nonce in unprotected headers (Label -259)
+  @euph_nonce -259
 
   typedstruct enforce: true do
     @typedoc "Decoded content of the TO2.ProveDevice message."
@@ -79,17 +71,22 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.ProveDevice do
     field :raw_eat_token, binary()
   end
 
+  def euph_nonce_claim_key, do: @euph_nonce
+
   @doc """
   Decodes the TO2.ProveDevice message, verifies the COSE signature using the
   Device Public Key, and extracts the required FDO fields.
   """
   @spec decode(binary(), any()) :: {:ok, t()} | {:error, atom()}
   def decode(binary_msg, device_pub_key) do
-    with {:ok, cose_object} <- EAToken.verify_decode_cbor(binary_msg, device_pub_key),
+    extra_uhdr_claims = %{euph_nonce_claim_key() => :euph_nonce}
+
+    with {:ok, cose_object} <-
+           EAToken.verify_decode_cbor(binary_msg, device_pub_key, %{}, extra_uhdr_claims),
          {:ok, xb_key} <- extract_fdo_payload(cose_object.payload),
          {:ok, nonce_prove} <- fetch_binary(cose_object.payload, :nonce),
          :ok <- check_expected_binary_size(nonce_prove, @nonce_binary_size),
-         {:ok, nonce_setup} <- fetch_binary(cose_object.uhdr, :euphnonce),
+         {:ok, nonce_setup} <- fetch_binary(cose_object.uhdr, :euph_nonce),
          :ok <- check_expected_binary_size(nonce_setup, @nonce_binary_size),
          {:ok, ueid} <- fetch_binary(cose_object.payload, :ueid),
          {:ok, guid} <- guid_from_ueid(ueid) do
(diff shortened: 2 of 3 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -17,19 +17,62 @@
...
+  def eat_random, do: @eat_random
+
+  @doc """
+  Encodes and signs an EAToken with the given payload claims, unprotected header claims,
+  and private key.
+
+  """
+  def encode_sign(payload_claims, uhdr_claims, priv_key) do
+    eat_cbor_payload = payload_claims |> CBOR.encode()
+    phdr = %{alg: :es256}
```
frossq, line 85: Consider that we are adding support for other types of algorithms (es384, rs256 et al.), better not to hardcode it and instead pass it as parameter. ES256 may be left as default

mizzet1: The scope of this pr was only to refractor the current codebase, but that's a good change to do also in my opinion

frossq (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -17,19 +17,62 @@
...
+
+  def eat_random, do: @eat_random
+
+  @doc """
+  Encodes and signs an EAToken with the given payload claims, unprotected header claims,
+  and private key.
+
+  """
+  def encode_sign(payload_claims, uhdr_claims, priv_key) do
+    eat_cbor_payload = payload_claims |> CBOR.encode()
```
noaccOS: claims need to be translated from atoms to integers

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -17,19 +17,62 @@
...
+    phdr = %{alg: :es256}
+
+    Sign1.build(eat_cbor_payload, phdr, uhdr_claims)
```
noaccOS: same for the uhdr

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/prove_device.ex
```diff
@@ -120,16 +109,18 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.ProveDevice do
...
-    eat_cbor_payload =
-      prove_device_payload
-      |> prove_device_payload_to_cbor()
-
-    phdr = %{alg: :es256}
+    payload_claims = %{
+      EAToken.eat_fdo_label() => [prove_device_payload.xb_key_exchange |> COSE.tag_as_byte()],
+      EAToken.eat_nonce_label() => prove_device_payload.nonce_to2_prove_dv |> COSE.tag_as_byte(),
+      EAToken.eat_ueid_label() =>
+        (EAToken.eat_random() <> prove_device_payload.guid) |> COSE.tag_as_byte()
```
noaccOS:
```suggestion
      fdo: [prove_device_payload.xb_key_exchange |> COSE.tag_as_byte()],
      nonce: prove_device_payload.nonce_to2_prove_dv |> COSE.tag_as_byte(),
      ueid:
        (EAToken.eat_random() <> prove_device_payload.guid) |> COSE.tag_as_byte()

File: apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/prove_device.ex [hunk 5]
```diff
@@ -120,16 +109,18 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.ProveDevice do
...
+    }
 
-    uhdr = %{@euph_nonce => prove_device_payload.nonce_to2_setup_dv |> COSE.tag_as_byte()}
+    uhdr_claims = %{
+      EAToken.euph_nonce_label() => prove_device_payload.nonce_to2_setup_dv |> COSE.tag_as_byte()
```
noaccOS:
```suggestion
      euph_nonce: prove_device_payload.nonce_to2_setup_dv |> COSE.tag_as_byte()
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -17,19 +17,62 @@
...
-    -257 => :fdo
+    @eat_nonce_label => :nonce,
+    @eat_ueid_label => :ueid,
+    @eat_fdo_label => :fdo
   }
 
   @known_unprotected_header_claims %{
     -258 => :maroeprefix,
-    -259 => :euphnonce
+    @euph_nonce => :euphnonce
```
noaccOS: I would actually not add it here personally as it's not part of `$EATUnprotectedHeaders` in the spec
I would suggest using the extra claims in the function call

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -129,4 +146,14 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
       {claim, value}
     end)
   end
+
+  defp translate_claims_encode(claims, known_claims, extra_claims) do
+    all_claims = Map.merge(known_claims, extra_claims)
+    atom_to_int = for {int, atom} <- all_claims, into: %{}, do: {atom, int}
+
+    Map.new(claims, fn {key, value} ->
+      int_key = if is_atom(key), do: Map.get(atom_to_int, key, key), else: key
```
noaccOS: the default for `Map.get` already matches the behavior to select the key itself if it was not found in the map, the `if` can be omitted

```suggestion
      int_key = Map.get(atom_to_int, key, key)
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/prove_device.ex
```diff
@@ -74,7 +74,10 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.ProveDevice do
   """
   @spec decode(binary(), any()) :: {:ok, t()} | {:error, atom()}
   def decode(binary_msg, device_pub_key) do
-    with {:ok, cose_object} <- EAToken.verify_decode_cbor(binary_msg, device_pub_key),
+    extra_uhdr_claims = %{EAToken.euph_nonce_label() => :euphnonce}
```
noaccOS: as the euph_nonce header is specific for prove_device and not eatoken, let's move the definition of `euph_nonce_label` here
(also, why are they called labels if they are the integer values?)

noaccOS: here the header is translated to `:euphnonce` but then the encode function expects `:euph_nonce`, they should be the same to allow `x == decode(encode(x))`
I would suggest `euph_nonce`

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -65,23 +64,41 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
...
 
-    Sign1.build(eat_cbor_payload, phdr, uhdr_claims)
+    Sign1.build(eat_cbor_payload, phdr, eat_uhdr)
     |> Sign1.sign_encode_cbor(priv_key)
   end
 
   def verify_decode_cbor(
         eatoken_cbor,
         device_public_key,
-        extra_payload_claims \\ %{},
```
noaccOS, line 36: why was this removed?

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/prove_device.ex
noaccOS: Actually, let's add a test now that we have symmetric encode/decode functions

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/prove_device_test.exs
```diff
@@ -92,13 +88,13 @@ defmodule Astarte.Pairing.OwnerOnboarding.Onboarding.ProveDevice do
...
        ) do
-    ueid = <<1>> <> guid_val
+    ueid = EAToken.eat_random() <> guid_val
     ueid = COSE.tag_as_byte(ueid)
 
     eat_claims = %{
-      10 => COSE.tag_as_byte(prove_dv_nonce_val),
-      256 => ueid,
-      -257 => [COSE.tag_as_byte(xb)]
+      EAToken.eat_nonce_label() => COSE.tag_as_byte(prove_dv_nonce_val),
```
noaccOS: now that we have a proper encode function, `build_test_cose_sign1` and `encode_sign_missing_cbor_tags` can use that instead
that way we also don't need to re-export all the `eat_*_label` functions

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/prove_device.ex
```diff
@@ -159,18 +155,11 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.ProveDevice do
   end
 
   defp guid_from_ueid(ueid) do
+    eat_random = EAToken.eat_random()
+
     case ueid do
-      <<@eat_random::binary, guid::binary-size(16)>> -> {:ok, guid}
+      <<^eat_random::binary-size(1), guid::binary-size(16)>> -> {:ok, guid}
```
noaccOS: you justifiably moved the eat_random to the eatoken module, as it's where it's defined
however, the whole ueid is also defined there, so I would suggest adding this function to the eatoken module directly

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/prove_device.ex
mizzet1: I saw that Iana use label for COSE [spec](https://www.iana.org/assignments/cose/cose.xhtml#:~:text=Name-,Label,-Value%20Type) but for these claims it uses [claim_key](https://www.iana.org/assignments/cwt/cwt.xhtml#:~:text=JWT%20Claim%20Name-,Claim%20Key,-Claim%20Value%20Type), so I will change label with claim_key

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
mizzet1: extra_payload_claims was passed inside verify_payload, which now uses the added @known_payload_claims, which includes all the payload claims.

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -17,28 +17,90 @@
...
   alias COSE.Messages.Sign1
 
+  # --- IANA & FDO EAT Constants (Appendix E & Spec 3.3.6) ---
+
+  # EAT-FDO: FDO Claim wrapping the FDO payload (Claim Key -257)
+  @eat_fdo_claim_key -257
+  # EAT-NONCE: Nonce claim for freshness (Claim Key 10)
+  @eat_nonce_claim_key 10
+  # UEID: Universal Entity ID claim (Claim Key 256)
+  @eat_ueid_claim_key 256
```
noaccOS: do we still need these?

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
noaccOS: I'd still leave the option to add extra payload claims (even though we're not using it in our implementation), as this is supposed to be a generic module

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -17,28 +17,90 @@
 #
 
 defmodule Astarte.Pairing.FDO.OwnerOnboarding.EAToken do
+  @moduledoc """
+  Functions for encoding and decoding Entity Attestation Tokens (EAT) used in FDO.
+
+  EAT tokens are the COSE_Sign1 objects that contain attestation claims about a device.
+
```
noaccOS: nit: extra newline

noaccOS (changes requested):
Great work

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
mizzet1: I removed it and added the keys directly inside `@known_payload_claims`

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -17,28 +17,85 @@
...
         eatoken_cbor,
         device_public_key,
-        extra_payload_claims \\ %{},
+        _extra_payload_claims \\ %{},
         extra_unprotected_header_claims \\ %{}
       ) do
     with {:ok, decoded_eatoken} <- decode_eatoken(eatoken_cbor),
          {:ok, :verified} <- verify_signature(decoded_eatoken, device_public_key),
-         {:ok, payload} <- verify_payload(decoded_eatoken.payload, extra_payload_claims) do
+         {:ok, payload} <- verify_payload(decoded_eatoken.payload, %{}) do
```
noaccOS: extra payload claims are still not being used

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/eatoken.ex
```diff
@@ -17,28 +17,85 @@
...
+        payload_claims,
+        uhdr_claims,
+        priv_key,
+        extra_unprotected_header_claims \\ %{}
+      ) do
+    eat_payload =
+      translate_claims_encode(
+        payload_claims,
+        @known_payload_claims,
+        %{}
```
noaccOS: also here

PR #1740, opened 2025-12-25
feat(pairing) Chunk OwnerServiceInfo messages 

nedimtokic:
- Rename field `max_service_info` to `max_owner_service_info_size` for improved clarity
- Implement chunking for `OwnerServiceInfo` when the message size exceeds the device limit

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/owner_service_info.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/owner_service_info.ex
@@ -72,6 +72,14 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.OwnerServiceInfo do
     |> CBOR.encode()
   end
 
+  def encode_with_service_info_chunk(owner_service_info, encoded_service_info_chunk) do
+    %OwnerServiceInfo{is_more_service_info: is_more_service_info, is_done: is_done} =
+      owner_service_info
+
+    [is_more_service_info, is_done, encoded_service_info_chunk]
+    |> CBOR.encode()
+  end
+
   def build(realm_name, credentials_secret, encoded_device_id) do
     service_info = %{
       "astarte:active" => true,
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
@@ -17,6 +17,8 @@
 #
 
 defmodule Astarte.Pairing.FDO.ServiceInfo do
+  import Astarte.Pairing.FDO.Types.ServiceInfo
+
   alias Astarte.Pairing.FDO.OwnerOnboarding.DeviceServiceInfo
   alias Astarte.Pairing.FDO.OwnerOnboarding.OwnerServiceInfo
   alias Astarte.Pairing.FDO.OwnerOnboarding.Session
@@ -38,6 +40,20 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
     {:ok, resp}
   end
 
+  # Case when the device yielded during Owner Service Info chunk transmission
+  # by sending an empty ServiceInfo map.
+  def build_owner_service_info(
+        realm_name,
+        session,
+        %DeviceServiceInfo{
+          is_more_service_info: false,
+          service_info: service_info
+        }
+      )
+      when is_empty(service_info) do
+    send_next_owner_chunk(session, realm_name)
+  end
+
   # Device has no more data to send, this function appends recevied part to the previous recevied
   # parts of service info and proceed sending OwnerService info
   def build_owner_service_info(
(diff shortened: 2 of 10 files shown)


frossq (approved):

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -48,25 +49,72 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
-    {:ok, session} = Session.add_device_service_info(session, realm_name, service_info)
+    with {:ok, session} <-
+           Session.add_device_service_info(session, realm_name, service_info) do
+      if ServiceInfo.empty?(service_info) do
+        send_next_owner_chunk(session, realm_name)
+      else
+        build_and_send_owner_service_info(session, realm_name)
+      end
+    end
+  end
```
noaccOS, line 72: I think we can avoid the `Session.add_device_service_info` call if the service info is empty
as we're already using multiple bodies for this function, I suggest doing something like

```elixir
# requires `empty?` to be defined using `defguard`
def build_owner_service_info(realm_name, session, service_info) when ServiceInfo.empty?(service_info) do
  # send_next_owner_chunk's body
end

def build_owner_service_info(realm_name, session, service_info) do
  encoded_device_id =
    session.device_service_info
    |> generate_device_id()
    |> Device.encode_device_id()

  with {:ok, session} <- Session.add_device_service_info(session, realm_name, service_info),
         encoded_device_id = generate_encoded_device_id(session.device_service_info),
         # build_and_send_owner_service_info's body
  end
end
```

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -48,25 +49,72 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+
+  def build_owner_service_info(_realm_name, _session_key, _) do
+    {:error, :message_body_error}
+  end
```
noaccOS: I don't think we need to check this, all patters should be handled by previous bodies

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -48,25 +49,72 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+
+  defp send_next_owner_chunk(session, realm_name) do
+    {nextIndex, response} = get_next_owner_chunk(session)
```
noaccOS: in general variables should be snake_case not camelCase

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -48,25 +49,72 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+    Session.update_last_chunk_sent(session, realm_name, nextIndex)
```
noaccOS: how about separating instruction based on concerns?

```elixir
with {:ok, _session, service_info_chunk} <- Session.next_owner_service_info_chunk(session, realm_name) do # session updates its index and does its magic, the user (ServiceInfo) does not need to know about it
  {:ok, service_info_chunk}
end
```

also, from the spec: 

> The Device MAY send additional (empty) TO2.DeviceServiceInfo messages to the Owner as a keepalive mechanism [[RFC1122]](https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#biblio-rfc1122). This is intended to allow the Device to perform lengthy computation between the end of ServiceInfo and the TO2.Done message without the Owner side timing out and declaring a failure.
> 
> A Device SHOULD send such keepalive messages if the interval between the first message containing IsDone=True and the TO2.Done message involves processing of long or unknown interval. A suggested keepalive interval for this phase of TO2 is 60 seconds.

so this `next_owner_service_info_chunk` should send an empty message if we've already sent all service info

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -48,25 +49,72 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+    |> Enum.map(fn {chunk_entries, idx} ->
+      is_last = idx == length(chunks) - 1
+
+      %OwnerServiceInfo{
+        is_more_service_info: !is_last,
+        is_done: is_last,
+        service_info: chunk_entries
+      }
+      |> OwnerServiceInfo.encode()
+    end)
```
noaccOS: IMO, it's better to be declarative in what we want to do instead of adding custom checks to _all_ elements

```suggestion
    # SAFETY: we always have at least one service info message
    {init_chunks, [last_chunk]} = Enum.split(chunks, -1)
    init_chunks = Enum.map(init_chunks, &%OwnerServiceInfo{is_more_service_info: true, is_done: false, service_info: &1})
    last_chunk = %OwnerServiceInfo{is_more_service_info: false, is_done: true, service_info: last_chunk}
    
    Enum.concat(init_chunks, [last_chunk])
    |> Enum.map(&OwnerServiceInfo.encode/1)
```

noaccOS (changes requested):

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
nedimtokic: Added guard and move logic for generating device Id to the `generate_encoded_device_id`

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
nedimtokic: Changed.

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
nedimtokic: Removed.

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
nedimtokic: Moved logic to the Session module, and it returns empty owner info message when necessary.

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -38,6 +40,20 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
...
+  def build_owner_service_info(
+        realm_name,
+        session,
+        %DeviceServiceInfo{
+          is_more_service_info: false,
+          service_info: service_info
+        }
+      )
+      when is_empty(service_info) do
+    send_next_owner_chunk(session, realm_name)
```
noaccOS, line 54: I think here we should always send the empty reply and not advancing the service info queue (a device should only send these messages after idDone=true but just to be extra safe)

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
```diff
@@ -48,25 +64,68 @@ defmodule Astarte.Pairing.FDO.ServiceInfo do
           service_info: service_info
         }
       ) do
-    # TODO: make use of other device service info
-    {:ok, session} = Session.add_device_service_info(session, realm_name, service_info)
+    with {:ok, session} <-
+           Session.add_device_service_info(session, realm_name, service_info),
+         encoded_device_id = generate_encoded_device_id(session.device_service_info) do
```
noaccOS: when possible, assignments should be placed either before or inside the with and not in the conditions (so that only "fallible" instructions are present in the assignments)

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/owner_service_info.ex
```diff
@@ -58,7 +57,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.OwnerServiceInfo do
     [
       t.is_more_service_info,
       t.is_done,
-      ServiceInfo.encode_map(t.service_info)
```
noaccOS, line 61: I think this is fine as it was: now we have separate chunks of owner service info and it's fine that we're chunking based on the encoded values directly, but I think the base structure should still use a map as the service info (and in fact it is used elsewhere in the module, such as `build_empty_owner_service_info`: this change would break those)

In your situation, we now want a shortcut to avoid multiple encoding/decoding to and from a map. Even without the shortcut, it wouldn't even that bad tbh, in normal situations we may have to encode/decode a map a couple more times at most. 

If we do want it, I suggest adding yet another function instead of editing the existing one:

```elixir
def encode_with_service_info_chunk(owner_service_info, encoded_service_info_chunk) do
  %OwnerServiceInfo{is_more_service_info: is_more_service_info, is_done: is_done} = owner_service_info
  [is_more_service_info, is_done, encoded_service_info_chunk]
  |> CBOR.encode()
end
```

```elixir
init_owner_service_info_length = Enum.count(chunks) - 1
init_owner_service_info = Enum.map(1..init_owner_service_info_length//1, fn _ ->
  %OwnerServiceInfo{is_more_service_info: true, is_done: false}
end)

last_owner_service_info = %OwnerServiceInfo{is_more_service_info: false, is_done: true}

owner_service_info = Enum.concat(init_owner_service_info, [last_owner_service_info])

Enum.zip(owner_service_info, chunks)
|> Enum.map(fn {owner_service_info, chunk} ->
  OwnerServiceInfo.encode_with_service_info(owner_service_info, chunk)
end)
```

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
nedimtokic: Acctually this part covers case from spec, see [5.5.10. TO2.DeviceServiceInfo, Type 68]:
```
If the previous TO2.OwnerServiceInfo.IsMoreServiceInfo had value True, then this message MUST contain:
- IsMoreServiceInfo = False
- ServiceInfo is an empty array
This permits the Owner to send arbitrary sized collections of ServiceInfo.
```

apps/astarte_pairing/lib/astarte_pairing/fdo/service_info.ex
nedimtokic: Changed.

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/owner_service_info.ex
nedimtokic: Implemented new `encode_with_service_info_chunk` function and updated `chunks_to_owner_service_info`.

PR #1742, opened 2026-01-09
build(docker): fix dev container build errors

mizzet1:
This pr fixes failures in building dev containers by hanging inter-service dependencies like astarte_realm_management and astarte_housekeeping from only: [:dev, :test] to only: :test


apps/astarte_housekeeping/mix.exs
```diff
@@ -98,7 +98,7 @@ defmodule Astarte.Housekeeping.Mixfile do
       # Workaround for Elixir 1.15 / ssl_verify_fun issue
       # See also: https://github.com/deadtrickster/ssl_verify_fun.erl/pull/27
       {:ssl_verify_fun, "~> 1.1.0", manager: :rebar3, override: true},
-      {:credo, "~> 1.7", only: [:dev, :test], runtime: false},
+      {:credo, "~> 1.7", only: :test, runtime: false},
```
noaccOS: somehow I missed this, but credo should also be available in :dev

noaccOS (approved):

PR #1747, opened 2026-01-13
ci: Add CI for shared libraries

nedimtokic:
Added reusable CI for all libraries.
Add build and test workflows for all existing libraries in `libs/` directory.

### DIFF
--- a/libs/astarte_data_access/config/test.exs
+++ b/libs/astarte_data_access/config/test.exs
@@ -1,10 +1,1 @@
 import Config
-
-cassandra_host = System.get_env("CASSANDRA_DB_HOST", "cassandra")
-
-cassandra_port =
-  System.get_env("CASSANDRA_DB_PORT", "9042")
-  |> String.to_integer()
-
-config :astarte_data_access,
-  xandra_nodes: "#{cassandra_host}:#{cassandra_port}"
--- a/libs/astarte_data_access/test/test_helper.exs
+++ b/libs/astarte_data_access/test/test_helper.exs
@@ -2,12 +2,6 @@ Mimic.copy(Xandra)
 Mimic.copy(Astarte.DataAccess.Realms.Realm)
 ExUnit.start(capture_log: true)
 
-children = [
-  {Astarte.DataAccess, xandra_options: Astarte.DataAccess.Config.xandra_options!()}
-]
-
-Supervisor.start_link(children, strategy: :one_for_one)
-
 # Register the Xandra cluster process so that it can be used in tests
 # Yes they both have id 'Astarte.DataAccess.Repo'
 Astarte.DataAccess


.github/workflows/astarte-libs-build-workflow.yaml
```diff
@@ -0,0 +1,137 @@
...
+
+      - name: Check if ExCoveralls is available
+        id: has_excoveralls
+        run: |
+          if mix deps | grep -q "excoveralls"; then
+            echo "found=true" >> $GITHUB_OUTPUT
+          else
+            echo "found=false" >> $GITHUB_OUTPUT
+          fi
+        shell: bash
```
noaccOS: let's just add the dependency to all projects instead of having this custom check

.github/workflows/astarte-libs-build-workflow.yaml
```diff
@@ -0,0 +1,137 @@
...
+          github.repository == 'astarte-platform/astarte' &&
+          steps.coverage_check.outputs.found == 'true'
+          }}
+        uses: codecov/codecov-action@v5
+        with:
+          fail_ci_if_error: true
+          token: ${{ secrets.CODECOV_TOKEN }}
+          verbose: true
+          directory: ./libs/${{ inputs.lib }}/coverage_results
+          flags: lib-${{ inputs.lib }}
```
noaccOS: we should add relevant flags to `.github/codecov.yml`
I'm also not sold on the `lib-` prefix, I think I'd rather see only the library's name, or something like `libs/library_name` if it is an accepted flag name

.github/workflows/astarte-data-access-workflow.yaml
```diff
@@ -0,0 +1,39 @@
+name: Build and Test Astarte Data Access (lib)
```
noaccOS: I'm also not sold on the `(lib)` suffix here but I don't have a strong opinion

noaccOS (changes requested):

.github/workflows/astarte-libs-build-workflow.yaml
```diff
@@ -0,0 +1,126 @@
...
+      - name: Check for Coverage Results
+        id: coverage_check
+        run: |
+          if [ -d "./libs/${{ inputs.lib }}/coverage_results" ] && [ "$(ls -A ./libs/${{ inputs.lib }}/coverage_results)" ]; then
+            echo "found=true" >> $GITHUB_OUTPUT
+          else
+            echo "found=false" >> $GITHUB_OUTPUT
+          fi
+        shell: bash
+
```
noaccOS: I think we also don't need this? It should always be present

.github/workflows/astarte-libs-build-workflow.yaml
```diff
@@ -0,0 +1,126 @@
...
+      - name: Upload Coverage Results to CodeCov
+        if: ${{
+          matrix.database == 'scylladb/scylla:2025.2.2' &&
+          github.repository == 'astarte-platform/astarte' &&
+          steps.coverage_check.outputs.found == 'true'
```
noaccOS: and this check right here

.github/workflows/astarte-libs-build-workflow.yaml
nedimtokic: If a lib doesn't have tests, like `astarte_events`,  coverage report is not generated.

noaccOS: Let's fix that! https://github.com/astarte-platform/astarte/pull/1754

noaccOS (approved):

PR #1748, opened 2026-01-13
docs(pairing): corrected shell commands in readme to run mix test

mizzet1:
This PR fix shell commands in README needed to run Mix Test.


noaccOS (approved):
:trollface:

PR #1751, opened 2026-01-14
Fix static headers type in AMQPTriggerTarget

mizzet1:
This PR add assertion that the type of static headers in AMQPTriggerTarget is actually a map and not a list.


libs/astarte_events/lib/astarte_events/triggers_handler/core.ex
```diff
@@ -60,13 +60,18 @@ defmodule Astarte.Events.TriggersHandler.Core do
       simple_event.parent_trigger_id
       |> UUID.binary_to_string!()
 
+    if is_list(static_headers) do
+      raise ArgumentError,
+            "static_headers must be a map, got a list instead: #{inspect(static_headers)}"
+    end
+
```
noaccOS: this is very useful for testing to get back a stacktrace when the use of the static headers was improper, but it should be removed before merging

PR #1755, opened 2026-01-15
Add credo to CI

nedimtokic:
Added Credo dependency to all Astarte apps where it was missing.
Added Credo check in CI for Astarte apps.


noaccOS (approved):
Thank you 🌻

PR #1758, opened 2026-01-16
docs: update for 1.3

noaccOS:
- update internal api section to mention the use of native erlang messages
- update source_ref value, so that clicking on the logo takes the user to the appropriate version documentation instead of 1.0
- fix astarte dashboard documentation reference
- update ex_doc to have access to the --warnings-as-errors flag
- update assets option to solve deprecation in new ex_doc version
- fix superscript block inside italic section not being rendered by using the html tags directly instead of markdown syntax
- add ci checks for docs generation without warnings


AliouneGaye21 (approved):

doc/pages/architecture/010-design_principles.md
```diff
@@ -24,7 +24,7 @@ This detail is relevant not only for identifying and querying the device, but al
...
 
-_Note: As much as Device IDs should effectively be unique per-realm and this configuration will always be supported, some future optional optimizations might be available on top of the assumption that Device IDs are globally unique to an Astarte installation. Given the Device ID format has a 2<sup>-128</sup> chance of collision, it is safe to assume that as long as best practices for Device ID generation are followed, Device IDs will always be globally unique._
+<p><em>Note: As much as Device IDs should effectively be unique per-realm and this configuration will always be supported, some future optional optimizations might be available on top of the assumption that Device IDs are globally unique to an Astarte installation. Given the Device ID format has a 2<sup>-128</sup> chance of collision, it is safe to assume that as long as best practices for Device ID generation are followed, Device IDs will always be globally unique.</em></p>
```
Annopaolo, line 28: If we change this line, why not changing also line 25 above? Both are notices that require some attention from the user

doc/pages/architecture/010-design_principles.md
```diff
@@ -8,9 +8,9 @@ Astarte does not allow exchanging raw data - it rather forces the user to descri
 
 Data is described with a mechanism named [Interfaces, explained in detail in the user guide](030-interface.html). Through Interfaces, Astarte creates and maintains a data model autonomously, sparing the user from the complexity of dealing with Databases and Data Management in general.
 
-## AMQP as internal API mechanism
+## Internal API mechanism
 
-Astarte services use a Protobuf-based API to exchange data over AMQP in a [gRPC](https://grpc.io/) like fashion. As such, as long as a service conforms with the policies defined by the queues, it is possible to extend Astarte in virtually any language that can deliver a compliant AMQP client.
+Astarte services generally use native erlang messages to exchange data, but resort to protobuf over AMQP for the main data path in order to increase resiliency.
```
Annopaolo:
```suggestion
Astarte services generally use native Erlang messages to exchange data, but resort to a Protobuf-based API over AMQP for the device-to-cloud data path in order to increase resiliency and to provide a backpressure mechanism.
```

.github/workflows/docs-check.yml
```diff
@@ -0,0 +1,63 @@
...
+    needs:
+      - warmup
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v6
+      - uses: team-alembic/staple-actions/actions/mix-compile@a74b3b61209d35d45526df174766632f8aee03ed
+        with:
+          mix-env: dev
+          working-directory: doc/
+      - run: mix docs --warnings-as-errors
```
Annopaolo, line 52: Docs generation is already handled in the `docs-workflow.yaml` workflow: https://github.com/astarte-platform/astarte/blob/f6ce850be57f71e1016502b9457e72ba00ff4b8b/.github/workflows/docs-workflow.yaml#L39-L41

I suggest to keep only one of the two

doc/pages/architecture/010-design_principles.md
noaccOS: > fix superscript block inside italic section not being rendered by using the html tags directly instead of markdown syntax

line 25 and 27 are rendered exactly the same, but using underscores the `sup` is not rendered correctly: https://docs.astarte-platform.org/astarte/latest/010-design_principles.html#device-id

Annopaolo: I see your point, nice catch. But I would value keeping both rows consistent in order to make sure our future selves do not come across this and change it back to `_`

.github/workflows/docs-check.yml
noaccOS: If all works as it should, most of it should be cached by staple-actions.

Unfortunately, we need to generate the documentation for multiple reasons:
- checking it compiles correctly and without warnings (in ci)
- publishing it (only when it's merged)

doc/pages/architecture/010-design_principles.md
noaccOS: Changing both to `<p><em>` imo would be worse in that regard: "why are these two html tags instead of standard markdown"?
As an alternative, I would suggest adding a markdown comment before the second note

```markdown
<!-- This needs explicit html tags instead of underscores so that the sup tag is correctly formatted -->
<p><em>Note: As much as Device IDs should effectively be unique per-realm and this configuration will always be supported, some future optional optimizations might be available on top of the assumption that Device IDs are globally unique to an Astarte installation. Given the Device ID format has a 2<sup>-128</sup> chance of collision, it is safe to assume that as long as best practices for Device ID generation are followed, Device IDs will always be globally unique.</em></p>
```

Annopaolo: Way to go!

noaccOS: Comment added

PR #1759, opened 2026-01-19
ci: Add conventional commit check

noaccOS:
Add a check for conventional commits using [committed]

As we want to generate the changelog from commit messages, add relevant checks such that entries can be placed in the changelog without manual intervention

A selected list of allowed scopes has been populated, but more can be added if anything's been forgotten

The subject and line length have been set to the maximum recommended by GitHub

See also [how we use it in edgehog] and the [configuration reference]

[committed]: https://github.com/crate-ci/committed
[how we use it in edgehog]: https://github.com/edgehog-device-manager/edgehog/blob/main/committed.toml
[configuration reference]: https://github.com/crate-ci/committed/blob/master/docs/reference.md

### DIFF
--- a/committed.toml
+++ b/committed.toml
@@ -1,0 +1,51 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2026 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


committed.toml
```diff
@@ -0,0 +1,51 @@
...
+  "generators",
+  # tools
+  "e2e",
+  "import",
+  "export",
+  "fleet",
+  # build/dev tools
+  "docker",
+  "credo",
+  "nix",
```
nedimtokic, line 44:
```suggestion
  "nix"
```

noaccOS: trailing commas are
1. standard practice in many languages nowadays, because they are
    1. easier for the programmer as all lines are the same
    2. diffs for adding new entries at the end are always 1 line long instead of 2 (you don't have to edit the last element)
2. required by our dprint ci check in toml

frossq (approved):

lusergit (approved):
:gold:

PR #1762, opened 2026-01-19
ci: Run libraries ci for astarte_generators

noaccOS:
after the merge of generators in the monorepo, the ci was not updated
to run its tests.

fix this by using the same workflow as other libraries

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -19,6 +19,7 @@ coverage:
           - astarte_trigger_engine
           - astarte_data_access
           - astarte_events
+          - astarte_generators
           - astarte_rpc
 
 ignore:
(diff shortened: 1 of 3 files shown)


noaccOS:
> No PR description provided. I would like there to be something like "[..] after integrating `astarte_generators` into the monorepo, it was necessary [..]".
> 
> In addition, there is a _template_ to follow for descriptions. If you consider it obsolete, we can change it, but until then we will follow it.

I can add a description if you find it to be important, though I'd argue that you don't need motivation to start testing something if it was not tested before

as for the template,
1. yes it would be a good idea to update it to match our current commit convention
2. it's good to have as a checklist for inexperienced developers approaching astarte to know what they need to do before opening a PR, but I would not treat it as a law
3. personally I open most of my prs using the `gh` cli, which of course doesn't allow _editing_ the template but instead you just add a description to it

PR #1766, opened 2026-01-20
feat: Implement FDO Replacement Ownership Voucher Persistence

eddbbt:
This PR implements the logic to generate and persist the replacement Ownership Voucher at the end of TO2

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
@@ -150,7 +150,6 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
          {:ok, private_key} <- Queries.get_owner_private_key(realm_name, guid),
          {:ok, owner_public_key} <- OwnershipVoucher.owner_public_key(ownership_voucher) do
       rendezvous_info = ownership_voucher.header.rendezvous_info
-      dbg(rendezvous_info)
       {:ok, private_key} = COSE.Keys.from_pem(private_key)
 
       connection_credentials = %{
@@ -175,7 +174,6 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
             replacement_pub_key: owner_public_key
         }
 
-        dbg(session)
         {:ok, session, resp_msg}
       end
     end
--- a/apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
@@ -62,6 +62,8 @@ defmodule Astarte.PairingWeb.FDOOnboardingController do
              cbor_body,
              conn.assigns.to2_session
            ) do
+      dbg(session)
+
       conn
       |> assign(:to2_session, session)
       |> render("secure.cbor", %{response: response})
@@ -71,6 +73,7 @@ defmodule Astarte.PairingWeb.FDOOnboardingController do
   def done(conn, _params) do
     realm_name = Map.fetch!(conn.params, "realm_name")
     to2_session = conn.assigns.to2_session
+    dbg(to2_session)
 
     with {:ok, response_msg} <- OwnerOnboarding.done(realm_name, to2_session, conn.assigns.body) do
       conn


apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
```diff
@@ -115,13 +118,19 @@ defmodule Astarte.Pairing.FDO.OwnershipVoucher do
...
+      ownership_voucher.header
+      |> Map.put(:guid, session.new_guid)
+      |> Map.put(:rendezvous_info, session.new_rv_info)
+      |> Map.put(:public_key, session.new_pub_key)
+
+    new_voucher =
+      ownership_voucher
+      |> Map.put(:hmac, session.new_hmac)
+      |> Map.put(:header, new_header)
+      |> Map.put(:entries, [])
```
noaccOS, line 130: I think this is for the resell protocol?

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
```diff
@@ -142,4 +151,14 @@ defmodule Astarte.Pairing.FDO.OwnershipVoucher do
   def cbor_encode(voucher) do
     encode(voucher) |> CBOR.encode()
   end
+
+  # TODO to be moved
```
noaccOS: let's move it!

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -158,6 +158,10 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
                body,
                connection_credentials
              ) do
+        session = %{session | new_guid: device_guid}
+        session = %{session | new_rv_info: rendezvous_info}
+        session = %{session | new_pub_key: owner_public_key}
```
noaccOS: nit: maybe use a single instruction?


```suggestion
        session = %{session | new_guid: device_guid, new_rv_info: rendezvous_info, new_pub_key: owner_public_key}
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -244,19 +248,22 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
...
            Session.add_max_owner_service_info_size(session, realm_name, max_owner_service_info_sz) do
-      # TODO: Store `new_voucher` into DB.
+      session =
+        if !is_nil(replacement_hmac) do
+          session = %{session | save_new_voucher: true}
+
+          %{session | new_hmac: replacement_hmac}
+        else
+          session
+        end
```
noaccOS: this is implemented differently compared to owner info: either we always store all `new_` info regardless if they changed or not, or we always leave the field `nil` if they didn't change

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -244,19 +248,22 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
...
-           OwnershipVoucher.fetch(realm_name, session.device_id),
-         {:ok, _new_voucher} <-
-           OwnershipVoucher.generate_replacement_voucher(old_voucher, replacement_hmac),
-         {:ok, _session} <-
+    with {:ok, session} <-
            Session.add_max_owner_service_info_size(session, realm_name, max_owner_service_info_sz) do
-      # TODO: Store `new_voucher` into DB.
+      session =
+        if !is_nil(replacement_hmac) do
+          session = %{session | save_new_voucher: true}
```
noaccOS: I think this check should be extended: https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#credreuse

**all** the conditions are required in order to have credential reuse.

Instead of adding this boolean to the session, I suggest simply having a `credential_reuse?` function to the session, which compares new values to old values

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -272,6 +279,33 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
            check_prove_dv_nonces_equality(prove_dv_nonce_challenge, to2_session.prove_dv_nonce),
          {:ok, _device} <- Queries.remove_device_ttl(realm_name, to2_session.device_id) do
       done2_message = build_done2_message(to2_session.setup_dv_nonce)
+
+      if to2_session.save_new_voucher do
+        with {:ok, old_voucher} <-
+               OwnershipVoucher.fetch(realm_name, to2_session.device_id),
+             {:ok, new_voucher} <-
+               OwnershipVoucher.generate_replacement_voucher(old_voucher, to2_session),
+             {:ok, private_key} <-
```
noaccOS, line 302: this doesn't even work, as it **always** fetches the old key
which is fine as we currently do not offer an api to change the key, but I think we should at least add a comment

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -247,7 +247,8 @@ defmodule Astarte.PairingWeb.FDOOnboardingControllerTest do
       conn: conn,
       create_path: path,
       message_id: id,
-      session: session
+      session: session,
+      realm_name: realm_name
```
noaccOS: ?

apps/astarte_pairing/test/astarte_pairing/fdo/owner_onboarding_test.exs
```diff
@@ -39,24 +40,27 @@ defmodule Astarte.Pairing.FDO.OwnerOnboardingTest do
...
       device_max_size = 2048
 
-      assert {:ok, response} =
+      assert {:ok, session, response} =
                OwnerOnboarding.build_owner_service_info_ready(
                  realm_name,
                  session,
                  %DeviceServiceInfoReady{
-                   replacement_hmac: new_hmac,
+                   replacement_hmac: %Hash{hash: new_hmac, type: :hmac_sha256},
```
noaccOS: this adds yet another nested layer, let's move this to the declaration above

```elixir
new_hmac_value = :crypto.strong_rand_bytes(32)
new_hmac = %Hash{type: :hmac_sha256, hash: new_hmac_value}
```

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/done_test.exs
```diff
@@ -69,14 +71,28 @@ defmodule Astarte.Pairing.FDO.Onboarding.DoneTest do
       realm: realm_name,
       session: session
     } do
-      done_msg = [%CBOR.Tag{tag: :bytes, value: session.prove_dv_nonce}]
+      {:ok, ownership_voucher} = OwnershipVoucher.fetch(realm_name, session.device_id)
+      {:ok, owner_public_key} = OwnershipVoucher.owner_public_key(ownership_voucher)
+      rendezvous_info = ownership_voucher.header.rendezvous_info
+      new_hmac = :crypto.strong_rand_bytes(32)
+
+      session = %{session | save_new_voucher: true}
```
noaccOS: let's add another test for `stores the new ownership voucher when not reusing credentials` instead of editing the one for the return type

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
eddbbt: > as it always fetches the old key

yeah that's kinda the point? at this step we do not have a new key nor a new voucher, so ok lets track this with a comment

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
eddbbt: if I remember correctly, those apply to all new vouchers, including resell and not reusing credential

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/done_test.exs
```diff
@@ -69,14 +71,27 @@ defmodule Astarte.Pairing.FDO.Onboarding.DoneTest do
...
-      done_msg = [%CBOR.Tag{tag: :bytes, value: session.prove_dv_nonce}]
+      {:ok, ownership_voucher} = OwnershipVoucher.fetch(realm_name, session.device_id)
+      {:ok, owner_public_key} = OwnershipVoucher.owner_public_key(ownership_voucher)
+      rendezvous_info = ownership_voucher.header.rendezvous_info
+      new_hmac = :crypto.strong_rand_bytes(32)
+
+      session = %{session | new_guid: session.device_id}
+      session = %{session | new_hmac: %Hash{hash: new_hmac, type: :hmac_sha256}}
+      session = %{session | new_rv_info: rendezvous_info}
+      session = %{session | new_pub_key: owner_public_key}
```
frossq: Also here may be more readable to do a single map update

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
```diff
@@ -142,4 +149,12 @@ defmodule Astarte.Pairing.FDO.OwnershipVoucher do
   def cbor_encode(voucher) do
     encode(voucher) |> CBOR.encode()
   end
+
+  def credential_reuse?(session) do
+    # TODO credential reuse requires also Owner2Key and/or rv info to be changed for credential reuse
+    # so far, there is no API to do so, so it-s limited to the guid
```
noaccOS, line 155: there is also no api to change the guid

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
```diff
@@ -142,4 +149,12 @@ defmodule Astarte.Pairing.FDO.OwnershipVoucher do
...
+
+    is_nil(session.new_hmac) && session.device_id == session.new_guid &&
```
noaccOS: why do we still have two distinct syntaxes for hmac update and guid update?
either both should be nil or both should have a value (I would argue the second option makes more sense)

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -134,7 +135,15 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
   end
 
   def prove_device(realm_name, body, session) do
-    device_guid = session.device_id
+    device_guid =
+      if Config.enable_credential_reuse!() do
+        session.device_id
+      else
+        :crypto.strong_rand_bytes(16)
+      end
```
noaccOS: why?

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
```diff
@@ -46,6 +47,10 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding.Session do
     field :device_service_info, map() | nil
     field :owner_service_info, [binary()] | nil
     field :last_chunk_sent, non_neg_integer() | nil
+    field :new_guid, binary() | nil
+    field :new_rv_info, [RvTO2Addr.t()] | nil
+    field :new_pub_key, struct() | nil
+    field :new_hmac, binary() | nil
```
noaccOS: nit: in the spec they're referred to as `replacement`, how about keeping it consistent?


```suggestion
    field :replacement_guid, binary() | nil
    field :replacement_rv_info, [RvTO2Addr.t()] | nil
    field :replacement_pub_key, struct() | nil
    field :replacement_hmac, binary() | nil
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
noaccOS: there is no comment still

apps/astarte_pairing/lib/astarte_pairing/config.ex
```diff
@@ -111,6 +111,14 @@ defmodule Astarte.Pairing.Config do
     type: :atom,
     default: :trigger_cache
 
+  @envdoc """
+  Enable credential reuse protocol for FDO onboarding
+  """
+  app_env :enable_credential_reuse, :astarte_pairing, :enable_credential_reuse,
```
noaccOS, line 132: when we add a new env (for now) let's also update the changelog + inform the platform team once the pr is merged

apps/astarte_pairing/test/astarte_pairing/fdo/onboarding/done_test.exs
noaccOS: still relevant

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
eddbbt: there shall not as it's randomly generated form the owner

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
eddbbt: > The Credential Reuse protocol is as follows:
In TO2.SetupDevice:
If TO2.SetupDevice.Guid == TO2.ProveOVHdr.OVHeader.OVGuid (GUID same as previous)

this if the owner allow to reuse the credentials

hmac nil or not equals to the ack from the device, so yes, syntax needs to be different

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
eddbbt: cause if credendial reuse is not enabled, we need a new GUID

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
noaccOS: thats-a-nice-argument-senator-why-dont-you-back-it-up-with-a-source

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
noaccOS: I am referring to our internal representation not how we receive the data from the device. I would expect the same representation for all the replacement_ variables

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
noaccOS: - the guid is not the id, it's the guid
- why random?

apps/astarte_pairing/lib/astarte_pairing/fdo/ownership_voucher/ownership_voucher.ex
eddbbt: @noaccOS 

https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#guid

 https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#to2setupdevice
 
 
image

as it (replacement guid) first appers here, and as you can see its value equals to `TO2SetupDevicePayload.Guid` its safe to say that it must be generated here, before message 65, by the owner. I can't find another place where the spec says "the new guid shall come from this calculation or this api or whatever"

GUIDs are also made to be disposable 
> The Device has a GUID that can be used to identify it to its new owner. This GUID is also known to the FIDO Device Onboard Manufacturing Component, and is in the Ownership Voucher. The GUID is not a secret. Specifically, the GUID is intended to be visible to the Owner when the device shipped in a box, perhaps being on the box itself with a bar code, perhaps being on the bill of lading. The GUID is used for one FIDO Device Onboard transfer of ownership only; after Transfer Ownership Protocol 2, the GUID is replaced, and the Device has no memory of the original GUID.



happy to implement a stronger GUID generation if you have to suggest something else

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
eddbbt: > the guid is not the id, it's the guid

looks the same to me sir, as its been set by to2.helloDevice (guid, see spec)
If thats the case, we have a naming problem here, so I'll be happy to be worng


image
image

image

image
image

noaccOS:
Superseded by #1801

PR #1769, opened 2026-01-21
refactor(fdo): use a phoenix.token as the session key 

mizzet1:
This PR add a 1:1 relationship between GUID and session for the FDO.TO2 protocol, by using a Phoenix Token where:

- The guid is stored in the claims, that way we can deduce the guid from a given jwt
- A nonce is used to prevent replay attacks
- The token is a string so that it can be valid for the header

This change prevents the situation in which a device start an infinite number of sessions and move forward with all of them, a behaviour that can potentially doss astarte.

### DIFF
--- a/apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
+++ b/apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
@@ -554,8 +554,8 @@ defmodule Astarte.Housekeeping.Realms.Queries do
     CREATE TABLE #{keyspace_name}.ownership_vouchers (
       private_key blob,
       voucher_data blob,
-      device_id uuid,
-      PRIMARY KEY (device_id, voucher_data)
+      guid blob,
+      PRIMARY KEY (guid, voucher_data)
     );
     """
 
@@ -567,10 +567,11 @@ defmodule Astarte.Housekeeping.Realms.Queries do
   defp create_to2_sessions_table(keyspace_name) do
     query = """
     CREATE TABLE #{keyspace_name}.to2_sessions (
-      session_key blob,
+      guid blob,
+      device_id uuid,
+      nonce blob,
       sig_type int,
       epid_group blob,
-      device_id uuid,
       device_public_key blob,
       prove_dv_nonce blob,
       setup_dv_nonce blob,
@@ -585,7 +586,7 @@ defmodule Astarte.Housekeeping.Realms.Queries do
       device_service_info map<tuple<text, text>, blob>,
       owner_service_info list<blob>,
       last_chunk_sent int,
-      PRIMARY KEY (session_key)
+      PRIMARY KEY (guid)
     )
     WITH default_time_to_live = 7200;
     """
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
@@ -52,12 +52,12 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
 
   def hello_device(realm_name, cbor_hello_device) do
     with {:ok, hello_device} <- HelloDevice.decode(cbor_hello_device),
-         device_id = hello_device.device_id,
-         {:ok, ownership_voucher} <- OwnershipVoucher.fetch(realm_name, device_id),
-         {:ok, owner_private_key} <- fetch_owner_private_key(realm_name, device_id),
+         guid = hello_device.guid,
+         {:ok, ownership_voucher} <- OwnershipVoucher.fetch(realm_name, guid),
+         {:ok, owner_private_key} <- fetch_owner_private_key(realm_name, guid),
          {:ok, pub_key} <- OwnershipVoucher.owner_public_key(ownership_voucher),
          :ok <- KeyExchangeStrategy.validate(hello_device.kex_name, owner_private_key),
-         {:ok, session} <-
+         {:ok, token, session} <-
            Session.new(
              realm_name,
              hello_device,
@@ -94,7 +94,7 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
           owner_private_key
         )
 
-      {:ok, session.key, message}
+      {:ok, token, message}
     else
       error ->
         Logger.error("Failed to process hello_device: #{inspect(error)}")
@@ -102,8 +102,8 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
     end
   end
 
-  defp fetch_owner_private_key(realm_name, device_id) do
-    with {:ok, pem_key} <- Queries.get_owner_private_key(realm_name, device_id) do
+  defp fetch_owner_private_key(realm_name, guid) do
+    with {:ok, pem_key} <- Queries.get_owner_private_key(realm_name, guid) do
       COSE.Keys.from_pem(pem_key)
     end
   end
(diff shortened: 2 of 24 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session_token.ex
```diff
@@ -0,0 +1,57 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2025 SECO Mind Srl
```
nedimtokic:
```suggestion
# Copyright 2026 SECO Mind Srl
```

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding.ex
```diff
@@ -52,12 +52,12 @@ defmodule Astarte.Pairing.FDO.OwnerOnboarding do
 
   def hello_device(realm_name, cbor_hello_device) do
     with {:ok, hello_device} <- HelloDevice.decode(cbor_hello_device),
-         device_id = hello_device.device_id,
-         {:ok, ownership_voucher} <- OwnershipVoucher.fetch(realm_name, device_id),
-         {:ok, owner_private_key} <- fetch_owner_private_key(realm_name, device_id),
+         guid = hello_device.guid,
+         {:ok, ownership_voucher} <- OwnershipVoucher.fetch(realm_name, guid),
+         {:ok, owner_private_key} <- fetch_owner_private_key(realm_name, guid),
```
nedimtokic, line 57: `OwnershipVoucher.fetch` and `fetch_owner_private_key` have device_id parameter. Let's rename it to be consistent.

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
```diff
@@ -567,10 +567,11 @@ defmodule Astarte.Housekeeping.Realms.Queries do
   defp create_to2_sessions_table(keyspace_name) do
     query = """
     CREATE TABLE #{keyspace_name}.to2_sessions (
-      session_key blob,
+      guid blob,
+      device_id blob,
```
nedimtokic: Why we changed device_id type?

mizzet1: I agree that uuid is better and more efficient than blob there

apps/astarte_pairing/lib/astarte_pairing_web/plug/fdo_session.ex
```diff
@@ -31,10 +31,12 @@ defmodule Astarte.PairingWeb.Plug.FDOSession do
   def call(conn, _opts) do
     realm_name = Map.fetch!(conn.path_params, "realm_name")
 
-    with [session_key] <- get_req_header(conn, "authorization"),
-         {:ok, session} <- Session.fetch(realm_name, session_key) do
+    [token] = get_req_header(conn, "authorization")
```
noaccOS: let's keep this on the with block, so that if the device sends us multiple authorization headers we don't just crash

apps/astarte_pairing/test/support/helpers/database.ex
```diff
@@ -562,12 +563,12 @@ defmodule Astarte.Helpers.Database do
...
   end
 
-  def delete_session(realm_name, session_key) do
+  def delete_session(realm_name, guid) do
     keyspace = Realm.keyspace_name(realm_name)
 
     query =
       from s in TO2Session,
-        where: s.session_key == fragment("uuidAsBlob(?)", ^session_key)
+        where: s.guid == fragment("uuidAsBlob(?)", ^guid)
```
noaccOS: the guid is not necessarily a uuid so I would just use it as it is


```suggestion
        where: s.guid == ^guid
```

apps/astarte_pairing/test/support/helpers/database.ex
noaccOS: here it should also be a binary then

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -361,15 +359,12 @@ defmodule Astarte.Pairing.Queries do
     end
   end
 
-  def fetch_session(realm_name, session_key) do
-    with {:ok, session_key} <- Astarte.DataAccess.UUID.dump(session_key) do
+  def fetch_session(realm_name, guid) do
+    with {:ok, guid} <- Astarte.DataAccess.UUID.dump(guid) do
```
noaccOS: the guid is just the binary

```suggestion
```

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -289,69 +289,67 @@ defmodule Astarte.Pairing.Queries do
     |> Repo.insert(opts)
   end
 
-  def store_session(realm_name, session_key, session) do
-    with {:ok, session_key} <- Astarte.DataAccess.UUID.dump(session_key) do
+  def store_session(realm_name, guid, session) do
+    with {:ok, guid} <- Astarte.DataAccess.UUID.dump(guid) do
```
noaccOS: same as above


```suggestion
```

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
noaccOS:
```suggestion
      guid blob,
```

also the migration file and the data_access type should be updated

apps/astarte_pairing/lib/astarte_pairing/fdo/owner_onboarding/session.ex
```diff
@@ -14,23 +14,22 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
-
```
noaccOS, line 18: let's not delete this

apps/astarte_pairing/lib/astarte_pairing_web/plug/fdo_session.ex
```diff
@@ -31,10 +31,12 @@ defmodule Astarte.PairingWeb.Plug.FDOSession do
...
+
+    with {:ok, guid, _nonce} <- Astarte.Pairing.FDO.OwnerOnboarding.SessionToken.verify(token),
```
noaccOS: we also need to validate that the nonce is the same as session.nonce

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -37,10 +37,11 @@ defmodule Astarte.Pairing.Queries do
     keyspace_name = Realm.astarte_keyspace_name()
 
     query =
-      from r in Realm,
+      from(r in Realm,
```
noaccOS: F formatter

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -243,11 +244,12 @@ defmodule Astarte.Pairing.Queries do
 
     # FIXME: functions that depends on this one shall handle one or more ownership voucher, keeping just the first for now
     query =
-      from o in OwnershipVoucher,
+      from(o in OwnershipVoucher,
```
noaccOS: F formatter

apps/astarte_pairing/lib/astarte_pairing/queries.ex
```diff
@@ -289,69 +292,65 @@ defmodule Astarte.Pairing.Queries do
...
+  def session_update_last_chunk_sent(realm_name, guid, last_chunk) do
     updates = [last_chunk_sent: last_chunk]
-    update_session(realm_name, session_key, updates)
+    update_session(realm_name, guid, updates)
   end
 
-  defp update_session(realm_name, session_key, updates) do
-    with {:ok, session_key} <- Astarte.DataAccess.UUID.dump(session_key) do
+  defp update_session(realm_name, guid, updates) do
+    with {:ok, guid} <- Astarte.DataAccess.UUID.dump(guid) do
```
noaccOS: this still has the dump

PR #1776, opened 2026-01-26
test(events): Add full coverage tests for TriggersHandler

nedimtokic:
Add tests for TriggersHandler to ensure full coverage. Tests now verify that dispatch_event/7 correctly builds the SimpleEvent struct and forwards it to Core.dispatch_event/3, including the edge case when policy_name is nil.
Also test that register_target/2 properly delegates to Core.

### DIFF
--- a/libs/astarte_events/test/triggers_handler/triggers_handler_test.exs
+++ b/libs/astarte_events/test/triggers_handler/triggers_handler_test.exs
@@ -1,0 +1,109 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2026 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


libs/astarte_events/test/triggers_handler/triggers_handler_test.exs
```diff
@@ -0,0 +1,109 @@
...
+      simple_trigger_id: UUID.uuid4(:raw),
+      parent_trigger_id: UUID.uuid4(:raw)
+    }
+
+    {:ok, realm: realm, target: target}
+  end
+
+  describe "register_target/2" do
+    test "register_target/2 delegates to Core", %{realm: realm, target: target} do
+      Core |> expect(:register_target, fn _realm, ^target -> :ok end)
```
noaccOS, line 48: is this line needed? I believe it should work regardless

PR #1777, opened 2026-01-26
test(events): add tests for Queries module

nedimtokic:
Mock trigger and device data and test all queries.

### DIFF
--- a/libs/astarte_events/test/astarte_events/triggers/queries_test.exs
+++ b/libs/astarte_events/test/astarte_events/triggers/queries_test.exs
@@ -1,0 +1,148 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2026 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


libs/astarte_events/test/astarte_events/triggers/queries_test.exs
```diff
@@ -0,0 +1,148 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Events.Triggers.QueriesTests do
+  use Astarte.Cases.Data, async: true
+
+  import Mimic
```
noaccOS, line 22: nit for the future: you can `use Mimic` which is just `import` + `verify_on_exit` https://github.com/edgurgel/mimic/blob/v2.3.0/lib/mimic.ex#L89

PR #1778, opened 2026-01-27
feat(pairing): FDO feature gate

frossq:
Implement the FDO feature gate, adding a new env var "PAIRING_ENABLE_FDO" to select whether FDO feature is enabled or not in an Astarte installation (default: false).
When FDO is disabled, all the FDO API endpoints return HTTP error 404 to any client request.

Added relevant unit tests and updated the CHANGELOG.

Fixes #

* [x] Yes
* [ ] No

```docs

```

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/config.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/config.ex
@@ -33,23 +33,29 @@ defmodule Astarte.Pairing.Config do
     type: :binary,
     required: true
 
+  @envdoc """
+  Set this variable to 'true' to enable FDO feature as device authentication mechanism.
+  WARNING: this feature is experimental and is not enabled by default
+  """
+  app_env :enable_fdo, :astarte_pairing, :enable_fdo,
+    os_env: "PAIRING_ENABLE_FDO",
+    type: :boolean,
+    default: false
+
   @envdoc "The port the ingress is listening on, used for FDO authentication mechanism"
   app_env :base_url_port, :astarte_pairing, :base_url_port,
     os_env: "ASTARTE_BASE_URL_PORT",
-    type: :integer,
-    required: true
+    type: :integer
 
   @envdoc "The protocol the ingress is listening on, used for FDO authentication mechanism"
   app_env :base_url_protocol, :astarte_pairing, :base_url_protocol,
     os_env: "ASTARTE_BASE_URL_PROTOCOL",
-    type: BaseURLProtocol,
-    required: true
+    type: BaseURLProtocol
 
   @envdoc "The astarte base domain, used for FDO authentication mechanism"
   app_env :base_url_domain, :astarte_pairing, :base_url_domain,
     os_env: "ASTARTE_BASE_URL_DOMAIN",
-    type: :binary,
-    required: true
+    type: :binary
 
   @envdoc "URL to the running CFSSL instance for device certificate generation."
   app_env :cfssl_url, :astarte_pairing, :cfssl_url,
@@ -82,6 +88,15 @@ defmodule Astarte.Pairing.Config do
 
         {:error, _reason} ->
           raise "No CA certificate available."
+      end
+    end
+
+    if enable_fdo!() do
+      # check that all mandatory FDO variables are configured before starting
+      variables_to_check = [:base_url_port, :base_url_protocol, :base_url_domain]
+
+      if !Enum.all?(variables_to_check, &is_variable_set?(&1)) do
+        raise "FDO feature is enabled but not all its parameters are configured"
       end
     end
   end
@@ -130,4 +145,14 @@ defmodule Astarte.Pairing.Config do
   Credential requests made by devices are always authenticated, even it this is true.
   """
   def authentication_disabled?, do: disable_authentication!()
+
+  defp is_variable_set?(var_name) do
+    case apply(__MODULE__, var_name, []) do
+      {:ok, val} when not is_nil(val) ->
+        true
+
+      _ ->
+        false
+    end
+  end
 end
--- a/apps/astarte_pairing/lib/astarte_pairing_web/plug/fdo_gate.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing_web/plug/fdo_gate.ex
@@ -1,0 +1,39 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2026 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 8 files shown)


apps/astarte_pairing/config/dev.exs
```diff
@@ -71,6 +71,7 @@ config :astarte_pairing, :broker_url, "mqtts://broker.beta.astarte.cloud:8883/"
 
 config :astarte_pairing, :cfssl_url, "http://localhost:8888"
 
+# FDO feature disabled
```
noaccOS: I don't think we need this comment, it's the same behavior as prod (ie: needs the env)

apps/astarte_pairing/lib/astarte_pairing_web/plug/fdo_gate.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  alias Astarte.Pairing.Config
+
+  def init(_opts) do
+    nil
+  end
+
+  def call(conn, _opts) do
+    case Config.enable_fdo() do
+      {:ok, true} ->
```
noaccOS: variables with a default (or an env variable associated with them) always have a value, so it's safe to 


```suggestion
    case Config.enable_fdo!() do
      true ->
```

apps/astarte_pairing/lib/astarte_pairing_web/plug/fdo_gate.ex
```diff
@@ -0,0 +1,39 @@
...
+        conn
+
+      _ ->
+        conn
+        |> send_resp(404, "FDO endpoint not available")
```
noaccOS: let's not leak experimental values here


```suggestion
        |> send_resp(404, "")
```

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
```diff
@@ -276,4 +277,21 @@ defmodule Astarte.PairingWeb.FDOOnboardingControllerTest do
       assert {100, id} == assert_cbor_error(conn)
     end
   end
+
+  describe "FDO feature disabled" do
+    setup context do
+      Config.put_enable_fdo(false)
+      on_exit(fn -> Config.put_enable_fdo(true) end)
```
noaccOS: if you use mimic instead, you don't need to run the test file synchronously


```suggestion
      Config
      |> stub(:enable_fdo, fn -> {:ok, false})
```

apps/astarte_pairing/test/astarte_pairing_web/controllers/ownership_voucher_controller_test.exs
```diff
@@ -72,6 +73,23 @@ defmodule Astarte.PairingWeb.Controllers.OwnershipVoucherControllerTest do
     end
   end
 
+  describe "FDO feature disabled" do
+    setup :ownership
+
+    setup do
+      Config.put_enable_fdo(false)
+      on_exit(fn -> Config.put_enable_fdo(true) end)
```
noaccOS: same as above

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
frossq: Do we prefer using Mimic in these cases? (aka, avoiding using async: false wherever possible is always preferable?)

apps/astarte_pairing/config/dev.exs
frossq: :heavy_check_mark:

apps/astarte_pairing/lib/astarte_pairing_web/plug/fdo_gate.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/lib/astarte_pairing_web/plug/fdo_gate.ex
frossq: :heavy_check_mark:

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
frossq: :heavy_check_mark:

apps/astarte_pairing/test/astarte_pairing_web/controllers/ownership_voucher_controller_test.exs
frossq: :heavy_check_mark:

apps/astarte_pairing/test/astarte_pairing_web/controllers/fdo_onboarding_controller_test.exs
noaccOS: yes, if we can run the tests asynchronously it's always preferrable. in case it's something more complex than just an env var, we may provide additional ad-hoc tests, but here we don't need to do that imo

apps/astarte_pairing/lib/astarte_pairing/config.ex
```diff
@@ -84,6 +90,15 @@ defmodule Astarte.Pairing.Config do
...
       end
     end
+
+    if enable_fdo!() do
+      # check that all mandatory FDO variables are configured before starting
+      variables_to_check = [:base_url_port, :base_url_protocol, :base_url_domain]
+
+      if !Enum.all?(variables_to_check, &is_variable_set?(&1)) do
+        raise "FDO feature is enabled but not all its parameters are configured"
+      end
```
noaccOS, line 100: this hides from the user what variable is missing, which is not ideal
I guess it's ok for now as the feature is experimental, but we should revisit this before releasing the feature as stable

apps/astarte_pairing/lib/astarte_pairing/config.ex
```diff
@@ -130,4 +145,14 @@ defmodule Astarte.Pairing.Config do
   Credential requests made by devices are always authenticated, even it this is true.
   """
   def authentication_disabled?, do: disable_authentication!()
+
+  defp is_variable_set?(var_name) do
```
noaccOS, line 149: nit: credo would say something like: boolean functions should not start with `is_`, they should just end with a `?`

apps/astarte_pairing/lib/astarte_pairing/config.ex
```diff
@@ -130,4 +145,14 @@ defmodule Astarte.Pairing.Config do
...
+    case apply(__MODULE__, var_name, []) do
```
noaccOS, line 150: cursed

PR #1781, opened 2026-01-29
test(events): Astarte events triggers cache tests

nedimtokic:
Test Triggers Cache module

### DIFF
--- a/libs/astarte_events/test/astarte_events/triggers/cache_test.exs
+++ b/libs/astarte_events/test/astarte_events/triggers/cache_test.exs
@@ -1,0 +1,147 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2026 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


libs/astarte_events/test/astarte_events/triggers/cache_test.exs
```diff
@@ -0,0 +1,147 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Events.Triggers.CacheTest do
+  use Astarte.Cases.Data, async: false
+  use Astarte.Cases.Cache, async: true
```
noaccOS: in general the first one should have async: true and you don't need to specify it in the others

noaccOS (approved):
technically some files are missing copyright

libs/astarte_events/test/astarte_events/triggers/cache_test.exs
nedimtokic: Solved.

AliouneGaye21 (approved):

PR #1783, opened 2026-01-29
test(events): add tests for AMQP events

nedimtokic:
Add integration test to test when `dispatch_event` function sends message using AMQP event module.
Update Consumer helper module to be able to use default exchange name and not use virtual host when not necessary.

### DIFF
--- a/libs/astarte_events/test/support/helpers/consumer.ex
+++ b/libs/astarte_events/test/support/helpers/consumer.ex
@@ -29,11 +29,21 @@ defmodule Astarte.Events.Test.AmqpTriggers.Consumer do
             channel: AMQP.Channel.t() | nil,
             realm_name: String.t(),
             routing_key: String.t(),
+            virtual_host: String.t() | nil,
+            exchange_name: String.t() | nil,
             message_handler: (binary(), map() -> :ok | :error | {:error, term()}),
             ready_pid: pid() | nil
           }
 
-    defstruct [:channel, :realm_name, :routing_key, :message_handler, :ready_pid]
+    defstruct [
+      :channel,
+      :realm_name,
+      :routing_key,
+      :message_handler,
+      :virtual_host,
+      :ready_pid,
+      :exchange_name
+    ]
   end
 
   use GenServer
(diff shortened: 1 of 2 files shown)


libs/astarte_events/test/support/helpers/consumer.ex
```diff
@@ -130,14 +143,22 @@ defmodule Astarte.Events.Test.AmqpTriggers.Consumer do
...
 
     amqp_consumer_opts =
       Config.amqp_options!()
-      |> Keyword.put(:virtual_host, Vhost.vhost_name(realm_name))
+
+    amqp_consumer_opts =
+      case virtual_host do
+        nil -> amqp_consumer_opts
+        _ -> Keyword.put(amqp_consumer_opts, :virtual_host, virtual_host)
+      end
```
noaccOS: why do we need custom vhost logic?

libs/astarte_events/test/support/helpers/consumer.ex
```diff
@@ -51,12 +60,16 @@ defmodule Astarte.Events.Test.AmqpTriggers.Consumer do
     routing_key = Keyword.fetch!(opts, :routing_key)
     message_handler = Keyword.fetch!(opts, :message_handler)
     ready_pid = Keyword.fetch!(opts, :ready_pid)
+    exchange_name = Keyword.get(opts, :exchange_name, "astarte_events_#{realm_name}_events")
+    virtual_host = Keyword.get(opts, :virtual_host)
```
noaccOS:
```suggestion
    virtual_host = Keyword.get(opts, :virtual_host, Vhost.vhost_name(realm_name))
```

libs/astarte_events/test/support/helpers/consumer.ex
noaccOS:
```suggestion
    |> Keyword.put(:virtual_host, virtual_host)
```

libs/astarte_events/test/support/helpers/consumer.ex
nedimtokic: In moment of Init, `Astarte.DataAccess.Config `is not mocked so we have empty Astarte Instance ID and this will not work.

noaccOS: Then I suggest delaying the initalization until we `:start`

```elixir
def init(opts) do
  wait_start = Keyword.get(opts, :wait_start, false)

  if wait_start do
    {:ok, {:uninitialized, opts}}
  else
    {:ok, {:uninitialized, opts}, {:continue, :init}}
  end
end

def handle_call(:start, _from, state) do
  {:reply, :ok, state, {:continue, :init}}
end

def handle_continue(:init, {:uninitialized, opts}) do
  state = %State{} # init
  {:noreply, state, {:continue, :connect}}
end
```

PR #1784, opened 2026-01-30
test(events): add comprehensive tests for Core module

nedimtokic:
Add extensive unit tests for functions in Astarte.Events.Triggers.Core module.
Updated data helper to be able to generate simple trigger.

### DIFF
--- a/libs/astarte_events/test/astarte_events/triggers/core_test.exs
+++ b/libs/astarte_events/test/astarte_events/triggers/core_test.exs
@@ -1,9 +1,38 @@
-defmodule Astarte.Events.Triggers.CoreTest do
-  use ExUnit.Case, async: true
+#
+# This file is part of Astarte.
+#
+# Copyright 2026 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


libs/astarte_events/test/astarte_events/triggers/core_test.exs
```diff
@@ -80,4 +109,447 @@ defmodule Astarte.Events.Triggers.CoreTest do
...
+
+      assert {:ok, new_state} =
+               Core.load_trigger(
+                 realm_name,
+                 {:device_trigger, trigger},
+                 target,
+                 initial_state
+               )
+
+      assert Map.has_key?(new_state.device_triggers, :on_device_connection)
```
noaccOS: nit: this line is not needed as it's covered by the line below (you would get `nil` on the lhs)

libs/astarte_events/test/support/helpers/database_test_helper.ex
```diff
@@ -527,15 +525,22 @@ defmodule Astarte.Helpers.Database do
...
-    })
+    }
+  end
+
+  def install_simple_trigger(realm_name, opts \\ []) do
+    keyspace_name = Realm.keyspace_name(realm_name)
+
+    realm_name
+    |> generate_device_trigger(opts)
+    |> Ecto.Changeset.change()
```
noaccOS: nit: this line is not needed

```suggestion
```

PR #1787, opened 2026-02-04
refactor(events): fix credo complains

nedimtokic:
Refactor functions with too many parameters
Add module documentation where necessary
Refactor too deep functions
Added MissedMetadataKeyInLoggerConfig to the credo config to avoid credo warnings for Logger

### DIFF
--- a/libs/astarte_events/lib/astarte_events/triggers/data_trigger_context.ex
+++ b/libs/astarte_events/lib/astarte_events/triggers/data_trigger_context.ex
@@ -1,7 +1,7 @@
 #
 # This file is part of Astarte.
 #
-# Copyright 2026 SECO Mind srl
+# Copyright 2026 SECO Mind Srl
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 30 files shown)


libs/astarte_events/lib/astarte_events/triggers/data_trigger_context.ex
```diff
@@ -0,0 +1,46 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Events.Triggers.DataTriggerContext do
+  @moduledoc """
+  Module representing the context of a data trigger event in Astarte Events.
+  """
+  @type t :: %__MODULE__{
```
noaccOS: nit: in general let's use `typedstruct` to avoid defining the same information twice

libs/astarte_events/lib/astarte_events/triggers/triggers.ex
```diff
@@ -147,80 +151,51 @@ defmodule Astarte.Events.Triggers do
...
-        device_id,
-        groups,
-        event_key,
-        path_tokens,
-        value,
-        data
-      )
-    end)
-    |> Enum.concat()
+    case query.path do
```
noaccOS: if we want to make this a single function, I suggest abstracting only on the fetch function instead of all the logic


```elixir
find_triggers =
  case query.path do
    nil -> 
      fn event_key -> Cache.find_data_trigger_targets(realm_name, device_id, groups, event_key, data) end

    path ->
      path_tokens = path |> String.split("/") |> Enum.drop(1)
      fn event_key -> 
          Cache.find_data_trigger_targets(
            realm_name,
            device_id,
            groups,
            event_key,
            path_tokens,
            query.value,
            data
          )
      end
  end

event_keys
|> Enum.map(find_triggers)
|> Enum.concat()
```
  
this makes it easier to follow the actual steps of the function

.credo.exs
```diff
@@ -19,7 +19,8 @@
       checks: [
         {Credo.Check.Design.TagFIXME, [priority: :normal, exit_status: 0]},
         {Credo.Check.Design.TagTODO, [priority: :normal, exit_status: 0]},
-        {Credo.Check.Readability.MaxLineLength, [priority: :low, max_length: 100]}
+        {Credo.Check.Readability.MaxLineLength, [priority: :low, max_length: 100]},
+        {Credo.Check.Warning.MissedMetadataKeyInLoggerConfig, false}
```
noaccOS: why are we removing this check?

nedimtokic: In libraries we do not have configuration for logger and Credo reports:
`Logger metadata key tag not found in Logger config.`

noaccOS: Let's add the configuration then!

nedimtokic: Done.

libs/astarte_events/lib/astarte_events/triggers/triggers.ex
nedimtokic: This function is refactored because it had too many parameters. Suggestion applied.

PR #1789, opened 2026-02-10
test(dup): fix flakyness

mizzet1:
This PR add fixes to errors detected during a 100 mix tests run to check flakyness in the dup service

### DIFF
--- a/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
+++ b/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
@@ -22,6 +22,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
   use ExUnitProperties
   use Mimic
 
+  @moduletag timeout: 180_000
+
   import Astarte.Helpers.DataUpdater
   import Astarte.InterfaceUpdateGenerators
   import Ecto.Query
@@ -101,9 +103,16 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
       %{
         state: state,
         interfaces: interfaces,
+        interface_descriptors: interface_descriptors,
         realm_name: realm_name,
         device: device
       } = context
+
+      descriptors_map =
+        interface_descriptors
+        |> Map.new(fn desc ->
+          {desc.interface_id, desc}
+        end)
 
       valid_interfaces =
         interfaces
@@ -131,7 +140,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
                     interface.name <> mapping_update.path
                   end) do
         Enum.each(mapping_updates_list, fn {interface, mapping_update} ->
-          Database.insert_values(realm_name, device, interface, [mapping_update])
+          descriptor = Map.fetch!(descriptors_map, interface.interface_id)
+          Database.insert_values(realm_name, device, interface, descriptor, [mapping_update])
         end)
 
         :ok = Core.Device.prune_device_properties(state, decoded_payload, DateTime.utc_now())
--- a/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/interface_test.exs
+++ b/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/interface_test.exs
@@ -123,16 +123,25 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.InterfaceTest do
       %{
         state: state,
         interfaces: interfaces,
+        interface_descriptors: interface_descriptors,
         realm_name: realm_name,
         device: device
       } = context
+
+      descriptors_map =
+        interface_descriptors
+        |> Map.new(fn desc ->
+          {desc.interface_id, desc}
+        end)
 
       valid_interfaces =
         interfaces |> Enum.filter(&(&1.type == :properties and &1.ownership == :device))
 
       check all interface <- member_of(valid_interfaces),
                 mapping_update <- valid_mapping_update_for(interface) do
-        Helpers.Database.insert_values(realm_name, device, interface, [mapping_update])
+        descriptor = Map.fetch!(descriptors_map, interface.interface_id)
+
+        Helpers.Database.insert_values(realm_name, device, interface, descriptor, [mapping_update])
 
         keyspace = Realm.keyspace_name(realm_name)
 
--- a/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs
+++ b/apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs
@@ -38,7 +38,7 @@ defmodule Astarte.DataUpdaterPlant.PayloadsDecoderTest do
     assert PayloadsDecoder.decode_bson_payload(unset_payload, timestamp) == {nil, nil, nil}
   end
 
-  property "individual value payloads without metadata and without timestamp" do
+  property "individual value payloads optional metadata and timestamp" do
     timestamp = decimicrosecond_timestamp() |> Enum.at(0)
     expected_timestamp = timestamp |> div(10_000)
 
@@ -59,11 +59,36 @@ defmodule Astarte.DataUpdaterPlant.PayloadsDecoderTest do
     end
   end
 
+  test "metadata keys with null byte cannot be preserved" do
+    payload = %{"v" => true, "m" => %{"bad\0key" => "x"}}
+    reception_timestamp = 0
+
+    case Cyanide.encode(payload) do
+      {:ok, encoded} ->
+        assert PayloadsDecoder.decode_bson_payload(encoded, reception_timestamp) ==
+                 {true, reception_timestamp, %{}}
+    end
+  end
+
+  test "metadata keys without null byte are preserved" do
+    payload = %{"v" => true, "m" => %{"good_key" => "x"}}
+    reception_timestamp = 0
+
+    case Cyanide.encode(payload) do
+      {:ok, encoded} ->
+        assert PayloadsDecoder.decode_bson_payload(encoded, reception_timestamp) ==
+                 {true, reception_timestamp, %{"good_key" => "x"}}
+    end
+  end
+
   test "deprecated object aggregation" do
     timestamp = decimicrosecond_timestamp() |> Enum.at(0)
     expected_timestamp = timestamp |> div(10_000)
 
-    object_payload = object_astarte_value() |> Enum.at(0)
+    object_payload =
+      object_astarte_value()
+      |> filter(&(not Map.has_key?(&1, "v")))
+      |> Enum.at(0)
 
     assert PayloadsDecoder.decode_bson_payload(Cyanide.encode!(object_payload), timestamp) ==
              {object_payload, expected_timestamp, %{}}
@@ -173,7 +198,7 @@ defmodule Astarte.DataUpdaterPlant.PayloadsDecoderTest do
   end
 
   defp object_astarte_value do
-    map_of(string(:alphanumeric), individual_astarte_value())
+    map_of(string(:alphanumeric, min_length: 1), individual_astarte_value())
   end
 
   defp astarte_value do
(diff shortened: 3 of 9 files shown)


apps/astarte_data_updater_plant/test/support/helpers/database.ex
```diff
@@ -26,18 +26,18 @@ defmodule Astarte.Helpers.Database do
   alias Astarte.DataUpdaterPlant.DataUpdater.Queries
 
   @create_keyspace """
-  CREATE KEYSPACE :keyspace
+  CREATE KEYSPACE IF NOT EXISTS :keyspace
```
noaccOS: let's remove these `IF NOT EXISTS` on creates, we want to fail if the database was setup before tests.
it's perfectly fine to leave the `IF EXISTS` on delete to account for failures during the setup process

apps/astarte_data_updater_plant/test/support/helpers/database.ex
```diff
@@ -511,12 +521,9 @@ defmodule Astarte.Helpers.Database do
     |> Mimic.stub(:astarte_instance_id!, fn -> astarte_instance_id end)
   end
 
-  def insert_values(realm_name, device, interface, mapping_updates) do
+  def insert_values(realm_name, device, interface, interface_descriptor, mapping_updates) do
     mappings_map = interface.mappings |> Map.new(&{&1.endpoint_id, &1})
 
-    {:ok, interface_descriptor} =
-      Interface.fetch_interface_descriptor(realm_name, interface.name, interface.major_version)
```
noaccOS, line 522: why should this lead to a failure? did you investigate the reason the descriptor wasn't in the database when it should be?

apps/astarte_data_updater_plant/test/support/database_test_helper.ex
```diff
@@ -363,8 +363,15 @@ defmodule Astarte.DataUpdaterPlant.DatabaseTestHelper do
   def create_test_keyspace(realm_name) do
     keyspace_name = Realm.keyspace_name(realm_name)
 
-    case execute(keyspace_name, @create_autotestrealm) do
-      {:ok, _} ->
+    result = execute(keyspace_name, @create_autotestrealm)
+
+    case result do
+      {:ok, _} -> :ok
+      {:error, %Xandra.Error{reason: :already_exists}} -> :ok
```
noaccOS: see my comment above regarding already existing database tables and keyspaces

apps/astarte_data_updater_plant/test/support/cases/device.ex
```diff
@@ -65,10 +65,17 @@ defmodule Astarte.Cases.Device do
...
       realm_name: realm_name,
       interfaces: interfaces,
+      interface_descriptors: interface_descriptors,
       server_property_with_all_endpoint_types: server_property_with_all_endpoint_types,
       device: device
     } = context
 
+    descriptors_map =
+      interface_descriptors
+      |> Enum.into(%{}, fn desc ->
```
noaccOS: nit: more explicit


```suggestion
      |> Map.new(fn desc ->
```

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/queries_test.exs
```diff
@@ -416,6 +417,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.QueriesTest do
     setup %{realm_name: realm_name} do
       result = populate_deletion_in_progress(realm_name)
 
+      validate_deletion_in_progress_entries(realm_name, result)
```
noaccOS, line 420: I would expect to not do anything really but it's fine to leave it in case new errors pop up in the future

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs
```diff
@@ -173,7 +175,11 @@ defmodule Astarte.DataUpdaterPlant.PayloadsDecoderTest do
   end
 
   defp object_astarte_value do
-    map_of(string(:alphanumeric), individual_astarte_value())
+    non_reserved_key =
+      string(:alphanumeric, min_length: 1)
+      |> filter(&(&1 not in ["v", "t", "m"]))
+
```
noaccOS: these are perfectly fine keys for object values, the logic for valid_result may be flawed instead

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs
```diff
@@ -38,21 +38,23 @@ defmodule Astarte.DataUpdaterPlant.PayloadsDecoderTest do
...
     check all payload <- decoded_payload() do
       encoded_payload = Cyanide.encode!(payload)
-      value = Map.fetch!(payload, "v")
+
+      {:ok, decoded_payload} = Cyanide.decode(encoded_payload)
+      value = Map.fetch!(decoded_payload, "v")
 
       expected_timestamp =
-        case Map.fetch(payload, "t") do
+        case Map.fetch(decoded_payload, "t") do
```
noaccOS: so if `payload` is already a decoded payload, why does encoding it and decoding it back solve issues?

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs
```diff
@@ -38,21 +38,23 @@ defmodule Astarte.DataUpdaterPlant.PayloadsDecoderTest do
     assert PayloadsDecoder.decode_bson_payload(unset_payload, timestamp) == {nil, nil, nil}
   end
 
-  property "individual value payloads without metadata and without timestamp" do
+  property "individual value payloads with optional metadata and with timestamp" do
```
noaccOS:
```suggestion
  property "individual value payloads with optional metadata and timestamp" do
```

the timestamp is not mandatory

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/interface_test.exs
```diff
@@ -122,16 +122,25 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.InterfaceTest do
...
         state: state,
         interfaces: interfaces,
+        interface_descriptors: interface_descriptors,
         realm_name: realm_name,
         device: device
       } = context
 
+      descriptors_map =
+        interface_descriptors
+        |> Enum.into(%{}, fn desc ->
```
noaccOS:
```suggestion
        |> Map.new(fn desc ->
```

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/device_test.exs
```diff
@@ -100,10 +102,17 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DeviceTest do
...
         state: state,
         interfaces: interfaces,
+        interface_descriptors: interface_descriptors,
         realm_name: realm_name,
         device: device
       } = context
 
+      descriptors_map =
+        interface_descriptors
+        |> Enum.into(%{}, fn desc ->
```
noaccOS:
```suggestion
        |> Map.new(fn desc ->
```

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs
mizzet1: Changed it so the current logic only reserve the "v" key for value, and the filter logic is not placed directly in the associated "deprecated" test

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs
mizzet1: This was added to check for errors in the decoding/encoding process but the logic now works as the change of reserving "v" as key used by the decoder to match the different types of paylod fixed the decoder related issues.

In particular this kind of error:
1) test deprecated object aggregation (Astarte.DataUpdaterPlant.PayloadsDecoderTest)
     test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs:64
     Assertion with == failed
     code:  assert PayloadsDecoder.decode_bson_payload(Cyanide.encode!(object_payload), timestamp) ==
              {object_payload, expected_timestamp, %{}}
     left:  {[13527883], 1284657573, %{}}
     right: {%{"L" => [false], "v" => [13527883]}, 1284657573, %{}}

apps/astarte_data_updater_plant/test/support/helpers/database.ex
mizzet1: This change was added because the interface_descriptors are now passed as argument in the tests instead of being fetched every time from the db.
To verify if the descriptors are present in the database, a check has been added in the get_interface_descriptors function, which now raise an exception if they are not present.

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs
```diff
@@ -173,7 +176,7 @@ defmodule Astarte.DataUpdaterPlant.PayloadsDecoderTest do
   end
 
   defp object_astarte_value do
-    map_of(string(:alphanumeric), individual_astarte_value())
+    map_of(string(:alphanumeric, min_length: 1), individual_astarte_value(), min_length: 2)
```
noaccOS: Empty values are valid values


```suggestion
    map_of(string(:alphanumeric, min_length: 1), individual_astarte_value())
```

apps/astarte_data_updater_plant/test/support/database_test_helper.ex
```diff
@@ -363,8 +363,14 @@ defmodule Astarte.DataUpdaterPlant.DatabaseTestHelper do
...
-    case execute(keyspace_name, @create_autotestrealm) do
-      {:ok, _} ->
+    result = execute(keyspace_name, @create_autotestrealm)
+
+    case result do
+      {:ok, _} -> :ok
+      {:error, error} -> {:error, error}
+    end
+    |> case do
+      :ok ->
```
noaccOS: this should be reverted

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs
```diff
@@ -173,17 +176,17 @@ defmodule Astarte.DataUpdaterPlant.PayloadsDecoderTest do
...
   defp astarte_value do
     one_of([individual_astarte_value(), object_astarte_value()])
   end
 
-  defp decoded_payload do
+  defp decoded_payload() do
     value = astarte_value()
     timestamp = map(TimestampGenerator.timestamp(), &DateTime.from_unix!/1)
-    metadata = map_of(string(:utf8), string(:utf8))
+    metadata = map_of(string(:alphanumeric, min_length: 1), string(:utf8, min_length: 1))
```
noaccOS, line 211: why is the key changed from utf8 to alphanumeric? also the value probably doesn't need min_length 1

apps/astarte_data_updater_plant/test/support/helpers/device.ex
```diff
@@ -42,13 +40,21 @@ defmodule Astarte.Helpers.Device do
   ]
 
   def insert_interface_cleanly(realm_name, interface) do
-    keyspace = Realm.keyspace_name(realm_name)
-    interface_db = %Interface{name: interface.name, major_version: interface.major_version}
+    capture_log(fn ->
+      Core.delete_interface(realm_name, interface.name, interface.major_version)
+    end)
```
noaccOS: as discussed, let's revert this and add more strict interface generation to the device case instead

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/payloads_decoder_test.exs
mizzet1: map_of(string(:utf8), string(:utf8)) allows for keys with name containing NUL (\x00) or empty string, while names of BSON fields are c-string which can't contain NUL. 
source: https://bsonspec.org/spec.html#:~:text=8%20encoded%20characters.-,cstring,NOT%20contain%20unsigned_byte(0)%2C%20it%20is%20not%20full%20UTF%2D8.,-binary

the two added tests in payload_decoder confirm this

PR #1796, opened 2026-02-12
 refactor(dup): fix credo complains

nedimtokic:
Add module documentation where necessary
    Order aliases
    Split functions with too many parameters
    Replace unnecessary with clause
    Split too deep functions


apps/astarte_data_updater_plant/config/test.exs
```diff
@@ -5,7 +5,7 @@ config :astarte_data_updater_plant, :amqp_consumer_options,
 
 config :logger, :console,
   format: {PrettyLog.UserFriendlyFormatter, :format},
-  metadata: [:realm, :device_id, :function]
+  metadata: [:realm, :device_id, :ip_address, :module, :function, :tag]
```
noaccOS: would it be possible here to apply the same strategy we used for realm management?
ie: the metadata is set on `config/config.exs`, while env config files (`config/test.exs`, `config/prod.exs`, ...) only set the format

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/interface.ex
```diff
@@ -130,34 +130,68 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.Interface do
...
+    |> Enum.each(fn path ->
+      maybe_prune_path(
+        state,
+        interface_descriptor,
+        endpoint_id,
+        path,
+        all_paths_set,
+        context
+      )
+    end)
```
noaccOS: nit: let's be more declarative

```suggestion
    database_paths =
      Queries.all_device_owned_property_endpoint_paths!(
        state.realm,
        state.device_id,
        interface_descriptor,
        endpoint_id
      )

    interface_name = interface_descriptor.name
    paths = database_paths |> Enum.map(&{interface_name, &1}) |> MapSet.new()
    unset_paths = MapSet.difference(paths, all_paths_set)

    unset_path
    |> Enum.each(fn path ->
      # not maybe
      prune_path(
        state,
        interface_descriptor,
        endpoint_id,
        path,
        all_paths_set,
        context
      )
    end)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/event_type_utils.ex
```diff
@@ -41,39 +58,6 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.EventTypeUtils do
...
-      :DEVICE_REGISTERED ->
-        :on_device_registered
-
-      :DEVICE_DELETION_STARTED ->
-        :on_device_deletion_started
-
-      :DEVICE_DELETION_FINISHED ->
-        :on_device_deletion_finished
-    end
+    Map.fetch!(@device_event_type_map, device_event_type)
```
noaccOS: we could reuse the functions from astarte_events instead!

btw, do you know why credo is complaining here but not for `pretty_data_trigger_type`?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/insert_context.ex
```diff
@@ -0,0 +1,40 @@
...
+defmodule Astarte.DataUpdaterPlant.DataUpdater.InsertContext do
+  @moduledoc false
+
+  alias Astarte.Core.InterfaceDescriptor
+  alias Astarte.Core.Mapping
+
+  use TypedStruct
+
+  typedstruct do
+    field :realm, term(), enforce: true
```
noaccOS: - we don't need to use `enforce` on all fields (and I think it doesn't do anything there was an issue on typedstruct)
- let's try to use actual types instead of `term`
  - realm is a `String.t()`
  - device_id is probably a `Astarte.DataAccess.UUID` (if it's not expecting the decoded version)
  - path is a string
  - the timestamps (unfortunately) are integers

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -78,18 +82,17 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
-        %InterfaceDescriptor{storage_type: :multi_interface_individual_properties_dbtable} =
-          interface_descriptor,
-        %Mapping{allow_unset: true} = mapping,
-        path,
-        nil,
-        _value_timestamp,
-        _reception_timestamp,
-        opts
-      ) do
+  def insert_value_into_db(%InsertContext{
```
noaccOS: usually it's better for readability to only pattern match on the function header what is needed for choosing the body

so for instance here you could write it like


```elixir
  # if you're not choosing the body of the function based on the struct types, you can add them for clarity but they're not mandatory
  def insert_value_into_db(%{interface_descriptor: %{storage_type: :multi_individual_properties_dtable}, mapping: %Mapping{allow_unset: true}, value: nil} = context)
    %InsertContext{
      # here you can assign other fields from the context
    } = context
```

apps/astarte_data_updater_plant/test/support/interface_update_generators.ex
```diff
@@ -257,24 +261,20 @@ defmodule Astarte.InterfaceUpdateGenerators do
...
-      :doublearray -> invalid_list_or(invalid_number())
-      :integerarray -> invalid_list_or(invalid_number())
-      :booleanarray -> invalid_list_or(invalid_boolean())
-      :longintegerarray -> invalid_list_or(invalid_number())
-      :stringarray -> invalid_list_or(invalid_string())
-      :binaryblobarray -> invalid_list_or(invalid_binary())
-      :datetimearray -> invalid_list_or(invalid_datetime())
-    end
-  end
+  def invalid_type(:double), do: invalid_number()
```
noaccOS, line 264: just a question: here we're splitting the match on the signature, while for the trigger type you opted for a map
why is that?

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/event_type_utils.ex
nedimtokic: Credo error for refactored function:
`Function is too complex (cyclomatic complexity is 11, max is 9).`

nedimtokic: This module is unused, so I will remove it.

apps/astarte_data_updater_plant/config/test.exs
nedimtokic: Done.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/insert_context.ex
nedimtokic: Updated struct with types.

apps/astarte_data_updater_plant/test/support/interface_update_generators.ex
nedimtokic: No strong reason. Making trigger type for matching would require more code, because it has more parameters.
This is helper module for tests so here I wanted only to solve credo complains on simple way.

noaccOS (approved):

PR #1798, opened 2026-02-13
refactor(hk): `astarte_generators` to mono-repo version

shinnokdisengir:
- changed dep

refactoring files:
- `realms_test`

### DIFF
--- a/apps/astarte_housekeeping/test/astarte_housekeeping/realms/realms_test.exs
+++ b/apps/astarte_housekeeping/test/astarte_housekeeping/realms/realms_test.exs
@@ -1,7 +1,7 @@
 #
 # This file is part of Astarte.
 #
-# Copyright 2017 - 2025 SECO Mind Srl
+# Copyright 2017 - 2026 SECO Mind Srl
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


apps/astarte_housekeeping/test/astarte_housekeeping/realms/realms_test.exs
```diff
@@ -17,14 +17,16 @@
 #
 
 defmodule Astarte.Housekeeping.RealmsTest do
-  use Astarte.Housekeeping.DataCase, async: true
   use ExUnitProperties
-  use Mimic
+  use Astarte.Housekeeping.DataCase, async: true
 
+  import Mimic
```
noaccOS: let's keep the `use Mimic` instead. it's equivalent to `import Mimic` + `verify_on_exit`: https://github.com/edgurgel/mimic/blob/main/lib/mimic.ex#L89

If you simply `import` it failed expectations won't result in errors

apps/astarte_housekeeping/test/astarte_housekeeping/realms/realms_test.exs
```diff
@@ -17,14 +17,16 @@
 #
 
 defmodule Astarte.Housekeeping.RealmsTest do
-  use Astarte.Housekeeping.DataCase, async: true
   use ExUnitProperties
-  use Mimic
```
noaccOS, line 22: this is still relevant and should be kept, otherwise the expectation at line 137 will not make the test fail if it is not called

apps/.DS_Store
noaccOS, line 1: tell me you use a mac without telling me you use a mac

PR #1800, opened 2026-02-16
refactor(pairing): use the guid as primary key for vouchers

noaccOS:
originally, we had decided to allow multiple ownership vouchers for a single device guid, with the motivation being that the rendezvous server was supposed to be in charge of deciding the validity of ownership vouchers and lift us from that job.

in practice, we implemented our application logic as if there was always a single voucher for a guid, by simply adding `limit: 1` to all our queries. this also added complexity and was investigated as a possible source of bugs multiple times.

for our implementation, supporting multiple vouchers for each guid doesn't really make sense anymore, as our vouchers are short-lived anyway, so let's just allow a single voucher for a guid!

### DIFF
--- a/apps/astarte_housekeeping/test/support/helpers/database.ex
+++ b/apps/astarte_housekeeping/test/support/helpers/database.ex
@@ -72,8 +72,8 @@ defmodule Astarte.Housekeeping.Helpers.Database do
   CREATE TABLE :keyspace.ownership_vouchers (
       private_key blob,
       voucher_data blob,
-      device_id uuid,
-      PRIMARY KEY (device_id, voucher_data)
+      guid blob,
+      PRIMARY KEY (guid)
    );
   """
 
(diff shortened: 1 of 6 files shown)


AliouneGaye21 (approved):

apps/astarte_housekeeping/test/support/helpers/database.ex
```diff
@@ -72,8 +72,8 @@ defmodule Astarte.Housekeeping.Helpers.Database do
   CREATE TABLE :keyspace.ownership_vouchers (
       private_key blob,
       voucher_data blob,
-      device_id uuid,
-      PRIMARY KEY (device_id, voucher_data)
+      guid uuid,
+      PRIMARY KEY (guid, voucher_data)
```
nedimtokic:
```suggestion
      guid blob,
      PRIMARY KEY (guid)
```

nedimtokic (changes requested):

nedimtokic (approved):

PR #1823, opened 2026-02-26
test(data_access): add comprehensive tests

nedimtokic:
Add missing tests for Astarte Data Access modules.

### DIFF
--- a/libs/astarte_data_access/test/c_system_test.exs
+++ b/libs/astarte_data_access/test/c_system_test.exs
@@ -1,0 +1,109 @@
+#
+# This file is part of Astarte.
+#
+# Copyright 2026 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


libs/astarte_data_access/test/c_system_test.exs
```diff
@@ -0,0 +1,109 @@
...
+
+  setup_all do
+    Xandra.Cluster.run(:astarte_data_access_xandra, fn conn ->
+      DatabaseTestHelper.create_test_keyspace(conn)
+    end)
+
+    on_exit(fn ->
+      Xandra.Cluster.run(:astarte_data_access_xandra, fn conn ->
+        DatabaseTestHelper.destroy_local_test_keyspace(conn)
+      end)
```
noaccOS, line 42: nit: in general, it's better to place the on_exit callbacks _before_ the setup so that if you happen to crash during creation (after you have already set up some resources), the on_exit still runs and cleans things up

libs/astarte_data_access/test/c_system_test.exs
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  describe "run_with_schema_agreement/2" do
+    test "wraps a schema DDL query" do
+      result =
+        CSystem.run_with_schema_agreement(fn ->
+          Repo.query(@create_table, [])
+        end)
+
+      assert match?({:ok, _}, result)
```
noaccOS: nit: using the default assertion match syntax better highlights differences in case of a mismatch (instead of just `false`)


```suggestion
      assert {:ok, _} = result
```

PR #1843, opened 2026-03-09
feat(pairing): Add rendezvous server SSL configuration options

nedimtokic:
SSL for FDO Rendezvous Server is disabled by default and can be enabled via environment variables.

New environment variables introduced:
- PAIRING_FDO_RENDEZVOUS_SSL_ENABLED: enables SSL (default: false)
- PAIRING_FDO_RENDEZVOUS_SSL_CA_FILE: path to a custom CA certificate
  file; when not set, the bundled cURL certificate bundle is used
- PAIRING_FDO_RENDEZVOUS_SSL_DISABLE_SNI: disables Server Name
  Indication (default: false)
- PAIRING_FDO_RENDEZVOUS_SSL_CUSTOM_SNI: custom SNI hostname; when not
  set, the value of PAIRING_FDO_RENDEZVOUS_HOST is used

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/config.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/config.ex
@@ -80,6 +80,71 @@ defmodule Astarte.Pairing.Config do
     type: :binary,
     default: "http://rendezvous:8041"
 
+  @envdoc "Enable SSL for the FDO Rendezvous Server connection. If not specified, SSL is disabled."
+  app_env :fdo_rendezvous_ssl_enabled, :astarte_pairing, :fdo_rendezvous_ssl_enabled,
+    os_env: "PAIRING_FDO_RENDEZVOUS_SSL_ENABLED",
+    type: :boolean,
+    default: false
+
+  @envdoc "Path to the CA certificate file for the FDO Rendezvous Server TLS connection. When not specified, the bundled cURL certificate bundle will be used."
+  app_env :fdo_rendezvous_ssl_ca_file, :astarte_pairing, :fdo_rendezvous_ssl_ca_file,
+    os_env: "PAIRING_FDO_RENDEZVOUS_SSL_CA_FILE",
+    type: :binary,
+    default: CAStore.file_path()
+
+  @envdoc "Disable FDO Rendezvous Server Name Indication. Defaults to false."
+  app_env :fdo_rendezvous_ssl_disable_sni,
+          :astarte_pairing,
+          :fdo_rendezvous_ssl_disable_sni,
+          os_env: "PAIRING_FDO_RENDEZVOUS_SSL_DISABLE_SNI",
+          type: :boolean,
+          default: false
+
+  @envdoc "Specify the hostname to be used in TLS Server Name Indication extension. If not specified, the FDO Rendezvous Server host will be used. This value is used only if Server Name Indication is enabled."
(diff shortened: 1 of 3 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/rendezvous/client.ex
```diff
@@ -35,9 +35,14 @@ defmodule Astarte.Pairing.FDO.Rendezvous.Client do
...
-      # ssl: Config.ssl_options!() no ssl?
-    ]
+    ssl_opts = Config.fdo_rendezvous_ssl_options!()
+
+    auth_opts =
+      if ssl_opts != [] do
+        [ssl: ssl_opts]
+      else
+        []
+      end
```
noaccOS: I think it's fine to just have 


```suggestion
    auth_opts = [
      ssl: Config.fdo_rendezvous_ssl_options!()
    ]
```

we have something equivalent in `Astarte.Events.AMQP`

apps/astarte_pairing/lib/astarte_pairing/config.ex
```diff
@@ -80,6 +80,38 @@ defmodule Astarte.Pairing.Config do
...
+  """
+  def fdo_rendezvous_ssl_options! do
+    if fdo_rendezvous_ssl_enabled!() do
+      build_rendezvous_ssl_options()
+    else
+      []
+    end
+  end
+
+  defp build_rendezvous_ssl_options do
```
noaccOS, line 120: I think we're missing sni options

apps/astarte_pairing/lib/astarte_pairing/fdo/rendezvous/client.ex
nedimtokic: Changed.

apps/astarte_pairing/lib/astarte_pairing/config.ex
nedimtokic: Added SNI.

apps/astarte_pairing/lib/astarte_pairing/config.ex
```diff
@@ -80,6 +80,67 @@ defmodule Astarte.Pairing.Config do
...
+  @envdoc "Specify the hostname to be used in TLS Server Name Indication extension. If not specified, the FDO Rendezvous Server host will be used. This value is used only if Server Name Indication is enabled."
+  app_env :fdo_rendezvous_ssl_custom_sni, :astarte_pairing, :fdo_rendezvous_ssl_custom_sni,
+    os_env: "PAIRING_FDO_RENDEZVOUS_SSL_CUSTOM_SNI",
+    type: :binary
+
+  @envdoc "The host for the FDO Rendezvous Server."
+  app_env :fdo_rendezvous_host, :astarte_pairing, :fdo_rendezvous_host,
+    os_env: "PAIRING_FDO_RENDEZVOUS_HOST",
+    type: :binary,
+    default: "localhost"
```
noaccOS: I don't think having rendezvous url + rendezvous host is great ux, either we change the config to use url + protocol + port (but that would be breaking AND a lot of work) or you could [copy my homework](https://github.com/astarte-platform/astarte/blob/master/apps/astarte_pairing/lib/astarte_pairing/config.ex#L139)

PR #1849, opened 2026-03-12
feat(dup): Validate required interface mapping

nedimtokic:
This PR introduces support for the `required` flag on object interface mapping endpoints. 
When a mapping has `required: true`, incoming data must include that endpoint's key — otherwise the message is rejected.

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
@@ -24,6 +24,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
   """
   alias Astarte.Core.Device
   alias Astarte.Core.InterfaceDescriptor
+  alias Astarte.Core.Mapping
   alias Astarte.Core.Mapping.ValueType
   alias Astarte.DataAccess.Data
   alias Astarte.DataUpdaterPlant.DataUpdater.Cache
@@ -63,7 +64,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
          :ok <- can_write_on_interface?(context, interface_descriptor.ownership),
          {:ok, mapping} <- resolve_path(context, interface_descriptor),
          {:ok, {value, value_timestamp, _metadata}} <- decode_bson_payload(context),
-         :ok <- validate_value_type(context, interface_descriptor, mapping, value) do
+         :ok <- validate_value(context, interface_descriptor, mapping, value) do
       maybe_explicit_value_timestamp =
         if mapping.explicit_timestamp,
           do: value_timestamp,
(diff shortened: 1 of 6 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -63,7 +63,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
          :ok <- can_write_on_interface?(context, interface_descriptor.ownership),
          {:ok, mapping} <- resolve_path(context, interface_descriptor),
          {:ok, {value, value_timestamp, _metadata}} <- decode_bson_payload(context),
-         :ok <- validate_value_type(context, interface_descriptor, mapping, value) do
+         :ok <- validate_value_type(context, interface_descriptor, mapping, value),
+         :ok <- validate_required_mappings(context, interface_descriptor, value) do
```
noaccOS: I suggest having a single `validate_value` which

1. calculates the `%{key => mapping}` map (basically modifying `Core.Interface.extract_expected_types` to return mappings instead of value types)
2. validate the value type using the `.value_type` key
3. validate required mappings uing the `required` key

```elixir
defp validate_value(context, interface_descriptor, mapping, value) do
  ...
  mappings = Core.Interface.extract_mappings(path, interface_descriptor, mapping, state.mappings)

  with :ok <- validate_value_type(mappings, value) do
    validate_required_mappings(interface_descriptor, mappings, value)
  end
end
```

this way we're only performing the heavy operation once and we reduce code size a bit

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/interface.ex
```diff
@@ -321,6 +321,23 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.Interface do
     end
   end
 
+  def extract_mappings(%InterfaceDescriptor{aggregation: :individual}, mapping, _mappings) do
```
noaccOS, line 314: are we still using `extract_expected_types`? if not I suggest removing it (and maybe adapting its tests for this new function)

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/interface.ex
```diff
@@ -321,6 +321,23 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.Interface do
...
+        %InterfaceDescriptor{aggregation: :object},
+        _mapping,
+        mappings
+      ) do
+    mappings
+    |> Enum.flat_map(fn {_id, m} ->
+      key = m.endpoint |> String.split("/") |> List.last()
+      [{key, m}]
+    end)
+    |> Map.new()
```
noaccOS:
```suggestion
    |> Map.new(fn {_id, m} ->
      key = m.endpoint |> String.split("/") |> List.last()
      {key, m}
    end)
```

a flat_map is useful when you return a variable number of items for each iteration, this is always 1:1 so a `map` works just as well

`Map.new` can be used with a transform function so the Enum.map is "implicit"

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/interface.ex
nedimtokic: Removed.

PR #1854, opened 2026-03-16
feat(appengine): Validate required mappings

nedimtokic:
If interface has required mapping, validate it before updating interface data

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
@@ -361,6 +361,33 @@ defmodule Astarte.AppEngine.API.Device do
     end
   end
 
+  defp validate_value(mappings_by_key, value) when is_map(value) do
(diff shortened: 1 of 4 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -379,9 +406,11 @@ defmodule Astarte.AppEngine.API.Device do
          {:ok, endpoint} <-
            resolve_object_aggregation_path(path, interface_descriptor, mappings),
          endpoint_id <- endpoint.endpoint_id,
-         expected_types <- extract_expected_types(mappings),
+         mappings_by_key <- extract_mappings(mappings),
+         expected_types <-
```
noaccOS: let's use `=` when we have simple assignments in with blocks, to signal that we always match

```suggestion
         mappings_by_key = extract_mappings(mappings),
         expected_types =
```

PR #1856, opened 2026-03-17
chore: add namespace functionality to openbao

noaccOS:
add two functions:
- create_namespace/3, which creates the nested key namespace
- list_namespaces/0, which lists all namespaces, and is used in tests

the client now always points to the "/v1" api


apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
```diff
@@ -20,4 +20,111 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
...
+    else
+      _ -> {:error, {:invalid_response_body, json_str}}
+    end
+  end
+
+  def fetch_data_key(data, key) do
+    with :error <- Map.fetch(data, key) do
+      {:error, {:unexpected_body_format, data}}
+    end
+  end
```
nedimtokic: Unused functions ?
```suggestion
```

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
```diff
@@ -20,4 +20,111 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
...
+  def list_namespaces(base_namespace \\ "", acc \\ MapSet.new()) do
+    with {:ok, children} <- list_relative_namespaces(base_namespace) do
+      child_namespaces = children |> Enum.map(&(base_namespace <> &1))
+      acc = child_namespaces |> MapSet.new() |> MapSet.union(acc)
+
+      Enum.reduce_while(children, {:ok, acc}, &do_list_namespaces(base_namespace, &1, &2))
+    end
+  end
+
+  def list_relative_namespaces(base_namespace) do
```
nedimtokic: Do we need this to be public ?
```suggestion
  defp list_relative_namespaces(base_namespace) do
```

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
noaccOS: They're used by `list_relative_namespaces`, but the same pattern is also needed in other calls (eg here: https://github.com/astarte-platform/astarte/pull/1845/changes#diff-6a85501ca6d132b99d1bbe0ec7c3d20744495bc4d1aa8512720093b8107536d7R56-R58)

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
noaccOS: we don't, we can make it public later if we need it

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
nedimtokic: It confused me that the module is aliased to itself

noaccOS: yeah in a previous iterations they were used by in `OpenBao` so it was needed
still in general there is one situation where we need to call the module itself instead of just using the function name: when we want to overriede the call with mimic

PR #1863, opened 2026-03-19
chore: add import_key functionality of OpenBAO

mizzet1:
This PR aim to add the implementation of the import_key functionality of OpenBAO, as defined in the offical spec


apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
```diff
@@ -68,6 +71,163 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
...
+      {:ok, %HTTPoison.Response{status_code: 200, body: resp_body}} ->
+        with {:ok, decoded_resp_body} <- Jason.decode(resp_body),
+             {:ok, data} <- Map.fetch(decoded_resp_body, "data"),
+             {:ok, public_key} <- Map.fetch(data, "public_key") do
+          {:ok, public_key}
+        else
+          _ ->
+            Logger.error("Failed to extract public_key from wrapping_key response")
+            {:error, :wrapping_key_parse_failed}
+        end
```
noaccOS:
```suggestion
        with {:error, reason} <- parse_data_key(resp_body, "public_key") do
          Logger.error("Failed to get wrapping key: #{inspect(reason)}")
          {:error, :wrapping_key_parse_failed}
        end
```

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
```diff
@@ -68,6 +71,163 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
...
+    - `:auth_token`             - override the configured auth token
+  """
+  @spec import_key(String.t(), String.t(), term(), keyword()) ::
+          {:ok, map()}
+          | :error
+          | {:error, :wrapping_key_parse_failed}
+          | {:error, :pem_decode_failed}
+  def import_key(key_name, key_type_string, ec_key, opts \\ []) do
+    auth_token = Keyword.get(opts, :auth_token)
+    namespace = Keyword.get(opts, :namespace)
```
noaccOS: here you get the values out of opts to then recreate opts when they aren't nil

how about just

```elixir
  opts = opts |> Keyword.take([:namespace, :token])
```

?

also we usually use just `token` so let's not call the same thing with two distinct names

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
```diff
@@ -68,6 +71,163 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
...
+
+    with {:ok, wrapping_key_pem} <- get_wrapping_key(auth_token, namespace),
+         {:ok, ciphertext} <-
+           prepare_import_ciphertext(encode_ec_key_to_pkcs8(ec_key), wrapping_key_pem) do
```
noaccOS: I suggest moving this `with` block to the `open_bao` module, and having this function accept `wrapping_key_pem` and `ciphertext` as parameters, to make these functions simpler

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
```diff
@@ -68,6 +71,163 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
...
+        |> Jason.encode!()
+
+      headers = [{"Content-Type", "application/json"}]
+
+      options =
+        Keyword.reject([token: auth_token, namespace: namespace], fn {_, v} -> is_nil(v) end)
+
+      case Client.post("/transit/keys/#{key_name}/import", req_body, headers, options) do
+        {:ok, %HTTPoison.Response{status_code: 204}} ->
+          {:ok, %{}}
```
noaccOS: if you don't need to return anything don't return anything


```suggestion
          :ok
```

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
```diff
@@ -68,6 +71,163 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
...
+
+        {:ok, %HTTPoison.Response{status_code: 200, body: resp_body}} ->
+          parse_json_data(resp_body)
```
noaccOS: is this important for the user?
if not or undecided, let's just return `:ok`

apps/astarte_pairing/test/astarte_pairing/fdo/open_bao/core_test.exs
```diff
@@ -229,4 +229,109 @@ defmodule Astarte.Pairing.FDO.OpenBao.CoreTest do
...
+        {:ok, 200, [], :wk_client}
+      end)
+
+      expect(:hackney, :body, fn :wk_client, _ -> {:ok, wk_body} end)
+
+      expect(:hackney, :request, fn :post, _url, _headers, _body, _opts ->
+        {:ok, 200, [], :import_client}
+      end)
+
+      expect(:hackney, :body, fn :import_client, _ -> {:ok, body} end)
```
noaccOS: let's maybe put expectations on the `Client` module instead, no reason to go that deep

apps/astarte_pairing/test/astarte_pairing/fdo/open_bao/core_test.exs
```diff
@@ -229,4 +229,109 @@ defmodule Astarte.Pairing.FDO.OpenBao.CoreTest do
...
+    end
+
+    test "returns :error for unknown key type" do
+      assert :error = Core.key_type_to_string(:unknown)
+      assert :error = Core.key_type_to_string("ecdsa-p256")
+      assert :error = Core.key_type_to_string(nil)
+    end
+  end
+
+  describe "import_key/4" do
```
noaccOS, line 157: wait are we even testing the module or are we just using expectations on everything? I'd like at least one actual test for the feature

apps/astarte_pairing/README.md
```diff
@@ -75,3 +75,18 @@ generate-key-and-cert device-ca\
...
+iex -S mix
+
+alias Astarte.Pairing.FDO.OpenBao.Core
+alias Astarte.Pairing.FDO.OpenBao.Client
+
+ec_key = :public_key.generate_key({:namedCurve, :secp256r1})
+
+Core.import_key("my-imported-key", "ecdsa-p256", ec_key, auth_token: "astarte_token")
+
+Client.get("/transit/keys/my-imported-key", [], token: "astarte_token")
```
noaccOS: let's add the test instead of documentation

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/open_bao.ex
```diff
@@ -90,4 +90,12 @@ defmodule Astarte.Pairing.FDO.OpenBao do
   def sign(key_name, payload, alg, opts \\ []) do
     Core.sign(key_name, payload, alg, opts)
   end
+
+  @spec import_key(String.t(), Core.key_algorithm(), term(), list()) ::
+          {:ok, map()} | {:error, term()}
+  def import_key(key_name, key_type, ec_key, opts \\ []) do
```
frossq: Do we handle only import of EC keys here? We need to be able to import also RSA2048 and RSA3072 keys

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
```diff
@@ -68,6 +71,139 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
...
+  """
+  @spec import_key(String.t(), String.t(), String.t(), keyword()) :: :ok | :error
+  def import_key(key_name, key_type_string, ciphertext, opts \\ []) do
+    client_opts = Keyword.take(opts, [:namespace, :token])
+
+    req_body =
+      %{
+        type: key_type_string,
+        ciphertext: ciphertext,
+        hash_function: Keyword.get(opts, :hash_function, "SHA256"),
```
noaccOS: should this depend on the key type?

frossq: Since we hardcoded SHA256 inthe RSA-OAEP step (https://github.com/astarte-platform/astarte/pull/1863/changes#diff-6a85501ca6d132b99d1bbe0ec7c3d20744495bc4d1aa8512720093b8107536d7R101) better to avoid the possibility to declare a different hashing algo.

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/open_bao.ex
```diff
@@ -90,4 +90,16 @@ defmodule Astarte.Pairing.FDO.OpenBao do
   def sign(key_name, payload, alg, opts \\ []) do
     Core.sign(key_name, payload, alg, opts)
   end
+
+  @spec import_key(String.t(), Core.key_algorithm(), term(), list()) :: :ok | :error
+  def import_key(key_name, key_type, ec_key, opts \\ []) do
+    client_opts = Keyword.take(opts, [:namespace, :token])
```
frossq: nit: To align with the create_keypair function I would use the same opts management (Keyword.fetch! the namespace, Keyword.get additional parameters), and propagate them down to the Core module accordingly.

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
```diff
@@ -68,6 +71,139 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
...
+        exportable: Keyword.get(opts, :exportable, false),
+        allow_plaintext_backup: Keyword.get(opts, :allow_plaintext_backup, false),
+        auto_rotate_period: Keyword.get(opts, :auto_rotate_period, "0")
+      }
+      |> Jason.encode!()
+
+    headers = [{"Content-Type", "application/json"}]
+
+    case Client.post("/transit/keys/#{key_name}/import", req_body, headers, client_opts) do
+      {:ok, %HTTPoison.Response{status_code: status_code}} when status_code in [200, 204] ->
```
frossq: I think OpenBao returns only one of these two status codes for this operation... Alas, documentation does not help in understanding which one. Did you manage to get both the responses?

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
mizzet1: I think not, it's the default one and independent of key type used
[](https://openbao.org/api-docs/secret/transit/#:~:text=is%20not%20required.-,hash_function%20(string%3A%20%22SHA256%22)%20%2D%20The%20hash%20function%20used%20for%20the%20RSA%2DOAEP%20step%20of%20creating%20the%20ciphertext.%20Supported%20hash%20functions%20are%3A%20SHA1%2C%20SHA224%2C%20SHA256%2C%20SHA384%2C%20and%20SHA512.%20If%20not%20specified%2C%20the%20hash%20function%20defaults%20to%20SHA256.,-type%20(string%3A%20%3Crequired)

apps/astarte_pairing/test/astarte_pairing/fdo/open_bao/core_test.exs
```diff
@@ -229,4 +229,101 @@ defmodule Astarte.Pairing.FDO.OpenBao.CoreTest do
...
+      key_name = "imported_ec256_#{unique_id}"
+
+      {:ok, namespace} = OpenBao.create_namespace(realm_name, nil, "import_#{unique_id}")
+
+      opts = [token: token, namespace: namespace]
+
+      %{key_name: key_name, opts: opts}
+    end
+
+    test "successfully imports an EC-256 key into OpenBao", %{key_name: key_name, opts: opts} do
```
frossq: Can't all the Core.* functions inside this test be wrapped in the OpenBao.import_key() function that we expose?
Also, it would be nice to check that GETting the key from OpenBao returns the right public key corresponding with what has been imported.

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
mizzet1: I agree it can be omited since the spec mention if nothing is delcared it will be used as default

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/open_bao.ex
mizzet1: X509.PrivateKey handles both, I added some tests in core for completeness

apps/astarte_pairing/test/astarte_pairing/fdo/open_bao/core_test.exs
mizzet1: yes that's a good idea, for both things, thank you.

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
mizzet1: I confirm it  returns only 200

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/open_bao.ex
noaccOS:
```suggestion
  def import_key(key_name, key_type, key, opts \\ []) do
```

Maybe also the typespec should be more specific in what it is compared to `term`

I suggest having the api of openbao accept a `COSE.Keys.RSA` or `COSE.Keys.ECC` as those are our key representation for FDO

```elixir
@type import_key :: %COSE.Keys.RSA{} | %COSE.Keys.ECC{}
```

apps/astarte_pairing/test/astarte_pairing/fdo/open_bao/open_bao_test.exs
```diff
@@ -519,4 +478,59 @@ defmodule Astarte.Pairing.FDO.OpenBaoTest do
     OpenBao.enable_key_deletion(key_name, opts)
     OpenBao.delete_key(key_name, opts)
   end
+
+  describe "import_key/4" do
+    setup do
+      ec_key = ECC.generate(:es256)
+
+      stub(Core, :key_type_to_string, fn key_type ->
```
frossq: Isn't it just like using the real function? Can the Mimic call be avoided?

apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/open_bao.ex
mizzet1: COSE.Keys.ECC and COSE.Keys.RSA are now being used

PR #1866, opened 2026-03-23
refactor: make list keys return the list of keys

noaccOS:
previously it was returning the response body as a whole

### DIFF
--- a/apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
+++ b/apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
@@ -245,8 +245,7 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
     end
   end
 
-  @spec list_keys(String.t()) ::
-          :error | {:error, Jason.DecodeError.t()} | {:ok, any()}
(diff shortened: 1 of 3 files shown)


apps/astarte_pairing/lib/astarte_pairing/fdo/open_bao/core.ex
```diff
@@ -245,16 +245,22 @@ defmodule Astarte.Pairing.FDO.OpenBao.Core do
...
 
     case Client.list("/transit/keys", headers, options) do
       {:ok, %HTTPoison.Response{status_code: 200, body: resp_body}} ->
-        parse_json_data(resp_body)
+        case parse_data_key(resp_body, "keys") do
+          {:ok, keys} ->
+            {:ok, keys}
+
+          {:error, reason} ->
+            Logger.error("Encountered HTTP error while getting keys list: #{inspect(reason)}")
```
nedimtokic: Here we can get error while parsing list/key:
```suggestion
            Logger.error("Encountered error while getting keys list: #{inspect(reason)}")
```

nedimtokic (changes requested):

frossq (approved):

eddbbt (approved):

lusergit (approved):

PR #1875, opened 2026-03-26
feat(pairing): load voucher api

mizzet1:
This PR aims to add the logic to validate and load  the Ownership Voucher following the associated request from the client


apps/astarte_pairing/lib/astarte_pairing_web/controllers/ownership_voucher_controller.ex
```diff
@@ -55,4 +57,48 @@ defmodule Astarte.PairingWeb.OwnershipVoucherController do
...
+  given ownership voucher.
+
+  Returns `200 OK` with `{"data": {"<algorithm>": ["key_name", ...]}}` on success.
+  """
+  def compatible_keys(conn, %{"data" => data, "realm_name" => realm_name}) do
+    ownership_voucher_pem = Map.get(data, "ownership_voucher")
+
+    with {:voucher, pem} when not is_nil(pem) <- {:voucher, ownership_voucher_pem},
+         {:ok, key_algorithm} <- LoadRequest.key_algorithm_from_voucher(pem),
+         [keys_map] <- SecretsCore.get_keys_from_algorithm(realm_name, [key_algorithm]) do
```
eddbbt: get_keys_from_algorithm/2 accepts also a second argument thats not a list

mizzet1: added the additional case

apps/astarte_pairing/lib/astarte_pairing_web/controllers/ownership_voucher_controller.ex
```diff
@@ -55,4 +57,48 @@ defmodule Astarte.PairingWeb.OwnershipVoucherController do
...
+    end
+  end
+
+  @doc """
+  Returns the list of registered owner keys that are compatible with the
+  given ownership voucher.
+
+  Returns `200 OK` with `{"data": {"<algorithm>": ["key_name", ...]}}` on success.
+  """
+  def compatible_keys(conn, %{"data" => data, "realm_name" => realm_name}) do
```
noaccOS: this is a distinct feature from the load_voucher, next time let's split these changes in multiple commits to facilitate review

apps/astarte_pairing/lib/astarte_pairing_web/controllers/ownership_voucher_controller.ex
```diff
@@ -55,4 +57,48 @@ defmodule Astarte.PairingWeb.OwnershipVoucherController do
...
+    ownership_voucher_pem = Map.get(data, "ownership_voucher")
+
+    with {:voucher, pem} when not is_nil(pem) <- {:voucher, ownership_voucher_pem},
+         {:ok, key_algorithm} <- LoadRequest.key_algorithm_from_voucher(pem),
```
noaccOS: can the function be placed in `Astarte.FDO.Core.OwnershipVoucher` instead?
1. it's a method on the ownership voucher
2. load request is literally a different call

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -506,15 +510,43 @@ defmodule Astarte.Secrets.Core do
 
   def get_keys_from_algorithm(realm_name, key_algorithms) when is_list(key_algorithms) do
     Enum.map(key_algorithms, fn key_algorithm ->
-      get_keys_from_algorithm(realm_name, key_algorithm)
+      {:ok, namespace} = Secrets.create_namespace(realm_name, key_algorithm)
+
+      with {:ok, keys} <- Secrets.list_keys_names(namespace: namespace) do
+        %{key_algorithm => keys}
```
noaccOS: if we return a result, let's always return a result `{:ok, value} | {:error, reason} | :error`
in this case, let's wrap it in an `:ok` tuple. same for following functions

apps/astarte_pairing/lib/astarte_pairing_web/controllers/ownership_voucher_controller.ex
```diff
@@ -55,4 +57,48 @@ defmodule Astarte.PairingWeb.OwnershipVoucherController do
...
+         %{} = keys_map <- SecretsCore.get_keys_from_algorithm(realm_name, key_algorithm) do
```
noaccOS:
```suggestion
         {:ok, keys_map} <- SecretsCore.get_keys_from_algorithm(realm_name, key_algorithm) do

File: apps/astarte_pairing/lib/astarte_pairing_web/controllers/ownership_voucher_controller.ex [hunk 6]
```diff
@@ -55,4 +57,48 @@ defmodule Astarte.PairingWeb.OwnershipVoucherController do
...
+  @doc """
+  Returns the list of registered owner keys that are compatible with the
+  given ownership voucher.
+
+  Returns `200 OK` with `{"data": {"<algorithm>": ["key_name", ...]}}` on success.
+  """
+  def compatible_keys(conn, %{"data" => data, "realm_name" => realm_name}) do
+    ownership_voucher_pem = Map.get(data, "ownership_voucher")
+
+    with {:voucher, pem} when not is_nil(pem) <- {:voucher, ownership_voucher_pem},
```
noaccOS: wrapping values in tuple to match on the `else` of a `with` is a bad pattern because it creates complex `else` clauses, which are hard to read and follow.

in general the simplest way to refactor this is having a custom function do what you want. in this case you don't strictly need to reply to conn and return the error instead, then the response can be sent by the fallback controller

```elixir
defp ensure_ownership_voucher_parameter(data) do
  case Map.fetch(data, "ownership_voucher") do
    {:ok, voucher_pem} -> {:ok, voucher_pem}
    :error -> {:error, :invalid_parameters} # better naming could be used here
  end
end
```

apps/astarte_pairing/lib/astarte_pairing_web/router.ex
```diff
@@ -110,6 +113,7 @@ defmodule Astarte.PairingWeb.Router do
 
       post "/owner_key", OwnerKeyController, :create_or_upload_key
       get "/owner_keys", OwnerKeyController, :list_keys
+      get "/owner_key/:key_name", OwnerKeyController, :get_key
```
noaccOS: let's use `/owner_keys` everywhere (we should probably also edit the create path)

```suggestion
      get "/owner_keys/:key_name", OwnerKeyController, :get_key

File: libs/astarte_secrets/lib/astarte_secrets/core.ex [hunk 8]
```diff
@@ -506,15 +510,43 @@ defmodule Astarte.Secrets.Core do
...
   end
+
+  @supported_key_algorithms [:es256, :es384, :rs256, :rs384]
+
+  @doc """
+  Searches all algorithm namespaces for a key with the given name.
+  Returns `{:ok, key}` if found, `:not_found` otherwise.
+  """
+  def find_key(realm_name, key_name) do
+    Enum.find_value(@supported_key_algorithms, :not_found, fn algorithm ->
```
noaccOS: let's make the algorithm mandatory. if you load two keys with the same name but distinct algorithms, you want to be able to fetch both

libs/astarte_fdo/lib/ownership_voucher/load_request.ex
```diff
@@ -0,0 +1,228 @@
...
+  alias Astarte.FDO.Core.PublicKey
+  alias Astarte.FDO.OwnershipVoucher.LoadRequest
+  alias Astarte.Secrets
+  alias Astarte.Secrets.Key
+
+  require Logger
+
+  import Ecto.Changeset
+
+  typed_embedded_schema do
```
noaccOS, line 37: we're missing replacement data, they should be part of the load request

.env
```diff
@@ -22,3 +22,8 @@ RELEASE_COOKIE=astarte-docker-compose
 
 ASTARTE_OPENBAO_AUTHENTICATION_MECHANISM="token"
 ASTARTE_OPENBAO_TOKEN="astarte_token"
+
+PAIRING_ENABLE_FDO=true
+ASTARTE_BASE_URL_PORT=50000
+ASTARTE_BASE_URL_PROTOCOL="http"
+ASTARTE_BASE_URL_DOMAIN="localhost"
```
noaccOS, line 29: let's not do this yet

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -506,15 +510,43 @@ defmodule Astarte.Secrets.Core do
...
+      {:ok, namespace} = Secrets.create_namespace(realm_name, algorithm)
```
nedimtokic: What if `create_namespace` returns an error?

noaccOS (changes requested):
first commit review only

libs/astarte_secrets/lib/astarte_secrets/core.ex
mizzet1: added a clause to handle the error case from create_namespace instead of crashing with a MatchError

apps/astarte_pairing/lib/astarte_pairing_web/controllers/ownership_voucher_controller.ex
mizzet1: thank you will keep this in mind

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fallback_controller.ex
```diff
@@ -94,6 +94,22 @@ defmodule Astarte.PairingWeb.FallbackController do
...
 
+  # The ownership_voucher parameter was missing from the request
+  def call(conn, {:error, :missing_ownership_voucher}) do
+    conn
+    |> put_status(:unprocessable_entity)
+    |> put_view(Astarte.PairingWeb.ErrorView)
+    |> render(:missing_ownership_voucher)
+  end
+
+  # Generic error fallback
```
noaccOS: let's not have generic error fallbacks, and especially not with 422

apps/astarte_pairing/lib/astarte_pairing_web/router.ex
```diff
@@ -82,7 +82,10 @@ defmodule Astarte.PairingWeb.Router do
     scope "/ownership" do
```
frossq: Scope /fdo/ownership_voucher? As per RFC document

apps/astarte_pairing/lib/astarte_pairing_web/controllers/ownership_voucher_controller.ex
```diff
@@ -20,8 +20,11 @@ defmodule Astarte.PairingWeb.OwnershipVoucherController do
   use Astarte.PairingWeb, :controller
 
   alias Astarte.DataAccess.FDO.OwnershipVoucher.CreateRequest
+  alias Astarte.FDO.Core.OwnershipVoucher, as: CoreOwnershipVoucher
   alias Astarte.FDO.OwnershipVoucher
+  alias Astarte.FDO.OwnershipVoucher.LoadRequest
   alias Astarte.FDO.TO0
+  alias Astarte.Secrets.Core, as: SecretsCore
```
noaccOS, line 26: if we need a core function somewhere let's re-export it in the base module

apps/astarte_pairing/lib/astarte_pairing_web/controllers/ownership_voucher_controller.ex
```diff
@@ -55,4 +58,43 @@ defmodule Astarte.PairingWeb.OwnershipVoucherController do
...
+
+  @doc """
+  Returns the list of registered owner keys that are compatible with the
+  given ownership voucher.
+
+  Returns `200 OK` with `{"data": {"<algorithm>": ["key_name", ...]}}` on success.
+  """
+  def compatible_keys(conn, %{"data" => data, "realm_name" => realm_name}) do
+    with {:ok, pem} <- ensure_ownership_voucher_parameter(data),
+         {:ok, key_algorithm} <- CoreOwnershipVoucher.key_algorithm_from_voucher(pem),
```
noaccOS: here I expect something like

```elixir
with {:ok, pem} <- ensure_ownership_voucher_parameter(data),
     {:ok, voucher} <- OwnershipVoucher.decode_binary_voucher(pem),
     key_algorithm = voucher |> OwnershipVoucher.key_algorithm(),
     {:ok, keys_map} <- SecretsCore.get_keys_from_algorithm(realm_name, key_algorithm)
```

apps/astarte_pairing/lib/astarte_pairing_web/router.ex
```diff
@@ -82,7 +82,10 @@ defmodule Astarte.PairingWeb.Router do
...
       pipe_through :fdo_feature_gate
       pipe_through :agent_api
+
       post "/", OwnershipVoucherController, :create
+      post "/load", OwnershipVoucherController, :load
```
noaccOS: let's use /fdo/ownership_vouchers for loading vouchers

apps/astarte_pairing/lib/astarte_pairing_web/router.ex
```diff
@@ -82,7 +82,10 @@ defmodule Astarte.PairingWeb.Router do
...
+      post "/compatible_keys", OwnershipVoucherController, :compatible_keys
```
noaccOS: maybe something like a get on /fdo/owner_keys_for_voucher?

noaccOS (changes requested):

noaccOS (approved):

PR #1897, opened 2026-04-02
fix(pairing): re add key algorithm type logic

mizzet1:


libs/astarte_fdo/lib/ownership_voucher/ownership_voucher.ex
```diff
@@ -91,4 +91,16 @@ defmodule Astarte.FDO.OwnershipVoucher do
...
+  @doc """
+  Returns the list of key algorithm atoms compatible with the given ownership voucher.
+  Returns an empty list if the key type is unsupported.
+  """
+  @spec key_algorithm(OwnershipVoucher.t()) :: [atom()]
+  def key_algorithm(voucher) do
+    case OwnershipVoucher.key_algorithm_from_type(voucher.header.public_key.type) do
+      {:ok, algorithms} -> algorithms
+      {:error, _} -> []
+    end
```
noaccOS: we can be assertive!


```suggestion
   # we can be pretty sure the public key in the header is a valid type if we parsed it correctly :3
   {:ok, algorithms} = OwnershipVoucher.key_algorithm_from_type(voucher.header.public_key.type)
   algorithms
```

libs/astarte_fdo/lib/ownership_voucher/load_request.ex
```diff
@@ -127,19 +129,39 @@ defmodule Astarte.FDO.OwnershipVoucher.LoadRequest do
...
+
+  defp fetch_key_for_algorithm(realm_name, key_name, key_algorithm) do
     with {:ok, namespace} <- Secrets.create_namespace(realm_name, key_algorithm),
          {:ok, key} <- Secrets.get_key(key_name, namespace: namespace) do
-      put_change(changeset, :extracted_owner_key, key)
+      {:ok, key}
     else
-      _err ->
-        add_error(changeset, :key_name, "does not exist in secrets store")
+      _ -> nil
```
noaccOS: if we return results, let's always return `{:ok, value} | :error | {:error, reason}` (only one of the last two!)

libs/astarte_fdo/lib/ownership_voucher/load_request.ex
```diff
@@ -109,12 +109,14 @@ defmodule Astarte.FDO.OwnershipVoucher.LoadRequest do
     with {:ok, binary_voucher} <- OwnershipVoucher.binary_voucher(ownership_voucher_pem),
          {:ok, decoded_voucher, _rest} <- CBOR.decode(binary_voucher),
          {:ok, voucher_struct} <- OwnershipVoucher.decode(decoded_voucher),
-         {:ok, owner_public_key} <-
-           OVCore.entry_public_key(List.last(voucher_struct.entries)) do
+         {:ok, %PublicKey{type: pubkey_type} = owner_public_key} <-
+           OVCore.entry_public_key(List.last(voucher_struct.entries)),
+         {:ok, key_algorithm} <- OwnershipVoucher.key_algorithm_from_type(pubkey_type) do
```
noaccOS: we don't need that!

libs/astarte_fdo/lib/ownership_voucher/load_request.ex
```diff
@@ -109,12 +109,14 @@ defmodule Astarte.FDO.OwnershipVoucher.LoadRequest do
...
       changeset
       |> put_change(:cbor_ownership_voucher, binary_voucher)
       |> put_change(:decoded_ownership_voucher, decoded_voucher)
       |> put_change(:voucher_struct, voucher_struct)
+      |> put_change(:owner_key_algorithm, key_algorithm)
```
noaccOS: let's delete the field from the struct

noaccOS (changes requested):

noaccOS (approved):

PR #1904, opened 2026-04-03
fix(pairing): fix ownership_voucher_controller errors

mizzet1:


apps/astarte_pairing/test/astarte_pairing_web/controllers/ownership_voucher_controller_test.exs
```diff
@@ -149,6 +151,8 @@ defmodule Astarte.PairingWeb.Controllers.OwnershipVoucherControllerTest do
   defp register_setup(context) do
     %{auth_conn: conn, realm_name: realm_name} = context
     register_path = ownership_voucher_path(conn, :register, realm_name)
+    stub(OwnershipVoucher, :save_voucher, fn _, _ -> :ok end)
+    stub(TO0, :claim_ownership_voucher, fn _, _, _ -> :ok end)
```
noaccOS: why do we need to stub? if these tests worked previously, let's make them work frfr

apps/astarte_pairing/test/astarte_pairing_web/controllers/ownership_voucher_controller_test.exs
```diff
@@ -171,7 +175,7 @@ defmodule Astarte.PairingWeb.Controllers.OwnershipVoucherControllerTest do
         |> post(path, %{data: %{"ownership_voucher" => sample_voucher()}})
         |> json_response(200)
 
-      assert %{"es256" => [@sample_key_name, "another_key"]} = get_in(body, ["data"])
+      assert [%{"es256" => [@sample_key_name, "another_key"]}] = get_in(body, ["data"])
```
noaccOS: why is this returning a list now? I think a map makes more sense

libs/astarte_fdo/lib/ownership_voucher/load_request.ex
```diff
@@ -101,6 +102,49 @@ defmodule Astarte.FDO.OwnershipVoucher.LoadRequest do
...
+  defp decode_replacement_guid(changeset) do
+    case fetch_change(changeset, :replacement_guid) do
+      {:ok, b64} ->
+        {:ok, bin} = Base.decode64(b64)
+        put_change(changeset, :replacement_guid, bin)
+
+      :error ->
+        changeset
+    end
+  end
```
noaccOS, line 125: this is assertive on user input: let's not do that


```suggestion
  defp decode_replacement_guid(changeset) do
    case fetch_change(changeset, :replacement_guid) do
      {:ok, b64} ->
        case Base.decode64(b64) do
          {:ok, bin} -> put_change(changeset, :replacement_guid, bin)
          :error -> put_error(changeset, :replacement_guid, ...)
        end

      :error ->
        changeset
    end
  end
```

libs/astarte_fdo/lib/ownership_voucher/load_request.ex
```diff
@@ -101,6 +102,49 @@ defmodule Astarte.FDO.OwnershipVoucher.LoadRequest do
...
+
+  defp decode_replacement_rendezvous_info(changeset) do
+    case fetch_change(changeset, :replacement_rendezvous_info) do
+      {:ok, b64} ->
+        {:ok, cbor_bin} = Base.decode64(b64)
+        {:ok, rv_info} = RendezvousInfo.decode_cbor(cbor_bin)
+        put_change(changeset, :decoded_replacement_rendezvous_info, rv_info)
```
noaccOS: same as above

libs/astarte_fdo/lib/ownership_voucher/load_request.ex
```diff
@@ -101,6 +102,49 @@ defmodule Astarte.FDO.OwnershipVoucher.LoadRequest do
...
+
+      :error ->
+        changeset
+    end
+  end
+
+  defp decode_replacement_public_key(changeset) do
+    case fetch_change(changeset, :replacement_public_key) do
+      {:ok, pem_string} ->
+        {:ok, public_key} = public_key_from_pem(pem_string)
```
noaccOS: same as above

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -532,22 +532,32 @@ defmodule Astarte.Secrets.Core do
...
+    {:ok, keys_list}
+  end
+
+  defp list_keys_for_algorithm(realm_name, key_algorithm) do
+    keys =
+      with {:ok, namespace} <- Secrets.create_namespace(realm_name, key_algorithm),
+           {:ok, keys} <- Secrets.list_keys_names(namespace: namespace) do
+        keys
+      else
+        _ -> []
```
noaccOS: this hides internal errors and returns "yes I promise you didn't have any keys", and eg says "you used the wrong key for the voucher" when really there was an error with openbao
it's not great ux, we should return the error to the user

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -532,22 +532,32 @@ defmodule Astarte.Secrets.Core do
...
+      end
 
-    {:ok, keys_map}
+    %{key_algorithm => keys}
```
noaccOS: (suggestion for an ideal world)

here if you 

```suggestion
    {key_algorithm, keys}
```

then in `get_keys_from_algorithm/2` you can

```elixir
keys_list = Map.new(key_algorithms, &list_keys_for_algorithm(realm_name, &1))
```

as this is not an ideal world, here we want to use a reduce_while so that we stop if an error happens at any time

```elixir
  def get_keys_from_algorithm(realm_name, key_algorithms) do
    key_algorithms
    |> Enum.reduce_while({:ok, %{}}, fn algorithm, {:ok, acc} ->
      case list_keys_for_algorithm(realm_name, algorithm) do
        {:ok, keys} -> {:cont, {:ok, Map.put(acc, algorithm, keys)}}
        error -> {:halt, error}
      end
    end)
  end
```

noaccOS (changes requested):

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -532,22 +532,31 @@ defmodule Astarte.Secrets.Core do
...
+  def get_keys_from_algorithm(realm_name, key_algorithm) when is_binary(key_algorithm) do
+    with {:ok, atom_alg} <- string_to_key_type(key_algorithm),
+         {:ok, namespace} <- Secrets.create_namespace(realm_name, atom_alg),
+         {:ok, keys} <- Secrets.list_keys_names(namespace: namespace) do
+      {:ok, %{key_algorithm => keys}}
+    else
+      :error -> :error
+      _ -> :error
+    end
+  end
```
noaccOS: let's not duplicate our functions, let's reuse the existing `get_keys_from_algorithm` (which should maybe be called get_keys_**for**_algorithm or even just get_keys)

libs/astarte_fdo/lib/ownership_voucher/load_request.ex
noaccOS: these are all still valid

noaccOS (changes requested):

libs/astarte_secrets/lib/astarte_secrets/core.ex
noaccOS: Sorry for the confusion: I meant _users_ should all use the same function and it shouldn't be any single function responsibility to accept multiple arguments and perform the cast itself. if callers need to cast, they can do so explicitly **before** calling this function

noaccOS (changes requested):

libs/astarte_secrets/lib/astarte_secrets/core.ex
mizzet1: get_keys_for_algorithm now accepts only a list of atoms and if callers need to cast, like in owner_key_controller, they do it before calling the function via string_to_key_type

noaccOS (approved):
👑

PR #1914, opened 2026-04-07
test(appengine): Fix groups and interfaces tests

nedimtokic:
Fix tests to not generate duplicated groups in property tests.
Fix interfaces duplicate cases.

### DIFF
--- a/apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/groups_controller_test.exs
+++ b/apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/groups_controller_test.exs
@@ -192,6 +192,10 @@ defmodule Astarte.AppEngine.APIWeb.GroupsControllerTest do
 
         assert json_response(show_conn, 200)["data"]["group_name"] == group_name,
                "Failed post/get same group_name #{group_name}"
+
+        Enum.each(@group_devices, fn device_id ->
+          Groups.remove_device(@realm, group_name, device_id)
+        end)
       end
     end
   end
--- a/apps/astarte_appengine_api/test/support/cases/device.ex
+++ b/apps/astarte_appengine_api/test/support/cases/device.ex
@@ -157,123 +157,171 @@ defmodule Astarte.Cases.Device do
   end
 
   defp interfaces_for_update do
+    interface_specs = [
+      {:individual_datastream_device,
+       fn acc -> new_interfaces(individual_datastream_device(), acc) end},
+      {:individual_datastream_server,
+       fn acc -> new_interfaces(individual_datastream_server(), acc) end},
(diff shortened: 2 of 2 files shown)


frossq (approved):

eddbbt (approved):

apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/groups_controller_test.exs
```diff
@@ -177,21 +177,24 @@ defmodule Astarte.AppEngine.APIWeb.GroupsControllerTest do
 
     @tag issue: 904
     property "creates the group with / in group name", %{conn: conn} do
-      check all group_name <- GroupGenerator.name() do
-        params = %{
-          "group_name" => group_name,
-          "devices" => @group_devices
-        }
+      check all group_names <- uniq_list_of(GroupGenerator.name(), min_length: 1, max_length: 10),
```
noaccOS: this looks a bit weird, and it doesn't solve the issue that it may generate the same group name in one of the 10 runs

instead, we should restore the previous logic but add a delete group after the last `assser`

noaccOS (changes requested):

noaccOS:
This should also be rebased on release-1.3

apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/groups_controller_test.exs
nedimtokic: That was intention, but I didn't find Query function for group deletion. Should I add it ?

noaccOS: You can also run a dirty `Repo.delete` and call it a day

apps/astarte_appengine_api/test/astarte_appengine_api_web/controllers/groups_controller_test.exs
```diff
@@ -192,6 +192,10 @@ defmodule Astarte.AppEngine.APIWeb.GroupsControllerTest do
 
         assert json_response(show_conn, 200)["data"]["group_name"] == group_name,
                "Failed post/get same group_name #{group_name}"
+
+        Enum.each(@group_devices, fn device_id ->
+          Groups.remove_device(@realm, group_name, device_id)
+        end)
```
noaccOS, line 198: we also need to delete the group itself afterwards otherwise the `:create` may fail

nedimtokic: We have `GroupedDevice` table where devices and groups are stored together. So deleting all devices from  `GroupedDevice` table, group will be deleted also.

noaccOS (approved):

PR #1915, opened 2026-04-07
feat(hk): Support for retrieving default Realm replication settings via API

eddbbt:


frossq (approved):

frossq:
Tests? :point_right: :point_left:

AliouneGaye21 (approved):

apps/astarte_housekeeping/lib/astarte_housekeeping_web/controllers/realm_controller.ex
```diff
@@ -77,6 +77,27 @@ defmodule Astarte.HousekeepingWeb.RealmController do
...
+            replication_class: :simple_strategy,
+            replication_factor: Config.astarte_keyspace_replication_factor!()
+          }
+
+        :network_topology_strategy ->
+          %{
+            replication_class: :network_topology_strategy,
+            datacenter_replication_factor: Config.astarte_keyspace_network_replication_map!()
+          }
+      end
```
noaccOS, line 257: this should really be the new default replication, and should be read from the database
of course it does not exist yet so I'm not sure this was the best task to start with

noaccOS (changes requested):

apps/astarte_housekeeping/lib/astarte_housekeeping_web/controllers/realm_controller.ex
eddbbt: i'll leave a TODO and change this in the next PRs

apps/astarte_housekeeping/lib/astarte_housekeeping_web/controllers/realm_controller.ex
```diff
@@ -77,6 +77,27 @@ defmodule Astarte.HousekeepingWeb.RealmController do
...
+
+    conn
+    |> put_resp_content_type("application/json")
+    |> send_resp(:ok, Jason.encode!(%{data: replication}))
```
noaccOS: how about using a view instead? look how eg `update/2` works

noaccOS (changes requested):
let's add a simple test (eg that a get on the path returns a replication class in "simple_strategy" or "network_topology_strategy" or whatever) and it's good to go

noaccOS (approved):

PR #1944, opened 2026-04-23
feat(data_access): add core to changeset - interface

shinnokdisengir:
## Description
Introduces the `Astarte.DataAccess.Adapters.Core.InterfaceMapping` module. This module defines the transformation pipelines required to convert core domain structures into database-ready Ecto changesets.

## Technical Changes
* **Interface Transformation (`from_core/1`)**: Maps `Astarte.Core.Interface` fields to `Astarte.DataAccess.Realms.Interface`.
* **Dynamic Storage Routing**: Computes `storage` and `storage_type` fields dynamically using pattern matching on the interface type and aggregation strategy.
* **Mapping Transformation (`from_core_mapping/1`)**: Transforms nested `Astarte.Core.Mapping` lists into `Astarte.DataAccess.Realms.Endpoint` changesets, injecting the parent `interface_id`.
* **Post-processing Hooks**: Integrates Ecto changeset validation structures via `Interface.changeset/2` and `Endpoint.changeset/2`.

## Testing
* Added comprehensive test suites validating structural parity, strict null-value invariants, and correct Ecto validation state.


libs/astarte_data_access/lib/astarte_data_access/helpers/ecto.ex
```diff
@@ -0,0 +1,63 @@
...
+      end
+  """
+  @spec cast_normalized(
+          Ecto.Schema.t() | Ecto.Changeset.t() | {map(), map()},
+          any(),
+          [atom()],
+          (any() -> map()),
+          Keyword.t()
+        ) :: Ecto.Changeset.t()
+  def cast_normalized(data, params, permitted, normalize_fun, opts \\ [])
```
eddbbt: what happens when `normalize_fun` is not a function? consider to handle the case

shinnokdisengir: there is `when is_function(normalize_fun, 1)` as _guard_

libs/astarte_adapters/lib/astarte/adapters/engine.ex
```diff
@@ -0,0 +1,165 @@
...
+         _source_path,
+         _dest_field_name,
+         _req,
+         nil
+       ) do
+    deep_put(acc, dest_path, val)
+  end
+
+  def apply_pipe(data, nil), do: data
+  def apply_pipe(data, fun) when is_function(fun, 1), do: fun.(data)
```
Annopaolo: This function seems to be called only in this module's tests. Do I miss something? If not, then it could be removed

libs/astarte_adapters/lib/astarte/adapters/engine.ex
```diff
@@ -0,0 +1,165 @@
...
+
+  def deep_put(map, path, value) when is_atom(path) or is_binary(path),
+    do: deep_put(map, [path], value)
+
+  def deep_put(map, [key], value), do: Map.put(map, key, value)
+
+  def deep_put(map, [key | rest], value) do
+    sub_map = ensure_map(Map.get(map, key))
+    Map.put(map, key, deep_put(sub_map, rest, value))
+  end
```
Annopaolo, line 43: These could probably be replaced by `get_in/2` and `put_in/2` from [Access](https://hexdocs.pm/elixir/Access.html#module-maps-and-structs)

libs/astarte_adapters/coveralls.json
Annopaolo, line 1: Why is this file skipped?

libs/astarte_adapters/lib/astarte/adapters.ex
```diff
@@ -0,0 +1,234 @@
...
+      import Astarte.Adapters
+      import Astarte.Adapters.Engine
+    end
+  end
+
+  @doc """
+  Defines a mapping rule. See the module documentation for `Astarte.Adapters`
+  for detailed usage and examples.
+  """
+  defmacro defmapping(name, opts \\ [], do: block) do
```
Annopaolo: The word `mapping` might contrast with some Astarte concepts. What about a slightly more general one?

```suggestion
  defmacro transform(name, opts \\ [], do: block) do
```

libs/astarte_data_access/lib/astarte_data_access/adapters/core/interface_mapping.ex
```diff
@@ -0,0 +1,108 @@
...
+    do: :multi_interface_individual_datastream_dbtable
+
+  defp storage_type(%InterfaceCore{type: :datastream, aggregation: :object}),
+    do: :one_object_datastream_dbtable
+
+  defp from_core_mappings(mappings, %InterfaceCore{interface_id: interface_id}),
+    do: Enum.map(mappings, &from_core_mapping(%{interface_id: interface_id, mapping: &1}))
+
+  defp from_core_post_process(%{interface: interface} = source),
+    do: %{source | interface: Interface.changeset(%Interface{}, interface)}
```
Annopaolo, line 81: The return type of this (and of the `:from_core` transform) is a map, while I would expect a changeset like `from_core_mapping_post_process` does. Is this choice due to the need to keep the `source` when `transform`-ing an Interface?

libs/astarte_data_access/test/adapters/core/interface_mapping_test.exs
```diff
@@ -0,0 +1,161 @@
...
+       do: CQLUtils.interface_name_to_table_name(name, major_version)
+
+  defp storage_type(%InterfaceCore{type: :properties, aggregation: :individual}),
+    do: :multi_interface_individual_properties_dbtable
+
+  defp storage_type(%InterfaceCore{type: :datastream, aggregation: :individual}),
+    do: :multi_interface_individual_datastream_dbtable
+
+  defp storage_type(%InterfaceCore{type: :datastream, aggregation: :object}),
+    do: :one_object_datastream_dbtable
```
Annopaolo: These could be public functions in the `Astarte.DataAccess.Adapters.Core.InterfaceMapping` module

libs/astarte_data_access/lib/astarte_data_access/adapters/core/interface_mapping.ex
shinnokdisengir: The choice depends on _what_ you want to achieve when designing the transformation, given that, in theory, you could consider 
- one-to-one
- one-to-many
- many-to-one
- many-to-many

A transformation into "edible" objects from `Ecto` in `Astarte.Core.Interface` consists of a list of `Changeset`s for `Endpoint`s plus a `Changeset` for the `Interface` itself; therefore, when designing the function, I chose %{interface: Changeset, endpoints: list(Changeset)`.

`from_core_mapping_post_process` is an internal function, and its purpose is precisely to create `Ecto.Changeset` objects that will then be included in the `Map` list.

Unfortunately, as the language is not strongly typed this does not help, and the only way to specify which types of _entities_ are taken as input and which as output can only be managed via `@spec` and documentation.

libs/astarte_data_access/test/adapters/core/interface_mapping_test.exs
shinnokdisengir: In theory, the tester shouldn’t assume that the code they’re about to test is correct, but should generate their own code even in cases where there is significant redundancy.
This is an extreme case, given that it is entirely _functional_ and simple, so it goes without saying that identical code will be generated.

I don’t have a strong opinion on the matter; I’ll go with whatever you prefer.

libs/astarte_data_access/lib/astarte_data_access/adapters/core/interface_mapping.ex
Annopaolo: Since this module is meant to transform an Interface (as defined in `astarte_core`) into the structures that will be inserted into the database, it makes sense to return all changeset structs that will be used by Ecto. Thank you for clarifying!

libs/astarte_data_access/test/adapters/core/interface_mapping_test.exs
Annopaolo: I'd say to make them public as it might happen to use these functions even outside of the `InterfaceMapping` module/tests

nedimtokic (approved):

eddbbt (approved):

Annopaolo (approved):

noaccOS (approved):
:sushi:

PR #1934, opened 2026-04-16
feat(housekeeping): persist keyspace replication settings to KvStore during migration

eddbbt:

PR #1947, opened 2026-04-27
feat(housekeeping): default astarte keyspace replication to NetworkTopologyStrategy

mizzet1:
This PR introduces the following change: when HOUSEKEEPING_ASTARTE_KEYSPACE_REPLICATION_STRATEGY is not set, derive the replication map from the live ScyllaDB network topology (one replica per node in each datacenter) instead of falling back to SimpleStrategy with replication_factor=1)


eddbbt:
check https://github.com/astarte-platform/astarte/pull/1934

eddbbt (approved):

PR #1953, opened 2026-04-29
feat: introduce Astarte.Adapters DSL for declarative data mapping

shinnokdisengir:
## Description
This PR introduces the `Astarte.Adapters` module, a declarative DSL designed to handle complex mappings between nested data structures (e.g., adapting external API payloads into internal domain structs).

## Key Features
* **Declarative DSL:** Use `defmapping` to map fields from deep source paths to destination paths.
* **Direct Function Generation:** The macro compiles directly into standard, highly performant public functions (arity 1). No routing overhead.
* **Static Typing Support:** `defmapping` accepts `source:` and `returns:` options to dynamically generate `@spec` definitions, providing full support for Dialyzer and editor autocompletion.
* **Strict Options Parser:** `field` definitions enforce explicit keyword lists for options:
  * `:required` (boolean): Strict checks on missing or `nil` values.
  * `:custom`: Strictly requires an arity 2 function (`fn extracted_val, full_source -> ... end`) for advanced transformations and context awareness.
* **Error Handling:** Gracefully handles missing keys, `nil` values, and path mismatches. Introduces a custom `Astarte.Adapters.MissingFieldError` exception for missing required fields.
* **100% Test Coverage:** The runtime engine is fully covered by unit tests and `StreamData` property tests, verifying N-depth flat-to-nested and nested-to-flat transformations. Macros are explicitly excluded from runtime coverage metrics.


nedimtokic (approved):

libs/astarte_adapters/lib/astarte/adapters.ex
```diff
@@ -0,0 +1,213 @@
...
+        end
+      end
+
+      source = %{
+        data: %{
+          id: "123",
+          attributes: %{name: "example"}
+        },
+        type: "user"
+      }
```
nedimtokic: What happens if source map keys are strings ?

libs/astarte_adapters/lib/astarte/adapters/engine.ex
```diff
@@ -0,0 +1,147 @@
...
+  defp ensure_map(_), do: %{}
+
+  def process_field(
+        acc,
+        source_map,
+        dest_path,
+        source_path,
+        dest_field_name,
+        required,
+        custom_fun
```
nedimtokic, line 55: What do you think of raising error if custom function arity is not 2 ?

libs/astarte_adapters/lib/astarte/adapters/engine.ex
```diff
@@ -0,0 +1,147 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Astarte.Adapters.Engine do
+  @moduledoc false
+
+  def deep_get(data, path) when is_atom(path), do: deep_get(data, [path])
+  def deep_get(data, []), do: {:ok, data}
```
eddbbt, line 22: put guarded functions first, fallback last, the other in the middle, i'll move this one before `  def deep_get(_data, _path), do: :error`

eddbbt (approved):

libs/astarte_adapters/lib/astarte/adapters.ex
shinnokdisengir: fixed, there wasn't correct guard in `defp normalize_path(path) when is_atom(path) or is_binary(path), do: [path]`

libs/astarte_adapters/lib/astarte/adapters/engine.ex
shinnokdisengir: @eddbbt asked me the same thing, but there’s no need: if the arity isn’t correct, it won’t compile.

I'm adding a sample test

shinnokdisengir: OK, I’ve changed my mind; I’d better do as you say: I’d forgotten that, when it comes to checking _arity_ at macro level, there would be the problem of a function assigned to a **variable**, which would invalidate the whole thing. 🙏

frossq (approved):

nedimtokic (approved):

PR #1954, opened 2026-04-29
feat(housekeeping): default realm keyspace replication uses data from db

mizzet1:
This PR introduces the following change: when a realm is created without specyfing replication strategy and factor, the default strategy stored in kv store is used

### DIFF
--- a/apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
+++ b/apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
@@ -59,6 +59,35 @@ defmodule Astarte.Housekeeping.Realms.Queries do
     end
   end
 
+  def fetch_keyspace_replication do
+    astarte_keyspace = Realm.astarte_keyspace_name()
+    consistency = Consistency.domain_model(:read)
+
+    result =
+      KvStore.fetch_value("astarte", "db_default_replication", :binary,
+        error: :replication_not_found,
+        consistency: consistency,
+        prefix: astarte_keyspace
+      )
+
+    case result do
+      {:ok, binary_replication} ->
+        try do
+          {:ok, :erlang.binary_to_term(binary_replication)}
+        rescue
+          _ ->
+            Logger.error("Failed to deserialize replication data from KvStore",
+              tag: "corrupted_replication_data"
+            )
+
+            {:error, :corrupted_replication_data}
+        end
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
   def get_realm(realm_name) do
     keyspace_name = Realm.keyspace_name(realm_name)
     do_get_realm(realm_name, keyspace_name)
@@ -1203,8 +1232,12 @@ defmodule Astarte.Housekeeping.Realms.Queries do
       value_type: :binary
     }
 
-    with {:error, xandra_error} <- KvStore.insert(kv_store, opts) do
-      raise xandra_error
+    case KvStore.insert(kv_store, opts) do
+      :ok ->
+        :ok
+
+      {:error, xandra_error} ->
+        {:error, xandra_error}
     end
   end
 
--- a/apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realm.ex
+++ b/apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realm.ex
@@ -27,9 +27,6 @@ defmodule Astarte.Housekeeping.Realms.Realm do
 
   require Logger
 
-  @default_replication_factor 1
-  @default_replication_class "SimpleStrategy"
-
   @required_create_fields [:realm_name, :jwt_public_key_pem]
   @allowed_create_fields [
     :replication_factor,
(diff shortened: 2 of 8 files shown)


eddbbt (approved):
review'd irl

nedimtokic (approved):

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
```diff
@@ -1203,8 +1232,12 @@ defmodule Astarte.Housekeeping.Realms.Queries do
       value_type: :binary
     }
 
-    with {:error, xandra_error} <- KvStore.insert(kv_store, opts) do
-      raise xandra_error
+    case KvStore.insert(kv_store, opts) do
```
frossq, line 1235: I'd stay with the raise condition if there aren't particular needs to move away from it. Apparently we want to treat a KVStore.insert error as an exception

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realm.ex
```diff
@@ -66,7 +66,7 @@ defmodule Astarte.Housekeeping.Realms.Realm do
     |> validate_change(:realm_name, &validate_realm_name/2)
     |> validate_number(:replication_factor, greater_than: 0)
```
frossq, line 64: At this point better to refine the validation of the replication parameters, like validate that 
-`replication_factor is in [simple, network_topology]`
-`if simple -> validate replication factor`
-`if network -> validate dc map`

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realm.ex
```diff
@@ -125,6 +125,17 @@ defmodule Astarte.Housekeeping.Realms.Realm do
     end
   end
 
+  defp maybe_put_default_replication_class(changeset) do
+    case {get_field(changeset, :replication_class), get_field(changeset, :replication_factor)} do
+      # Neither provided: leave nil so core.ex auto-inherits from astarte keyspace
+      {nil, nil} -> changeset
+      # Factor provided but no class: default to SimpleStrategy
+      {nil, _factor} -> put_change(changeset, :replication_class, @default_replication_class)
```
frossq: Do we have a use case for this clause?

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realm.ex
frossq: Not sure we want to keep these defaults

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realm.ex
frossq: (also accepting nil/nil if we want to leave the settings as housekeeping defaults)

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realm.ex
```diff
@@ -125,6 +125,17 @@ defmodule Astarte.Housekeeping.Realms.Realm do
     end
   end
 
+  defp maybe_put_default_replication_class(changeset) do
+    case {get_field(changeset, :replication_class), get_field(changeset, :replication_factor)} do
+      # Neither provided: leave nil so core.ex auto-inherits from astarte keyspace
+      {nil, nil} -> changeset
```
frossq: Idea for alternative implementation: if  == {nil, nil} derive the replication class and factor from KVStore and update directly here the changeset, propagating the full map of realm parameters. In this way we can fully reuse the rest of the code without adding further logic and code paths, harder to track. Let's discuss!

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realm.ex
mizzet1: removed the second clause, better to return an erorr

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
mizzet1: this change was done so the error can propagate through the do_ensure_migrated function, which raise in case of error

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realm.ex
```diff
@@ -66,9 +63,8 @@ defmodule Astarte.Housekeeping.Realms.Realm do
     |> validate_change(:realm_name, &validate_realm_name/2)
     |> validate_number(:replication_factor, greater_than: 0)
     |> validate_change(:jwt_public_key_pem, &validate_pem_public_key/2)
-    |> put_default_if_missing(:replication_class, @default_replication_class)
     |> validate_inclusion(:replication_class, [
```
frossq, line 66: Also nil is a valid replication_class now (?)

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realms.ex
```diff
@@ -79,6 +80,14 @@ defmodule Astarte.Housekeeping.Realms do
     end
   end
 
+  defp resolve_replication(%Realm{replication_class: nil} = realm) do
```
frossq, line 83: Nit: this resolve_replication check could be contained within the Realm module (within changeset operations), before passing the Realm struct to Realms module.

frossq (approved):

PR #1969, opened 2026-05-19
ci: Fix versions dropdown on Astarte github pages

nedimtokic:
Generate `docs_config.js` required by `ex_doc` to autogenerate app version dropdown on page.


noaccOS (approved):
🚀

PR #1971, opened 2026-05-20
feat(hk): Add configurable default datastream maximum storage retention for new realms

eddbbt:


Dockerfile
```diff
@@ -26,7 +26,6 @@ COPY libs/astarte_data_access/mix.exs libraries/astarte_data_access/mix.exs
 COPY libs/astarte_data_access/mix.lock libraries/astarte_data_access/mix.lock
 COPY libs/astarte_events/mix.exs libraries/astarte_events/mix.exs
 COPY libs/astarte_events/mix.lock libraries/astarte_events/mix.lock
-RUN mix do deps.get, deps.compile
```
noaccOS: mhhh this should technically be 

```dockerfile
RUN mix do deps.get, deps.compile --skip-local-deps
```

instead as it is in all other services, but: was this change made because it was creating problems as it was?

apps/astarte_housekeeping/lib/astarte_housekeeping/realms/realms.ex
```diff
@@ -80,6 +82,16 @@ defmodule Astarte.Housekeeping.Realms do
...
   end
 
+  defp maybe_inject_default_retention(attrs) do
+    case Config.default_datastream_maximum_storage_retention() do
+      {:ok, nil} ->
+        attrs
+
+      {:ok, default_retention} ->
+        attrs |> Map.put_new("datastream_maximum_storage_retention", default_retention)
+    end
```
noaccOS, line 92: nit: this can technically be simplified as putting nil is the same as not putting anything

```elixir
default_retention = Config.default_datastream_maximum_storage_retention!()
attrs |> Map.put_new("datastream_maximum_storage_retention", default_retention)
```

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -50,6 +50,16 @@ defmodule Astarte.Housekeeping.Config do
     type: :boolean,
     default: false
 
+  @envdoc """
+  By default realms do not have a default retention, meaning that data is stored indefinitely. Set this variable to set default values for new realms.
```
frossq: Unit of measurement?

frossq:
Update changelog :pray:

frossq (approved):

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
eddbbt: bananas, usually /s

shall be seconds

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -50,6 +50,16 @@ defmodule Astarte.Housekeeping.Config do
...
+  It does not have constraint like retention in interfaces and can be overridden at any time
```
noaccOS: this is documenting the feature of datastream maximum storage retention not describing what the variable does

```
Default maximum storage retention for new realms
```

would be enough, with maybe a link to the documentation for maximum storage retention (which I'm not sure we have as of now)

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## unreleased
+
+### added
```
noaccOS: let's keep capitalization as per https://keepachangelog.com/en/1.1.0/

noaccOS (changes requested):

Dockerfile
eddbbt: yeah, it was trying to compile before copying files, thus the error, its still compiling after that

eddbbt: yeah, it was trying to compile before copying files, thus the error, its still compiling after that

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
eddbbt: i'll check for the existence of that piece of documentation

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
frossq: But in the changelog you say milliseconds :laughing:

frossq: (also, the changelog entry seems like a supercazzola)

apps/astarte_housekeeping/lib/astarte_housekeeping/config.ex
```diff
@@ -50,6 +50,17 @@ defmodule Astarte.Housekeeping.Config do
     type: :boolean,
     default: false
 
+  @envdoc """
+  Default maximum storage retention for new realms (in seconds).
+  Can be overridden at any time during realm creation or update
+  [Check docs for more info](https://docs.astarte-platform.org/astarte/latest/030-interface.html#datastream-specific-features)
+  """
+  app_env :default_datastream_maximum_storage_retention,
```
nedimtokic, line 58: What will happen if `HOUSEKEEPING_DEFAULT_DATASTREAM_MAXIMUM_STORAGE_RETENTION` is not exported ?

eddbbt: defaults to nil, so no override takes place and astarte behaves as usual

noaccOS (approved):

nedimtokic (approved):

noaccOS:
drafted while we figure out the correct astarte version to target to avoid accidental merges

noaccOS (approved):

Issue #101, opened 2019-11-29
Implement sensors viewer example


bettio:
Implement a JavaScript example which allows to choose a sensor and to plot a chart for the given interval. The example should also display min, max and average values.
All data should be displayed along with units and sensor names.

The example should use `cloud.astarte.genericsensors.Values` properties interface and `cloud.astarte.genericsensors.AvailableSensors` (when available). It would be useful to implement it using React as well.

See also https://github.com/astarte-platform/astarte/issues/86

Issue #100, opened 2019-11-29
Implement sensors viewer JS example


bettio:
Implement a JavaScript/HTML5 example which allows to display latest received values for each sensor. 
It should also display sensor name and units when available.

The example should use `cloud.astarte.genericsensors.Values` properties interface and `cloud.astarte.genericsensors.AvailableSensors` (when available). It would be useful to implement it using React as well.

It is required to implement it in the most simple and clear way, so it can be used as a starting point for third party users.

See also https://github.com/astarte-platform/astarte/issues/86

PR #1972, opened 2026-05-21
chore: add support for endpoints encryption in database

frossq:
Add encryptedblob value type and encrypted DEK field for device data tables
Add 'encrypted' flag for endpoints 

---

<details>

> `fixes #<issue number>`

> `depends on #<pr number>`

> ```mermaid
> gitGraph
>     branch feat1
>     checkout feat1
>     branch feat2
>     checkout feat2
>     branch feat3
>     checkout feat3
> ```

</details>

### DIFF
--- a/apps/astarte_housekeeping/priv/migrations/realm/0019_add_encryptedblob_value_individual_properties.sql
+++ b/apps/astarte_housekeeping/priv/migrations/realm/0019_add_encryptedblob_value_individual_properties.sql
@@ -1,0 +1,5 @@
+ALTER TABLE :keyspace.individual_properties
+ADD (
+   encryptedblob_value blob,
+   encrypted_dek blob
+);
--- a/apps/astarte_housekeeping/priv/migrations/realm/0020_add_encryptedblob_value_individual_datastreams.sql
+++ b/apps/astarte_housekeeping/priv/migrations/realm/0020_add_encryptedblob_value_individual_datastreams.sql
@@ -1,0 +1,5 @@
+ALTER TABLE :keyspace.individual_datastreams
+ADD (
+   encryptedblob_value blob,
+   encrypted_dek blob
+);
--- a/apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
+++ b/apps/astarte_housekeeping/lib/astarte_housekeeping/realms/queries.ex
@@ -759,6 +759,8 @@ defmodule Astarte.Housekeeping.Realms.Queries do
       stringarray_value list<varchar>,
       binaryblobarray_value list<blob>,
       datetimearray_value list<timestamp>,
+      encryptedblob_value blob,
+      encrypted_dek blob,
 
       PRIMARY KEY((device_id, interface_id), endpoint_id, path)
     )
@@ -875,6 +877,7 @@ defmodule Astarte.Housekeeping.Realms.Queries do
       description varchar,
       doc varchar,
       required boolean,
+      encrypted boolean,
 
       PRIMARY KEY ((interface_id), endpoint_id)
     );
--- a/apps/astarte_housekeeping/priv/migrations/realm/0021_add_encryption_data_to_endpoints.sql
+++ b/apps/astarte_housekeeping/priv/migrations/realm/0021_add_encryption_data_to_endpoints.sql
@@ -1,0 +1,4 @@
+ALTER TABLE :keyspace.endpoints
+ADD (
+   encrypted boolean
+);
--- a/apps/astarte_realm_management/lib/astarte_realm_management/interfaces/create_datastream_individual_multi_interface.ex
+++ b/apps/astarte_realm_management/lib/astarte_realm_management/interfaces/create_datastream_individual_multi_interface.ex
@@ -22,7 +22,7 @@ defmodule Astarte.RealmManagement.CreateDatastreamIndividualMultiInterface do
   @moduledoc """
   Migration to create individual datastreams table.
 
-  This table stores all individual datastream sample across all interfaces within a realm. 
+  This table stores all individual datastream sample across all interfaces within a realm.
   """
   use Ecto.Migration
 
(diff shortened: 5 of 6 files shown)


apps/astarte_housekeeping/priv/migrations/realm/0020_add_encryptedblob_value_individual_datastreams.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE :keyspace.individual_datastreamss
```
nedimtokic:
```suggestion
ALTER TABLE :keyspace.individual_datastreams
```

nedimtokic (changes requested):

apps/astarte_realm_management/lib/astarte_realm_management/interfaces/queries.ex
nedimtokic, line 107: Consider that we here create new table for each aggregate interface.

apps/astarte_housekeeping/priv/migrations/realm/0020_add_encryptedblob_value_individual_datastreams.sql
frossq: :open_mouth:

mizzet1 (approved):

apps/astarte_housekeeping/priv/migrations/realm/0019_add_encryptedblob_value_individual_properties.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE :keyspace.individual_properties
+ADD (
+   encryptedblob_value blob,
+   encrypted_dek blob
+);
```
noaccOS: let's make sure these files have a newline at the end

apps/astarte_housekeeping/priv/migrations/realm/0020_add_encryptedblob_value_individual_datastreams.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE :keyspace.individual_datastreams
```
nedimtokic, line 1: `individual_datastreams` table is creating on interface installation. So, this migration will fail on fresh instance.

Maybe we can add [migration](https://github.com/astarte-platform/astarte/blob/929849d9f4b7cfd968cb2e1b6ce5c6e52946ed29/apps/astarte_realm_management/lib/astarte_realm_management/interfaces/create_datastream_individual_multi_interface.ex#L30) like [here](https://github.com/astarte-platform/astarte/blob/929849d9f4b7cfd968cb2e1b6ce5c6e52946ed29/apps/astarte_realm_management/lib/astarte_realm_management/interfaces/queries.ex#L82) and add migrations in list for example:
```
Ecto.Migrator.run(Repo, [
  {0, CreateDatastreamIndividualMultiInterface},
  {1, AddEncryptedToIndividualDatastreams}
], ....)
```

frossq: Yes, I noticed unfortunately. Also using the IF EXISTS condition fails (Scylla does not support it :angry: ). For now I put a workaround but your solution seems more elegant, looking into it

noaccOS: Adding the migration to `install_interface` won't be enough, as a user might update an existing interface with a new encrypted endpoint instead of installing a new interface.

To avoid duplicating the logic throughout the codebase, I think keeping migrations alongside all other migrations (ie in housekeeping) is the better practice, this way we know that the database has already been properly migrated once housekeeping is healthy.

Creating two migrations in housekeeping (CREATE IF EXISTS + ALTER) would work :tm: but is not strictly _correct_ for Astarte's architecture: each interface is by design allowed to be installed in its own table, though this logic has not been implemented as of yet. Hence, always creating the `individual_datastreams` table would technically be incorrect.

The current solution is a workaround for the not-implemented ALTER IF EXISTS. It's not pretty but it's simple and correct (it's correct because **right now** the only allowed interface for individual datastreams **is in fact** `indivudal_datastreams`).

As it's important for this feature to be merged quickly in order to unlock other tasks, I think the current solution is acceptable.

In a following step, we can implement this by allowing custom (ie: ecto, ie: not simple cql query) migrations, which are also needed for the realm KEK and rabbit vhost.

noaccOS (approved):

apps/astarte_housekeeping/priv/migrations/realm/0020_add_encryptedblob_value_individual_datastreams.sql
nedimtokic: Altering table when it does not exists do nothing. So, if we add one more migration on startup to create "individual_datastreams" table then ok.

noaccOS: See my previous point about CREATE IF EXISTS + ALTER: we _should_ not blindly create it

apps/astarte_housekeeping/priv/migrations/realm/0019_add_encryptedblob_value_individual_properties.sql
eddbbt: @noaccOS 
>  as a user might update an existing interface with a new encrypted endpoint instead of installing a new interface.

good point, your comment is spot on. But, what if we consider the change to be breaking for older interfaces? That way, the user is forced to install a new interface (similar to what we do with the major/minor version)


ps: wrong comment, this is the one https://github.com/astarte-platform/astarte/pull/1972#discussion_r3288657705

noaccOS: Adding a new endpoint is never breaking

noaccOS (approved):
🍜

PR #1976, opened 2026-05-22
chore(secrets): add DEK support via OpenBao transit

mizzet1:
This PR aim to add DEK generation logic for the Encrypted Endpoints feature

### DIFF
--- a/libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
+++ b/libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
@@ -165,6 +165,48 @@ defmodule Astarte.Secrets do
   end
 
   @doc """
+  Generates a new Data Encryption Key (DEK) wrapped under the named transit key.
+  Optional `:bits` (128 or 256, default 256).
+  """
+  @spec generate_dek(String.t(), String.t(), keyword()) ::
+          {:ok, %{plaintext: binary(), ciphertext: String.t()}} | :error
+  def generate_dek(key_name, namespace, opts \\ []) do
+    Core.generate_dek(key_name, namespace, opts)
+  end
+
+  @doc """
+  Unwraps a DEK ciphertext using the named transit key.
+  """
+  @spec unwrap_dek(String.t(), String.t(), String.t(), keyword()) :: {:ok, binary()} | :error
+  def unwrap_dek(key_name, ciphertext, namespace, opts \\ []) do
+    client_opts = [namespace: namespace] ++ Keyword.take(opts, [:token])
+    headers = [{"Content-Type", "application/json"}]
+
+    case Client.post(
+           "/transit/decrypt/#{key_name}",
+           Jason.encode!(%{ciphertext: ciphertext}),
+           headers,
+           client_opts
+         ) do
+      {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
+        with {:ok, data} <- Core.parse_json_data(body),
+             plaintext_b64 when is_binary(plaintext_b64) <- Map.get(data, "plaintext"),
+             {:ok, plaintext} <- Base.decode64(plaintext_b64) do
+          {:ok, plaintext}
+        else
+          _ -> :error
+        end
+
+      error_resp ->
+        Logger.error(
+          "Failed to unwrap DEK with key #{key_name} in namespace #{namespace}: #{inspect(error_resp)}"
+        )
+
+        :error
+    end
+  end
+
+  @doc """
   Decrypts the provided ciphertext using OpenBao Transit Engine.
   Useful for ASYMKEX where the device encrypts a secret with the owner's RSA public key.
   """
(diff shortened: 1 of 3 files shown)


libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -121,6 +121,24 @@ defmodule Astarte.Secrets do
...
   end
 
+  @doc """
+  Generates a new Data Encryption Key (DEK) wrapped under the named transit key.
+  Requires `:namespace` in `opts`. Optional `:bits` (128 or 256, default 256).
+  """
+  @spec generate_dek(String.t(), keyword()) ::
+          {:ok, %{plaintext: binary(), ciphertext: String.t()}} | :error
+  def generate_dek(key_name, opts \\ []) do
+    Core.generate_dek(key_name, opts)
```
noaccOS: let's not blindly push down arbitrary `opts`

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -121,6 +121,24 @@ defmodule Astarte.Secrets do
     end
   end
 
+  @doc """
+  Generates a new Data Encryption Key (DEK) wrapped under the named transit key.
+  Requires `:namespace` in `opts`. Optional `:bits` (128 or 256, default 256).
```
noaccOS: if it's required why is it an optional field in opts, with opts itself being optional?

I suggest having `def generate_dek(key_name, namespace, opts \\ [])` (if we even need opts?)

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -121,6 +121,24 @@ defmodule Astarte.Secrets do
...
+    Core.generate_dek(key_name, opts)
+  end
+
+  @doc """
+  Unwraps a DEK ciphertext using the named transit key.
+  """
+  @spec unwrap_dek(String.t(), String.t(), keyword()) :: {:ok, binary()} | :error
+  def unwrap_dek(key_name, ciphertext, opts \\ []) do
+    Core.unwrap_dek(key_name, ciphertext, opts)
+  end
```
noaccOS: this function is doing nothing. I suggest placing the code from core directly here

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -84,6 +84,79 @@ defmodule Astarte.Secrets.Core do
...
+        Logger.error("Failed to unwrap DEK with key #{key_name}")
+        :error
+    end
+  end
+
+  defp post_transit(path, body, opts) do
+    headers = [{"Content-Type", "application/json"}]
+    client_opts = Keyword.take(opts, [:namespace, :token])
+    Client.post(path, Jason.encode!(body), headers, client_opts)
+  end
```
noaccOS: let's not make custom wrappers which are used by 2 of the 20 functions in this module. if you find the api to be clunky, let's find ways to improve it globally instead!

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -84,6 +84,79 @@ defmodule Astarte.Secrets.Core do
...
+
+    with {:ok, %{status_code: 200, body: body}} <-
+           post_transit("/transit/datakey/plaintext/#{key_name}", %{bits: bits}, opts),
+         {:ok, data} <- parse_json_data(body),
+         {:ok, plaintext} <- decode_base64_field(data, "plaintext"),
+         {:ok, ciphertext} <- Map.fetch(data, "ciphertext") do
+      {:ok, %{plaintext: plaintext, ciphertext: ciphertext}}
+    else
+      _ ->
+        Logger.error("Failed to generate DEK with key #{key_name}")
```
noaccOS: the error is perhaps more important than the key name, and the key name without the namespace does not add much information


```suggestion
      error ->
        Logger.error("Failed to generate DEK with key #{key_name} in namespace #{namespace}: #{inspect(error)}")
```

same for other errors

libs/astarte_secrets/test/astarte_secrets/core_test.exs
```diff
@@ -614,4 +614,97 @@ defmodule Astarte.Secrets.CoreTest do
...
+      ciphertext = "vault:v1:#{Base.encode64(:crypto.strong_rand_bytes(48))}"
+
+      body =
+        Jason.encode!(%{
+          "data" => %{"plaintext" => Base.encode64(plaintext_bin), "ciphertext" => ciphertext}
+        })
+
+      expect(Client, :post, fn _url, _body, _headers, _opts ->
+        {:ok, %HTTPoison.Response{status_code: 200, body: body}}
+      end)
```
noaccOS: why are we using stubs instead of making real calls?

noaccOS (changes requested):

noaccOS:
superseded by #1983

PR #1978, opened 2026-05-22
ci: Alert for security vulnerabilities in app/lib dependencies

nedimtokic:
Adds `mix deps.audit` as a CI job to automatically check for known
security vulnerabilities in app/lib dependencies.

## Changes

- Added `mix_audit` dependency (`~> 2.1`) to all apps and libs
- Added `deps_audit` job to `astarte-apps-build-workflow.yaml`
- Added `deps_audit` job to `astarte-libs-build-workflow.yaml`

## Details

The `deps_audit` job runs in parallel with `formatter`, `credo`, and
`dialyzer` after the `warmup` cache step. It uses the existing
`mix-task` staple action and fails the build if any dependency has a
known CVE in the GitHub Advisory Database.


noaccOS:
Would it be possible to use dependabot only for security issues but keep renovate for dependency updates?

noaccOS (approved):

PR #1983, opened 2026-05-25
chore(secrets): add DEK encrypt/decrypt payload support

mizzet1:
This PR aim to add payload encryption/decryption with DEK for the Encrypted Endpoints feature

### DIFF
--- a/libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
+++ b/libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
@@ -165,6 +165,48 @@ defmodule Astarte.Secrets do
   end
 
   @doc """
+  Generates a new Data Encryption Key (DEK) wrapped under the named transit key.
+  Optional `:bits` (128 or 256, default 256).
+  """
+  @spec generate_dek(String.t(), String.t(), keyword()) ::
+          {:ok, %{plaintext: binary(), ciphertext: String.t()}} | :error
+  def generate_dek(key_name, namespace, opts \\ []) do
+    Core.generate_dek(key_name, namespace, opts)
+  end
+
+  @doc """
+  Unwraps a DEK ciphertext using the named transit key.
+  """
+  @spec unwrap_dek(String.t(), String.t(), String.t(), keyword()) :: {:ok, binary()} | :error
+  def unwrap_dek(key_name, ciphertext, namespace, opts \\ []) do
+    client_opts = [namespace: namespace] ++ Keyword.take(opts, [:token])
+    headers = [{"Content-Type", "application/json"}]
+
+    case Client.post(
+           "/transit/decrypt/#{key_name}",
+           Jason.encode!(%{ciphertext: ciphertext}),
+           headers,
+           client_opts
+         ) do
+      {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
+        with {:ok, data} <- Core.parse_json_data(body),
+             plaintext_b64 when is_binary(plaintext_b64) <- Map.get(data, "plaintext"),
+             {:ok, plaintext} <- Base.decode64(plaintext_b64) do
+          {:ok, plaintext}
+        else
+          _ -> :error
+        end
+
+      error_resp ->
+        Logger.error(
+          "Failed to unwrap DEK with key #{key_name} in namespace #{namespace}: #{inspect(error_resp)}"
+        )
+
+        :error
+    end
+  end
+
+  @doc """
   Decrypts the provided ciphertext using OpenBao Transit Engine.
   Useful for ASYMKEX where the device encrypts a secret with the owner's RSA public key.
   """
--- a/libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
+++ b/libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
@@ -182,28 +182,45 @@ defmodule Astarte.Secrets do
     client_opts = [namespace: namespace] ++ Keyword.take(opts, [:token])
     headers = [{"Content-Type", "application/json"}]
 
-    case Client.post(
-           "/transit/decrypt/#{key_name}",
-           Jason.encode!(%{ciphertext: ciphertext}),
-           headers,
-           client_opts
-         ) do
-      {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
-        with {:ok, data} <- Core.parse_json_data(body),
-             plaintext_b64 when is_binary(plaintext_b64) <- Map.get(data, "plaintext"),
-             {:ok, plaintext} <- Base.decode64(plaintext_b64) do
-          {:ok, plaintext}
-        else
-          _ -> :error
-        end
-
-      error_resp ->
+    with {:ok, %HTTPoison.Response{status_code: 200, body: body}} <-
+           Client.post(
+             "/transit/decrypt/#{key_name}",
+             Jason.encode!(%{ciphertext: "vault:v1:" <> ciphertext}),
+             headers,
+             client_opts
+           ),
+         {:ok, data} <- Core.parse_json_data(body),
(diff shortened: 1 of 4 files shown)


libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -164,6 +164,79 @@ defmodule Astarte.Secrets do
...
+  Encrypts `payload` using AES-256-GCM with the provided plaintext DEK.
+  Generates a fresh 12-byte IV on every call. Returns
+  `{:ok, %{ciphertext: binary(), tag: binary(), iv: binary()}}` on success.
+  """
+  @spec encrypt_with_dek(binary(), binary()) ::
+          {:ok, %{ciphertext: binary(), tag: binary(), iv: binary()}} | :error
+  def encrypt_with_dek(payload, dek)
+      when is_binary(payload) and byte_size(dek) == 32 do
+    Core.encrypt_with_dek(payload, dek)
+  end
```
noaccOS: I'd remove checks altogether: we assume internal data is correct, if it's not we crash

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -164,6 +164,79 @@ defmodule Astarte.Secrets do
...
+        else
+          _ -> :error
+        end
+
+      error_resp ->
+        Logger.error(
+          "Failed to unwrap DEK with key #{key_name} in namespace #{namespace}: #{inspect(error_resp)}"
+        )
+
+        :error
```
noaccOS: as it's written, errors from the response parsing do not log anything, I'd rewrite this as a single `with` block which logs in the `else` clause

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -164,6 +164,79 @@ defmodule Astarte.Secrets do
...
+  Decrypts `ciphertext` using AES-256-GCM with the provided plaintext DEK.
+
+  Requires the `tag` (16 bytes) and `iv` (12 bytes) that were produced during
+  encryption. Returns `{:ok, plaintext}` on success, or `:error` if authentication
+  fails.
+  """
+  @spec decrypt_with_dek(binary(), binary(), binary(), binary()) ::
+          {:ok, binary()} | :error
+  def decrypt_with_dek(ciphertext, tag, iv, dek)
+      when is_binary(ciphertext) and byte_size(tag) == 16 and byte_size(iv) == 12 and
```
noaccOS: same as encrypt. also, I'd skip `iv`s for now

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -108,6 +108,117 @@ defmodule Astarte.Secrets.Core do
...
+  The `tag` (16 bytes) and `iv` must be stored alongside the `ciphertext` so the
+  payload can later be decrypted with `decrypt_with_dek/4`.
+  """
+  @spec encrypt_with_dek(binary(), binary()) ::
+          {:ok, %{ciphertext: binary(), tag: binary(), iv: binary()}}
+  def encrypt_with_dek(plaintext, dek)
+      when is_binary(plaintext) and byte_size(dek) == 32 do
+    iv = :crypto.strong_rand_bytes(12)
+
+    {ciphertext, tag} =
```
noaccOS, line 123: what's this tag?

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -108,6 +108,117 @@ defmodule Astarte.Secrets.Core do
...
+        :error
+
+      plaintext when is_binary(plaintext) ->
+        {:ok, plaintext}
+    end
+  end
+
+  # TODO: add a proper public API (Astarte.Secrets) for creating/managing the KEK
+  # (mount transit + create AES-256 key)
+  def create_encryption_key(key_name, namespace) do
```
noaccOS: is this only for tests?

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -108,6 +108,117 @@ defmodule Astarte.Secrets.Core do
...
+        Logger.error(
+          "Failed to generate DEK with key #{key_name} in namespace #{namespace}: #{inspect(error)}"
+        )
+
+        :error
+    end
+  end
+
+  defp decode_base64_field(data, field) do
+    with b64 when is_binary(b64) <- Map.get(data, field),
```
noaccOS: nit: `when` clauses are kind of hard to read, prefer proper matches when possible

in this case,

```suggestion
    with {:ok, b64} <- Map.fetch(data, field),
```

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -108,6 +108,117 @@ defmodule Astarte.Secrets.Core do
...
+    with {:ok, %{status_code: 200, body: body}} <-
+           Client.post(
+             "/transit/datakey/plaintext/#{key_name}",
+             Jason.encode!(%{bits: bits}),
+             headers,
+             client_opts
+           ),
+         {:ok, data} <- parse_json_data(body),
+         {:ok, plaintext} <- decode_base64_field(data, "plaintext"),
+         {:ok, ciphertext} <- Map.fetch(data, "ciphertext") do
```
noaccOS: is the cipher not base64 encoded?

noaccOS (changes requested):

libs/astarte_secrets/lib/astarte_secrets/core.ex
mizzet1: It is returned by crypto_one_time_aead when the encryption flag is set to true.
It's an internal value of AES-GCM used for extra security

sources:
 [crypto_one_time_aead return value](https://www.erlang.org/doc/apps/crypto/crypto.html#crypto_one_time_aead/7:~:text=Result%20%3A%3A%20EncryptResult%20%7C%20DecryptResult%2C-,EncryptResult%20%3A%3A%20%7BOutCryptoText%2C%20OutTag%7D,-%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20DecryptResult%20%3A%3A%20OutPlainText%20%7C%20error)

[tag](https://www.chilkatsoft.com/gcm_authentication_tag.asp)

libs/astarte_secrets/lib/astarte_secrets/core.ex
mizzet1: yes

libs/astarte_secrets/lib/astarte_secrets/core.ex
mizzet1: it's a vault-prefixed string

[source](https://openbao.org/api-docs/secret/transit/#generate-data-key:~:text=%22ciphertext%22%3A%20%22vault%3Av1%3Aabcdefgh%22)

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
mizzet1: ok removed and it's part of the ciphertext blob now

libs/astarte_secrets/lib/astarte_secrets/core.ex
noaccOS: let's un-vault prefix stringify it

libs/astarte_secrets/lib/astarte_secrets/core.ex
noaccOS: then let's either:
- put it in test helpers (this would be my preferred approach)
- use `@doc false` to signal it shouldn't be used

nedimtokic (approved):

noaccOS (approved):
💣

PR #1984, opened 2026-05-25
chore(secrets): add realm kek functionality

noaccOS:
revision now works for both asymmetrical and symmetrical key algorithms

parameters are wrapped inside a map, as ecto expects a map or array
to be passed from cast_embed, while for symmetrical keys vault returns
a simple integer as value.

given this situation, the chance was taken to use vault's revision ids
instead of indexes as revision id, by embedding them into the
parameters

### DIFF
--- a/libs/astarte_secrets/lib/astarte_secrets/core.ex
+++ b/libs/astarte_secrets/lib/astarte_secrets/core.ex
@@ -30,7 +30,7 @@ defmodule Astarte.Secrets.Core do
 
   require Logger
 
-  @type key_algorithm() :: :es256 | :es384 | :rs256 | :rs384
+  @type key_algorithm() :: :aes128 | :aes256 | :es256 | :es384 | :rs256 | :rs384
   @type digest_type() :: :crypto.sha1() | :crypto.sha2()
 
   # RFC 5649 AES Key Wrap with Padding magic constant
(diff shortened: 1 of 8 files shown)


mizzet1 (approved):

mizzet1 (approved):

mizzet1 (approved):

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -314,6 +328,14 @@ defmodule Astarte.Secrets.Core do
     end
   end
 
+  @doc """
+  Returns the namespace name for the realm KEKs namespace, represented as a list of tokens
+  """
+  def realm_kek_namespace_tokens do
+    ["astarte_encrypted_messages_kek", instance_tokens()]
```
frossq: Don't we want to also include the `realm` in the tokens?

nedimtokic (approved):

libs/astarte_secrets/lib/astarte_secrets/core.ex
noaccOS: do we want a namespace for the realm with a single key (the realm kek) inside it?

AliouneGaye21 (approved):

frossq (approved):

PR #1996, opened 2026-05-29
chore(secrets): add encrypted device data decryption

AliouneGaye21:

---

<details>

> `fixes #<issue number>`

> `depends on #<pr number>`

> ```mermaid
> gitGraph
>     branch feat1
>     checkout feat1
>     branch feat2
>     checkout feat2
>     branch feat3
>     checkout feat3
> ```

</details>

### DIFF
--- a/libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
+++ b/libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
@@ -227,4 +227,12 @@ defmodule Astarte.Secrets do
   def encrypt_device_data(plaintext, session_key, key_type) do
     Core.encrypt_device_data(plaintext, session_key, key_type)
   end
+
+  @doc """
+  Decrypts a COSE Encrypt0 binary payload using the shared `session_key`.
+  """
+  @spec decrypt_device_data(binary(), binary(), atom()) :: {:ok, binary()}
+  def decrypt_device_data(cbor_binary, session_key, key_type) do
+    Core.decrypt_device_data(cbor_binary, session_key, key_type)
+  end
 end
--- a/libs/astarte_secrets/lib/astarte_secrets/core.ex
+++ b/libs/astarte_secrets/lib/astarte_secrets/core.ex
@@ -617,4 +617,23 @@ defmodule Astarte.Secrets.Core do
 
     Encrypt0.encrypt_encode(msg, key_type, key, iv)
   end
+
+  @doc """
+  Decrypts a COSE Encrypt0 binary payload using the shared `session_key`.
+  The `key_type` represents the cipher suite determined during the handshake (e.g., :aes_128_gcm, :aes_256_gcm).
+  """
+  @spec decrypt_device_data(binary(), binary(), atom()) :: {:ok, binary()} | {:error, atom()}
+  def decrypt_device_data(cbor_binary, session_key, key_type) do
+    key = %{k: session_key}
+
+    with {:ok, msg} <- Encrypt0.decode_cbor(cbor_binary),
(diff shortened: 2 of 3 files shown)


libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -617,4 +617,26 @@ defmodule Astarte.Secrets.Core do
...
+  """
+  @spec decrypt_device_data(binary(), binary(), atom()) :: {:ok, binary()}
+  def decrypt_device_data(cbor_binary, session_key, key_type) do
+    {:ok, msg} = Encrypt0.decode_cbor(cbor_binary)
+
+    # Extract the IV 
+    iv =
+      case msg.uhdr.iv do
+        binary when is_binary(binary) -> binary
+        %{value: binary} -> binary
```
frossq: Can it be both cases?

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -227,4 +227,12 @@ defmodule Astarte.Secrets do
   def encrypt_device_data(plaintext, session_key, key_type) do
     Core.encrypt_device_data(plaintext, session_key, key_type)
   end
+
+  @doc """
+  Decrypts a COSE Encrypt0 binary payload using the shared `session_key`.
+  """
+  @spec decrypt_device_data(binary(), binary(), atom()) :: {:ok, binary()}
```
frossq, line 234: I'm getting confused by names, we have already a `decrypt` function above, adding a `decrypt_device_data` puts some ambiguity. Can it become something like `decrypt_cbor_payload`? Or even be moved to another module that is not Secrets (since we stated that this module is used to interact with Bao APIs).

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -617,4 +617,26 @@ defmodule Astarte.Secrets.Core do
...
+      end
+
+    key = %{k: session_key}
+
+    {:ok, decrypted_msg} = Encrypt0.decrypt(msg, key_type, key, iv)
```
noaccOS: in contrast with the encrypt operation, here we don't have ownership of the data, the device does.
this function receives an anonymous binary, so I would expect this function to _not_ be assertive here

```elixir
with {:ok, decripted_msg} <- Encrypt0.decript(...) do
  {:ok, decripted_msg.payload}
end
```

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
noaccOS: I agree with the suggestion of splitting this into more modules, eg `Astarte.Secrets.EncryptedMessages`

noaccOS (changes requested):

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
AliouneGaye21: Good,as we want the encrypt and decrypt to be in the same module. I'm opening a new PR on top of this to move them on the suggested module.

libs/astarte_secrets/lib/astarte_secrets/core.ex
noaccOS: I would allow a single representation for the iv. we can loosen up the requirement if we find we need to support multiple versions (eg different clients encode it differently)

For now let's see how it would be represented by our encode/decode and use that

noaccOS (approved):

PR #1998, opened 2026-06-03
chore(dup): InitExchange logic for Astarte-Device handshake

mizzet1:
This PR aims to add the logic for implementing the first message (InitExchange) to start the handshake between Astarte and a Device (or viceversa) for the Encrypted Endpoints feature


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,159 @@
...
+      {:error, _reason} -> {:error, :invalid_payload}
+    end
+  end
+
+  defp parse([
+         key_type,
+         %CBOR.Tag{tag: :bytes, value: ec_public_key},
+         %CBOR.Tag{tag: :bytes, value: hkdf_salt},
+         %CBOR.Tag{tag: :bytes, value: nonce}
+       ]) do
```
noaccOS: nit: usually it's better to avoid complex clauses in function argument definitions, when they are not needed to distinguish between function implementations.

in this case, I'd suggest moving the `list of 4 elements` and `params are binaries` checks to the `with` block

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -1,7 +1,7 @@
 #
 # This file is part of Astarte.
 #
-# Copyright 2025 SECO Mind Srl
```
noaccOS, line 4: 2025 - 2026

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,159 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.KeyAgreement.InitExchange do
```
noaccOS, line 21: I'm not sold this should belong in DUP
But we also don't have a generic "device communication" library so I guess it's ok

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -149,6 +148,61 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
+
+  On success the message is acknowledged and the message-stats counters are
+  incremented.  On any parse or validation error the message is discarded
+  and error triggers are executed
+  """
+  def handle_control(state, "/keyAgreement", payload, timestamp) do
+    new_state = TimeBasedActions.execute_time_based_actions(state, timestamp)
+
+    case InitExchange.decode(payload) do
+      {:ok, _init_exchange} ->
```
noaccOS, line 161: I'd leave the TODO here, as this implementation still doesn't make use of this value

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -171,6 +225,71 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
+
+  Returns `{:ok, private_key, %InitExchange{}}` so the caller can retain the
+  ephemeral private key for the subsequent shared-secret derivation step
+  """
+  @spec send_init_exchange(String.t(), binary()) ::
+          {:ok, binary(), InitExchange.t()} | {:error, term()}
+  def send_init_exchange(realm, device_id) do
+    with {:ok, init_exchange, private_key} <- InitExchange.new(),
+         :ok <- publish_init_exchange(realm, device_id, init_exchange) do
+      {:ok, private_key, init_exchange}
```
noaccOS: as this return value is pretty much identical to the value returned by `InitExchange.new`, I suggest keeping the order of parameters in the records consistent with each other to avoid mistakes

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,159 @@
...
+    end
+  end
+
+  defp parse(_), do: {:error, :invalid_payload}
+
+  defp unwrap_bytes(%CBOR.Tag{tag: :bytes, value: value}), do: {:ok, value}
+  defp unwrap_bytes(_), do: {:error, :invalid_payload}
+
+  defp validate_key_type(@key_type), do: :ok
+  defp validate_key_type(_other), do: {:error, :unsupported_key_type}
```
noaccOS: how about parsing the key type instead? ie, having an enum of "values from device" -> "internal atoms"
also, while `"ECDH_X25519-HKDF_SHA256-AES_256_GCM"`, what we receive from the device is actually an integer

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,159 @@
...
+  defp validate_hkdf_salt(_), do: {:error, :invalid_hkdf_salt}
+
+  defp validate_nonce(nonce) when byte_size(nonce) == @nonce_size, do: :ok
+  defp validate_nonce(_), do: {:error, :invalid_nonce}
+
+  @doc """
+  Returns the only accepted key-agreement suite identifier for this protocol.
+  """
+  @spec key_type() :: String.t()
+  def key_type, do: @key_type
```
noaccOS: while right now we only have one supported type, this won't necessarily be true in the future
so I wouldn't hardcode a single value here

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
noaccOS: still valid

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,177 @@
...
+
+  @doc """
+  Builds a new `%InitExchange{}` with a freshly generated ephemeral X25519
+  key pair, a random HKDF salt, and a random AES-GCM nonce.
+
+  Returns `{:ok, %InitExchange{}, private_key}` so the caller can store the
+  private key for the shared-secret derivation.
+  """
+  @spec new() :: {:ok, t(), binary()}
+  def new do
```
noaccOS:
```suggestion
  def new(key_type \\ :ecdh_x25519_hkdf_sha256_aes_256_gcm) do
```

this way the "default key suite" is constrained to this function instead of being exposed as a module attribute

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/control_handler_test.exs
```diff
@@ -222,10 +224,22 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandlerTest do
...
+    defp valid_key_agreement_payload do
+      # Well-formed CBOR payload that satisfies all validation rules.
+      # key_type 0 = :ecdh_x25519_hkdf_sha256_aes_256_gcm
+      CBOR.encode([
+        0,
+        %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(32)},
+        %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(32)},
+        %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(12)}
+      ])
+    end
```
noaccOS: 1. let's use a setup to pass down setup values to tests instead
2. let's use internal structures instead of anonymous cbor encoded lists to make this simpler to maintain

```suggestion
    setup do
      init_exchange =
        %InitExchange{
          key_type: :ecdh_x25519_hkdf_sha256_aes_256_gcm,
          ec_public_key: :crypto.strong_rand_bytes(32)
          hkdf_salt: :crypto.strong_rand_bytes(32)
          nonce: :crypto.strong_rand_bytes(12)
        }
      # Or, as we've done it already
      init_exchange = InitExchange.new()

      init_exchange_payload = InitExchange.cbor_encode(init_exchange)

      %{init_exchange: init_exchange, init_exchange_payload: init_exchange_payload}
    end
```

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/control_handler_test.exs
```diff
@@ -234,6 +248,167 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandlerTest do
...
+         context do
+      %{state: state} = context
+
+      payload =
+        CBOR.encode([
+          99,
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(32)},
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(32)},
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(12)}
+        ])
```
noaccOS: this is quite complex to follow as there is a lot of low level detail (payload format) in the "how does my application behave for invalid key type?" test

prefer receiving this value from outside (a setup is a great place to put this)


```suggestion
      %{state: state, invalid_key_type_payload: payload} = context
```

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,180 @@
...
+  @typedoc """
+  Internal atom representing a supported key-agreement suite.
+  """
+  @type key_suite :: :ecdh_x25519_hkdf_sha256_aes_256_gcm
+
+  typedstruct enforce: true do
+    @typedoc "InitExchange key-agreement message."
+
+    field :key_type, key_suite()
+    field :ec_public_key, binary()
```
noaccOS: maybe just public_key, as this can technically be an rsa or anything else (type is dependent on key_type)


```suggestion
    field :public_key, binary()
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,180 @@
...
+  @doc """
+  Returns the list representation of an `%InitExchange{}` ready for CBOR encoding.
+  """
+  @spec encode(t()) :: list()
+  def encode(%__MODULE__{} = msg) do
+    {:ok, key_type_id} = Ecto.Type.dump(@key_suites, msg.key_type)
+
+    [
+      key_type_id,
+      %CBOR.Tag{tag: :bytes, value: msg.ec_public_key},
```
noaccOS: wait I think this was supposed to be a COSE public key, not the simple binary?

see https://www.rfc-editor.org/info/rfc9679/#name-example

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/control_handler_test.exs
```diff
@@ -234,6 +254,159 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandlerTest do
...
+      %{state: state} = context
+
+      payload =
+        CBOR.encode([
+          0,
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(32)},
+          # 16 bytes instead of the required 32
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(16)},
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(12)}
+        ])
```
noaccOS: all these should be part of the setup (a setup_all would be fine as well, we don't really need distinct values for each test)

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/control_handler_test.exs
```diff
@@ -234,6 +254,159 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandlerTest do
...
+      %{state: state} = context
+
+      payload =
+        CBOR.encode([
+          0,
+          # 16 bytes instead of the required 32
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(16)},
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(32)},
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(12)}
+        ])
```
noaccOS: same as above

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/control_handler_test.exs
```diff
@@ -222,10 +224,28 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandlerTest do
   end
 
   describe "/keyAgreement" do
-    test "acks any payload and increments message counters", context do
-      %{state: state} = context
+    setup do
```
noaccOS: this can be the setup_all outside the `describe`, it's ok because it doesn't perform any side effects and is only setting up the context

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/control_handler_test.exs
```diff
@@ -234,6 +254,159 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandlerTest do
...
+      %{state: state} = context
+
+      payload =
+        CBOR.encode([
+          0,
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(32)},
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(32)},
+          # 8 bytes instead of the required 12
+          %CBOR.Tag{tag: :bytes, value: :crypto.strong_rand_bytes(8)}
+        ])
```
noaccOS: in setup

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/control_handler_test.exs
```diff
@@ -243,4 +416,55 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandlerTest do
...
+
+      realm = state.realm
+      device_id = state.device_id
+
+      expect(VMQPlugin, :publish, fn topic, payload_bytes, qos ->
+        encoded_device_id = Astarte.Core.Device.encode_device_id(device_id)
+
+        assert topic == "#{realm}/#{encoded_device_id}/control/keyAgreement"
+        assert qos == 2
+        assert is_binary(payload_bytes)
```
noaccOS: I think this is redundant given the following line?


```suggestion
```

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
mizzet1: That's right! need to construct the COSE_Key map (kty, crv, x) instead of passing the binary of the X25519 key

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,186 @@
...
+  @typedoc """
+  Internal atom representing a supported key-agreement suite.
+  """
+  @type key_suite :: :ecdh_x25519_hkdf_sha256_aes_256_gcm
+
+  typedstruct enforce: true do
+    @typedoc "InitExchange key-agreement message."
+
+    field :key_type, key_suite()
+    field :public_key, binary()
```
noaccOS: Let's use the whole key instead of the binary. This module is doing too much, let the cose key handle itself


```suggestion
    field :public_key, Cose.Keys.Key.t() # maybe use an alias
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,186 @@
...
+  @doc """
+  Builds a new `%InitExchange{}` with a freshly generated ephemeral X25519
+  key pair, a random HKDF salt, and a random AES-GCM nonce.
+
+  Returns `{:ok, %InitExchange{}, private_key}` so the caller can store the
+  private key for the shared-secret derivation.
+  """
+  @spec new() :: {:ok, t(), binary()}
+  def new(key_type \\ :ecdh_x25519_hkdf_sha256_aes_256_gcm) do
+    %OKP{x: public_key, d: private_key} = OKP.generate(:enc)
```
noaccOS: this should be an ecc for ecdh_* key types


```suggestion
    key = ECC.generate(:es256)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,186 @@
...
+  end
+
+  @doc """
+  Returns the list representation of an `%InitExchange{}` ready for CBOR encoding.
+  """
+  @spec encode(t()) :: list()
+  def encode(%__MODULE__{} = msg) do
+    {:ok, key_type_id} = Ecto.Type.dump(@key_suites, msg.key_type)
+    cose_key = %OKP{kty: :okp, crv: :x25519, x: msg.public_key}
+    cose_key_bytes = COSE.Keys.encode_cbor(cose_key)
```
noaccOS:
```suggestion
    cose_key_bytes = COSE.Keys.encode_cbor(msg.public_key)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,186 @@
...
+    case Ecto.Type.cast(@key_suites, int) do
+      {:ok, suite} -> {:ok, suite}
+      :error -> {:error, :unsupported_key_type}
+      {:error, _} -> {:error, :unsupported_key_type}
+    end
+  end
+
+  defp parse_key_suite(_), do: {:error, :unsupported_key_type}
+
+  defp extract_x25519_pub(%{1 => 1, -1 => 4, -2 => %CBOR.Tag{tag: :bytes, value: pub_key}}),
```
noaccOS: reading this, this has no meaning for me
pls let's name parameters

if (when) we add support for more algs, parsing the key like this will become way too much 🍝  and we'll need to add a new module for the key used here. it may make sense to create it already, even if for now it would be really simple

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,186 @@
...
+    do: {:ok, pub_key}
+
+  defp extract_x25519_pub(_), do: {:error, :invalid_cose_key}
+
+  defp validate_public_key(key) when byte_size(key) == @public_key_size,
+    do: :ok
+
+  defp validate_public_key(_), do: {:error, :invalid_public_key}
```
noaccOS: this practically works but it's logically incorrect, because the key size depends on the alg and is not a constant. it should be part of the extract public key function

```elixir
defp to_cose_key(%{@key_type => @ecc, @crv => crv, @ecc_x => ecc_x, @ecc_y => ecc_y}) do
  with {:ok, alg} <- alg_from_crv(crv),
       {:ok, expected_size} <- expected_ecc_coordinate_size(alg),
       {:ok, ecc_x} <- parse_ecc_coordinate(ecc_x, expected_size),
       {:ok, ecc_y} <- parse_ecc_coordinate(ecc_y, expected_size) do
    cose_key = 
      %COSE.Key.ECC{
        kty: :ecc,
        crv: alg, # from a quick look, I think we're use the same format here?
        alg: alg,
        x: ecc_x,
        y: ecc_y
      }

    {:ok, cose_key}
  end
end
```

(this may indeed be cleaner on its own file)

apps/astarte_data_updater_plant/test/astarte_data_updater_plant/data_updater/core/control_handler_test.exs
```diff
@@ -234,6 +284,132 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandlerTest do
...
+
+      expect(Core.Trigger, :execute_device_error_triggers, fn _state,
+                                                              "key_agreement_error",
+                                                              _meta,
+                                                              _ts ->
+        :ok
+      end)
+
+      assert {{:discard, _result, new_state, {:continue, continue_arg}}, _log} =
+               with_log(fn ->
```
noaccOS: why are we using with_log and then discarding the log?

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,228 @@
...
+  def new(key_type \\ :ecdh_x25519_hkdf_sha256_aes_256_gcm)
+
+  def new(:ecdh_x25519_hkdf_sha256_aes_256_gcm) do
+    %OKP{x: public_key, d: private_key} = OKP.generate(:enc)
+    hkdf_salt = :crypto.strong_rand_bytes(@hkdf_salt_size)
+    nonce = :crypto.strong_rand_bytes(@nonce_size)
+
+    msg = %__MODULE__{
+      key_type: :ecdh_x25519_hkdf_sha256_aes_256_gcm,
+      public_key: public_key,
```
noaccOS: I think this should be the whole `%OKP{}` struct instead

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,228 @@
...
+  end
+
+  def new(:ecdh_p256_hkdf_sha256_aes_256_gcm) do
+    %ECC{x: x, y: y, d: private_key} = ECC.generate(:es256)
+    hkdf_salt = :crypto.strong_rand_bytes(@hkdf_salt_size)
+    nonce = :crypto.strong_rand_bytes(@nonce_size)
+
+    msg = %__MODULE__{
+      key_type: :ecdh_p256_hkdf_sha256_aes_256_gcm,
+      public_key: <<4, x::binary, y::binary>>,
```
noaccOS: same as above

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,228 @@
...
+    {:ok, msg, private_key}
+  end
+
+  @doc """
+  Returns the list representation of an `%InitExchange{}` ready for CBOR encoding.
+  """
+  @spec encode(t()) :: list()
+  def encode(%__MODULE__{key_type: :ecdh_x25519_hkdf_sha256_aes_256_gcm} = msg) do
+    {:ok, key_type_id} = Ecto.Type.dump(@key_suites, msg.key_type)
+    cose_key_bytes = COSE.Keys.encode_cbor(%OKP{kty: :okp, crv: :x25519, x: msg.public_key})
```
noaccOS:
```suggestion
    cose_key_bytes = COSE.Keys.encode_cbor(msg.public_key)
```

you probably don't even need to create separate bodies for each `key_type`, the generic part is handled by `Ecto.Type` and `Cose.Keys`

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,228 @@
...
+         public_key: raw_public_key,
+         hkdf_salt: hkdf_salt,
+         nonce: nonce
+       }}
+    end
+  end
+
+  defp parse(_), do: {:error, :invalid_payload}
+
+  defp decode_cose_key(:ecdh_x25519_hkdf_sha256_aes_256_gcm, cose_key_map) do
```
noaccOS: nit: I would separate the decode and validation steps:

first we just `COSE.Keys.decode(cose_key_map)`, then if valid we validate that the key is compatible with `key_type`

noaccOS (changes requested):

eddbbt (approved):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,212 @@
...
+  """
+  @type key_suite ::
+          :ecdh_x25519_hkdf_sha256_aes_256_gcm
+          | :ecdh_p256_hkdf_sha256_aes_256_gcm
+
+  typedstruct enforce: true do
+    @typedoc "InitExchange key-agreement message."
+
+    field :key_type, key_suite()
+    field :public_key, %OKP{} | %ECC{}
```
noaccOS:
```suggestion
    field :public_key, COSE.Keys.Key.t()
```

and in general, don't use structs when you want a typespec (credo warns you when you do it in function specs)

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,212 @@
...
+  key pair, a random HKDF salt, and a random AES-GCM nonce.
+
+  Returns `{:ok, %InitExchange{}, private_key}` so the caller can store the
+  private key for the shared-secret derivation.
+  """
+  @spec new(key_suite()) :: {:ok, t(), binary()}
+  def new(key_type \\ :ecdh_x25519_hkdf_sha256_aes_256_gcm)
+
+  def new(:ecdh_x25519_hkdf_sha256_aes_256_gcm) do
+    %OKP{x: x, d: private_key} = OKP.generate(:enc)
```
noaccOS:
```suggestion
    key = OKP.generate(:enc)
```

we don't care that we also have the private key field set, we don't use it when we encode the key

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,212 @@
...
+      public_key: %OKP{kty: :okp, crv: :x25519, x: x},
+      hkdf_salt: hkdf_salt,
+      nonce: nonce
+    }
+
+    {:ok, msg, private_key}
+  end
+
+  def new(:ecdh_p256_hkdf_sha256_aes_256_gcm) do
+    %ECC{x: x, y: y, d: private_key} = ECC.generate(:es256)
```
noaccOS:
```suggestion
    key = ECC.generate(:es256)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,212 @@
...
+
+  defp decode_cose_key(key_suite, cose_key_map) do
+    with {:ok, cose_key} <- COSE.Keys.decode(cose_key_map),
+         :ok <- validate_key_suite_compatibility(key_suite, cose_key) do
+      {:ok, cose_key}
+    end
+  end
+
+  defp validate_key_suite_compatibility(:ecdh_x25519_hkdf_sha256_aes_256_gcm, %OKP{}), do: :ok
+  defp validate_key_suite_compatibility(:ecdh_p256_hkdf_sha256_aes_256_gcm, %ECC{}), do: :ok
```
noaccOS: should this also validate that it's a p256?


```suggestion
  defp validate_key_suite_compatibility(:ecdh_p256_hkdf_sha256_aes_256_gcm, %ECC{crv: :p256}), do: :ok
```

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,217 @@
...
+  def new(key_type \\ :ecdh_x25519_hkdf_sha256_aes_256_gcm)
+
+  def new(:ecdh_x25519_hkdf_sha256_aes_256_gcm) do
+    key = OKP.generate(:enc)
+    hkdf_salt = :crypto.strong_rand_bytes(@hkdf_salt_size)
+    nonce = :crypto.strong_rand_bytes(@nonce_size)
+
+    msg = %__MODULE__{
+      key_type: :ecdh_x25519_hkdf_sha256_aes_256_gcm,
+      public_key: %OKP{kty: :okp, crv: :x25519, x: key.x},
```
noaccOS:
```suggestion
      public_key: key,
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,217 @@
...
+  end
+
+  def new(:ecdh_p256_hkdf_sha256_aes_256_gcm) do
+    key = ECC.generate(:es256)
+    hkdf_salt = :crypto.strong_rand_bytes(@hkdf_salt_size)
+    nonce = :crypto.strong_rand_bytes(@nonce_size)
+
+    msg = %__MODULE__{
+      key_type: :ecdh_p256_hkdf_sha256_aes_256_gcm,
+      public_key: %ECC{kty: :ecc, crv: :p256, alg: :es256, x: key.x, y: key.y},
```
noaccOS:
```suggestion
      public_key: key,
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/init_exchange.ex
```diff
@@ -0,0 +1,217 @@
...
+  end
+
+  @doc """
+  Builds a new `%InitExchange{}` with a freshly generated ephemeral X25519
+  key pair, a random HKDF salt, and a random AES-GCM nonce.
+
+  Returns `{:ok, %InitExchange{}, private_key}` so the caller can store the
+  private key for the shared-secret derivation.
+  """
+  @spec new(key_suite()) :: {:ok, t(), binary()}
```
noaccOS: why is this returning `{:ok, _}`? if the operation is not fallible, we can return the message itself (?)
also, the key is part of the message, why do we need to return it separately?

```suggestion
  @spec new(key_suite()) :: t()
```

noaccOS (changes requested):

noaccOS (approved):

PR #2002, opened 2026-06-03
refactor(secrets): stop handling vault signatures

mizzet1:
This PR re add the "vault:v1"prefix to the string returned by OpenBao for the ciphertext


noaccOS (approved):
Only small nit: this is not so much a fix as it is a refactor: you're changing responsibilities but the workflow worked before just as well

edit: also describe the what not the how

PR #2005, opened 2026-06-05
chore(dup): add cache for storage of data encryption keys (DEKs)

frossq:
Setup a new cache for storing and retrieving current realm DEK (plaintext + ciphertext)
to be used when saving encrypted data to DB. A TTL is set to allow DEKs to expire.
The cache automatically retrieves new DEKs upon expiration
    

---

<details>

> `fixes #<issue number>`

> `depends on #<pr number>`

> ```mermaid
> gitGraph
>     branch feat1
>     checkout feat1
>     branch feat2
>     checkout feat2
>     branch feat3
>     checkout feat3
> ```

</details>


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_encryption_key_cache.ex
```diff
@@ -0,0 +1,52 @@
...
+    ConCache.update(@dek_cache_name, :current_dek, fn _ -> {:ok, dek_keypair} end)
+  end
+
+  @spec fetch_data_encryption_key() :: {:ok, map()} | {:error, :not_found}
+  def fetch_data_encryption_key do
+    case ConCache.get(@dek_cache_name, :current_dek) do
+      nil -> {:error, :not_found}
+      val -> {:ok, val}
+    end
+  end
```
noaccOS: I suggest having a single function instead:

```elixir
# realm_name is compulsory as each realm has its own kek, but perhaps scoping the dek to the
# device_id would be even better, to reduce the impact radius of a leaked key to a single device
def fetch_dek(realm_name) do 
  ConCache.fetch_or_store(@dek_cache_name, realm_name, fn -> fetch_new_dek(realm_name) end)
end

defp get_new_dek(realm_name) do
  case Secrets.generate_dek(realm_name, why_should_the_caller_know_the_namespace_here?) do
    {:ok, result} -> {:ok, result} # we may choose a different representation?
    :error -> {:error, :dek_generation_error}
  end
end
```

no reason to leave the responsibility of updating the dek to the caller

apps/astarte_data_updater_plant/config/test.exs
```diff
@@ -16,3 +16,5 @@ config :astarte_data_updater_plant, :amqp_data_queue_range_end, 0
 config :astarte_data_updater_plant, :amqp_data_queue_range_start, 0
 
 config :astarte_events, :connection_backoff, 0
+
+config :astarte_data_updater_plant, :dek_cache_ttl_seconds, 3
```
noaccOS: why do we need a distinct ttl for tests?

noaccOS (changes requested):

apps/astarte_data_updater_plant/config/test.exs
frossq: _If_ we want to test that the TTL is actually applied, I made it very short so that the test suite does not take 5 minutes to run

noaccOS: I think testing that is concache's responsibility

frossq: Reworked a bit, now the TTL expiration check in tests may be more meaningful. If not I'll remove the redundant test

eddbbt (approved):

mizzet1 (approved):

apps/astarte_data_updater_plant/mix.exs
```diff
@@ -63,7 +63,8 @@ defmodule Astarte.DataUpdaterPlant.Mixfile do
       {:astarte_generators, github: "astarte-platform/astarte_generators", only: [:dev, :test]},
       {:astarte_realm_management,
        path: "../astarte_realm_management", only: :test, runtime: false},
-      {:astarte_events, path: astarte_lib("astarte_events")}
+      {:astarte_events, path: astarte_lib("astarte_events")},
+      {:astarte_secrets, path: astarte_lib("astarte_secrets")}
```
noaccOS, line 67: when adding astarte_secrets you also need to use the init method on its config otherwise it won't work on docker compose / kubernetes (you don't see it now because it's not generating deks at any point)

see also https://github.com/astarte-platform/astarte/pull/2018

noaccOS (changes requested):

noaccOS (approved):

noaccOS (approved):

PR #2006, opened 2026-06-05
fix: Remove vulnerable cowlib dependency

nedimtokic:
Since vulnerability scanner alerts on `cowlib`,it is removed by replacing `plug_cowboy` with`bandit` HTTP server


noaccOS (approved):
♟️

PR #2023, opened 2026-06-10
feat(dup): Decrypt incoming data

nedimtokic:
When device send data on encrypted endpoint Astarte secret will decrypt it before casting and validation.


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -63,6 +64,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
          {:ok, interface_descriptor, context} <- maybe_handle_cache_miss(context),
          :ok <- can_write_on_interface?(context, interface_descriptor.ownership),
          {:ok, mapping} <- resolve_path(context, interface_descriptor),
+         {:ok, context} <- do_decrypt_payload(context, mapping),
```
frossq: Shouldn't it go after the bson decode and be performed only on the `value`?

noaccOS: :yes:

nedimtokic: Changed, thanks.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -627,4 +629,25 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
     do: realm_max_ttl
 
   defp max_ttl(_, _, _), do: nil
+
+  defp do_decrypt_payload(context, %Mapping{encrypted: true}) do
+    %{state: %{realm: realm, device_id: device_id}, payload: encrypted_payload} = context
+
+    case Secrets.decrypt_payload(realm, device_id, encrypted_payload) do
```
noaccOS: this should be using the astarte<->device shared secret generated during the handshake
the dek will be used later to re-encrypt the value
the session key is stored in the data updater process (or will be, for now let's add a todo and use `device_shared_secret = nil` or something)

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -64,6 +65,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
          :ok <- can_write_on_interface?(context, interface_descriptor.ownership),
          {:ok, mapping} <- resolve_path(context, interface_descriptor),
          {:ok, {value, value_timestamp, _metadata}} <- decode_bson_payload(context),
+         {:ok, value} <- do_decrypt_payload(value, mapping),
```
noaccOS: here we're passing `value` but the function expects `context`
I don't mind either passing it context or just the shared secret, let's just be consistent

libs/astarte_secrets/lib/astarte_secrets/core.ex
```diff
@@ -415,6 +415,14 @@ defmodule Astarte.Secrets.Core do
     |> List.flatten()
   end
 
+  @doc """
+  Returns the namespace name for the device session keys namespace, represented as a list of tokens
+  """
+  def device_session_key_namespace_tokens(realm_name, device_id) do
```
noaccOS: shared secrets with the device are not stored on openbao
this is because
1. we don't need it, a binary is simple enough
2. we don't need to call bao for each and every device data we receive
3. storing one secret per device is A LOT of keys on bao. on vault you pay good money for those

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
nedimtokic: Passed both values now, for logging purpose.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
nedimtokic: Changed to use erlang :crypto to decrypt value.

libs/astarte_secrets/lib/astarte_secrets/core.ex
nedimtokic: Removed.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/payloads_decoder.ex
```diff
@@ -30,7 +30,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.PayloadsDecoder do
   reception_timestamp is used if no timestamp has been sent with the payload.
   """
   @spec decode_bson_payload(binary, integer) ::
-          {:ok, {map, integer, map}}
+          {:ok, {term, integer, map}}
```
noaccOS, line 33: nit: this is a bit generic and does not really help users of this function

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -627,4 +629,38 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
     do: realm_max_ttl
 
   defp max_ttl(_, _, _), do: nil
+
+  defp do_decrypt_payload(value, %Mapping{encrypted: true}, context, key)
+       when is_binary(value) and is_binary(key) do
+    case value do
+      <<iv::binary-12, tag::binary-16, ciphertext::binary>> ->
+        decrypted_payload =
+          :crypto.crypto_one_time_aead(:aes_256_gcm, key, iv, ciphertext, <<>>, tag, false)
```
noaccOS, line 638: I think iv size, tag size and algorithm depend on the shared secret
this will probably be refactored later into its own module but for now it's ok

noaccOS (approved):

frossq (approved):

PR #2029, opened 2026-06-15
feat(appengine): Decrypt stored values

nedimtokic:
Decrypt values when endpoint flag `encrypted` is set to true


apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -929,6 +945,37 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
+         {:ok, decrypted_value} <- Secrets.decrypt_with_dek(encrypted_value, unwrapped_dek) do
+      %{row | encryptedblob_value: decrypted_value}
+    else
+      error ->
+        Logger.error(
+          "Failed to decrypt value: #{inspect(error)}",
+          tag: "decryption_error"
+        )
+
+        row
```
noaccOS: this function can fail, eg if the key was deleted or if the http call fails. it should fail hard instead of returning the row. I suggest returning the error here, and `{:ok, updated_row}` on the ok branch

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -929,6 +945,37 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
+  defp maybe_decrypt_result(result, _, _), do: result
+
+  defp decrypt_row(row, realm_name) do
+    dek = row.encrypted_dek
+    encrypted_value = row.encryptedblob_value
+
+    with {:ok, key} <- Secrets.fetch_realm_kek(realm_name),
+         {:ok, unwrapped_dek} <- Secrets.unwrap_dek(key.name, dek, key.namespace),
+         {:ok, decrypted_value} <- Secrets.decrypt_with_dek(encrypted_value, unwrapped_dek) do
+      %{row | encryptedblob_value: decrypted_value}
```
noaccOS: how about populating the appropriate column instead? such as `integer_value`

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -929,6 +945,37 @@ defmodule Astarte.AppEngine.API.Device.Queries do
     end
   end
 
+  defp maybe_decrypt_values(values, %{encrypted: true}, realm_name) do
+    Enum.map(values, &decrypt_row(&1, realm_name))
```
noaccOS: while this is _correct_, this is by no means _efficient_, as each key is fetched from bao only to be forgotten on the next instruction

I suggest using something like this instead

```elixir
  defp maybe_decrypt_values(values, %{encrypted: true}, realm_name) do
    with {:ok, keys} <- fetch_deks(values) do
      decrypt_values(keys, values)
        key = Map.fetch!(cache, entry.encrypted_dek)
        decrypted = decrypt_row(entry, key)
    end
  end

  defp decrypt_values(keys, values) do
    decrypted = values |> Enum.map(&decrypt_row(keys, &1))

    case Enum.find(decrypted, & &1 == :error) do
      nil ->
        decrypted = decrypted |> Enum.map(fn {:ok, value} -> value)
        {:ok, decrypted}

      :error ->
        {:error, :decrypt_error}
    end
  end

  defp fetch_deks(data) do
    data
    |> Enum.map(& &1.encrypted_dek)
    |> Enum.uniq()
    |> Enum.reduce_while({:ok, %{}}, fn encrypted_dek, {:ok, cache} ->
      case fetch_dek(encrypted_dek) do
        {:ok, dek} -> {:cont, {:ok, Map.put(cache, encrypted_dek, dek)}}
        error -> {:halt, error}
      end
    end)
  end

  defp fetch_dek(encrypted_dek) do
    # your fetch function, though I'd like to hide the realm kek behind secret's api but
    # this is what we have for now
  end

  defp decrypt_row(keys, row) do
    key = Map.fetch!(keys, row.encrypted_dek)
    Secrets.decrypt_with_dek(row.encrypted_value, key)
  end
```

which results in each key only being fetched once (and handles errors)

noaccOS (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
nedimtokic: Added `CQLUtils.type_to_db_column_name(row.value_type) |> String.to_atom() ` instead of `encryptedblob_value`

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
nedimtokic: Changed.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -268,6 +270,7 @@ defmodule Astarte.AppEngine.API.Device do
     if db_max_ttl, do: [ttl: db_max_ttl], else: []
   end
 
+  # Helper to handle database persistence logic
```
noaccOS: duplicated comment?

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -40,6 +40,8 @@ defmodule Astarte.AppEngine.API.Device do
   alias Astarte.DataAccess.Device.InsertContext
   alias Astarte.DataAccess.Interface, as: InterfaceQueries
   alias Astarte.DataAccess.Mappings
+  alias Astarte.Secrets
+  alias Astarte.Secrets.Core
```
noaccOS, line 44: Ideally we shouldn't need to access `Core` modules from outside `Astarte.Secrets`
I suggest adding a TODO to not require the kek when generating the dek, but instead having astarte secrets know how to do it

libs/astarte_data_access/lib/astarte_data_access/device/insert_context.ex
```diff
@@ -36,6 +36,8 @@ defmodule Astarte.DataAccess.Device.InsertContext do
     field :value, term()
     field :value_timestamp, non_neg_integer()
     field :reception_timestamp, non_neg_integer()
+    field :encrypted_endpoints, list(String.t()), default: []
```
noaccOS: I'm not sure having the caller be required to populate this field is the best design, I think it would be better if the function to insert the data could infer it instead

noaccOS (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Added TODO

libs/astarte_data_access/lib/astarte_data_access/device/insert_context.ex
nedimtokic: Removed the field and used mapping to check if there are any encrypted endpoints.

noaccOS (approved):

AliouneGaye21 (approved):

PR #2034, opened 2026-06-17
chore(dup): add shared secret computation module

mizzet1:
This PR aim to add a module for handling the shared secret computation logic for the handshake protocol of the encrypted endpoints feature


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/crypto.ex
```diff
@@ -0,0 +1,78 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.KeyAgreement.Crypto do
```
noaccOS: IMO having objects instead of "collections of functions" is clearer
So instead of `Crypto.derive_shared_secret` I personally would prefer `SharedSecret.derive`

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/shared_secret.ex
```diff
@@ -0,0 +1,78 @@
...
+
+  @hkdf_info "astarte-kdf"
+  @x25519_key_size 32
+  @p256_coordinate_size 32
+
+  @doc """
+  Derives a 256-bit AES-GCM symmetric key using the given COSE keys and salt.
+  Supports both X25519 (OKP) and P-256 (ECC).
+  """
+  def derive_shared_secret(my_cose_key, peer_cose_key, salt) do
```
noaccOS: we're already on the `SharedSecret` in the name


```suggestion
  def derive(local_cose_key, peer_cose_key, salt) do
```

noaccOS (approved):
1️⃣

PR #2040, opened 2026-06-18
feat(dup): persist shared secret

eddbbt:
add functions and fields to persist shared secret in the db

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
@@ -217,7 +217,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
         :total_received_bytes,
         :introspection,
         :exchanged_bytes_by_interface,
-        :exchanged_msgs_by_interface
+        :exchanged_msgs_by_interface,
+        :shared_secret
       ])
       |> put_query_prefix(keyspace_name)
       |> Repo.one(consistency: Consistency.device_info(:read))
@@ -233,7 +234,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
       total_received_msgs: stats.total_received_msgs,
       total_received_bytes: stats.total_received_bytes,
       initial_interface_exchanged_bytes: stats.exchanged_bytes_by_interface,
-      initial_interface_exchanged_msgs: stats.exchanged_msgs_by_interface
+      initial_interface_exchanged_msgs: stats.exchanged_msgs_by_interface,
+      shared_secret: stats.shared_secret
     }
   end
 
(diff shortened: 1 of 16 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -59,9 +59,16 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
     # TODO this could be a bang!
     {:ok, ttl} = Queries.get_datastream_maximum_storage_retention(state.realm)
 
+    shared_secret =
+      case Queries.get_shared_secret(state.realm, device_id) do
```
noaccOS: can we fetch it with the rest of the device status instead? that way we do 1 less query on startup

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -974,6 +974,36 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
+      Device
+      |> select([d], d.shared_secret)
+      |> where(device_id: ^device_id)
+      |> put_query_prefix(keyspace_name)
+
+    case Repo.safe_fetch_one(query, consistency: Consistency.device_info(:read)) do
+      {:ok, shared_secret} -> {:ok, shared_secret}
+      {:error, :not_found} -> {:error, :not_found}
+      {:error, reason} -> {:error, reason}
+    end
```
noaccOS: nit: this would be equivalent to


```suggestion
    Repo.safe_fetch_one(query, consistency: Consistency.device_info(:read))
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -974,6 +974,36 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
...
+
+    device =
+      %Device{device_id: device_id}
+      |> Ecto.Changeset.change(%{shared_secret: shared_secret})
+
+    opts = [prefix: keyspace_name, consistency: Consistency.device_info(:write)]
+
+    with {:ok, _device} <- Repo.update(device, opts) do
+      :ok
+    end
```
noaccOS: nit: you can bang in updates if data is valid. we can assume it won't generate an invalid changeset as this is internal data


```suggestion
    Repo.update!(device, opts)
    :ok
```

noaccOS (changes requested):

noaccOS (changes requested):
the current code is good, but it's missing
1. adding the field to the device schema in data_access
2. loading the field when the data updater comes up

noaccOS (changes requested):
still missing tests

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/queries.ex
```diff
@@ -233,7 +234,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Queries do
       total_received_msgs: stats.total_received_msgs,
       total_received_bytes: stats.total_received_bytes,
       initial_interface_exchanged_bytes: stats.exchanged_bytes_by_interface,
-      initial_interface_exchanged_msgs: stats.exchanged_msgs_by_interface
+      initial_interface_exchanged_msgs: stats.exchanged_msgs_by_interface,
+      shared_secret: shared_secret
```
noaccOS: isn't this 


```suggestion
      shared_secret: stats.shared_secret
```

?

noaccOS (approved):
🚀

mizzet1 (approved):

frossq:
Saving the session key in plaintext here :open_mouth: 
Approved but we should further discuss this topic

frossq (approved):

PR #2041, opened 2026-06-18
chore(dup): Reuse insert queries from DataAccess

nedimtokic:
Move functions for insterting device values into DataAccess so it can be reused in DUP and AppEngine


frossq:
Not sure about putting these functions to handle data exchanged on interfaces in `Astarte.DataAccess.Device` (maybe something like `DeviceData` could be more appropriate?), but for the rest great work :rocket:

frossq (approved):

nedimtokic:
@frossq
> Not sure about putting these functions to handle data exchanged on interfaces in `Astarte.DataAccess.Device` (maybe something like `DeviceData` could be more appropriate?), but for the rest great work 🚀

We already have `Astarte.DataAccess.Data` module. What do you think to move insert function there?

noaccOS (approved):

frossq:
> @frossq
> 
> > Not sure about putting these functions to handle data exchanged on interfaces in `Astarte.DataAccess.Device` (maybe something like `DeviceData` could be more appropriate?), but for the rest great work 🚀
> 
> We already have `Astarte.DataAccess.Data` module. What do you think to move insert function there?

Maybe yes, but too late :D

PR #2044, opened 2026-06-19
chore(dup): add SecretHash logic for Astarte-Device handshake

mizzet1:
This PR aim to add control handler logic for the SecretHash message of the Device-Astarte handshake for the Encrypted Endpoints feature

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
@@ -27,6 +27,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
   alias Astarte.DataUpdaterPlant.DataUpdater.Core.KeyAgreement.ExchangeResp
   alias Astarte.DataUpdaterPlant.DataUpdater.Core.KeyAgreement.HandshakeState
   alias Astarte.DataUpdaterPlant.DataUpdater.Core.KeyAgreement.InitExchange
+  alias Astarte.DataUpdaterPlant.DataUpdater.Core.KeyAgreement.SecretHash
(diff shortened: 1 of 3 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -248,6 +287,57 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
+          | total_received_msgs: state.total_received_msgs + 1,
+            total_received_bytes: state.total_received_bytes + byte_size(payload)
+        }
+
+        {:ack, :ok, final_state}
+
+      {:error, reason} ->
+        Logger.error("[keyAgreement/2] Failed to renegotiate key: #{inspect(reason)}")
+        # We ack the message anyway so the broker doesn't infinitely loop toxics
+        {:ack, :ok, state}
```
noaccOS: If it's failed doesn't it make more sense to discard it instead?


```suggestion
        {:discard, reason, state}
```

frossq (approved):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -226,6 +228,43 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
+        )
+
+        context = %{
+          state: new_state,
+          payload: payload,
+          path: "/keyAgreement/2",
+          timestamp: timestamp
+        }
+
+        error = %{
```
noaccOS, line 257: maybe it would make sense to not disconnect the device here? not sure though, and it's something we can reason about later, also with the embedded team

noaccOS (approved):

PR #2049, opened 2026-06-23
chore(dup): add send exchange response and shared secret derivation calls flow

mizzet1:
This PR aim to add the send_exchange_resp call and use the returned key pair for the ECDH + HKDF shared-secret derivation step in dup control handler for the encrypted endpoints feature

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
@@ -177,20 +177,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
           %{realm: new_state.realm}
         )
 
-        # TODO: call send_exchange_resp/3 here and use the returned key pair
-        #       for the ECDH + HKDF shared-secret derivation step.
-        case HandshakeState.transition(
-               new_state.encrypted_endpoints_key,
-               {:receive_init, init_exchange}
-             ) do
-          {:ok, new_key_state} ->
-            final_state = %{
-              new_state
-              | total_received_msgs: new_state.total_received_msgs + 1,
-                total_received_bytes: new_state.total_received_bytes + byte_size(payload),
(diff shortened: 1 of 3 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -177,22 +177,35 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
-            final_state = %{
-              new_state
-              | total_received_msgs: new_state.total_received_msgs + 1,
-                total_received_bytes: new_state.total_received_bytes + byte_size(payload),
-                encrypted_endpoints_key: new_key_state
-            }
-
-            {:ack, :ok, final_state}
-
+        with {:ok, state_after_receive} <-
```
noaccOS: this could be extracted to its own function as it's pretty much self contained

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -920,4 +940,8 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
 
     Core.Error.handle_error(context, error, opts)
   end
+
+  # HandshakeState.transition/2 returns {:ok, state} | {:error, reason}.
+  defp wrap_transition_error({:ok, _} = ok), do: ok
+  defp wrap_transition_error({:error, reason}), do: {:error, {:state_transition_failed, reason}}
```
noaccOS: Instead of needing to re-wrap the error, how about editing the error body to give more details?
either returning the error as defined here or, maybe better, adding a log line like so:

```elixir
    "Handshake: tried to transition with event #{inspect(event)} from state #{inspect(current_state)}"
    |> Logger.warning()
```

nedimtokic (approved):

noaccOS (approved):
🫣

PR #2053, opened 2026-06-24
refactor(dup): simplify InitExchange struct creation

mizzet1:
This PR includes a change to remove the nonce field from the InitExchange struct, as it should be randomized for each encrypted message.
It also includes a change for which the /0 suffix is added to the InitExchange control route

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
@@ -538,14 +538,14 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
   Sends an `InitExchange` message from Astarte to a device to trigger
   a new key-agreement handshake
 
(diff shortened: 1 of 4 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -520,22 +524,30 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
            ) do
-      {:ok, %{state | encrypted_endpoints_key: new_key_state}}
+      {:ok,
+       %{
+         state
+         | encrypted_endpoints_key: new_key_state,
+           key_agreement_seq_num: seq_num + 1
+       }}
+    else
+      {:error, reason} -> {:error, reason}
```
noaccOS: this is the default behavior


```suggestion
```

noaccOS (approved):

AliouneGaye21 (approved):

noaccOS (approved):

PR #2055, opened 2026-06-29
chore(all): update httposion due to fixed CVE

eddbbt:
This PR globally updates our HTTP client dependencies (httpoison to ~> 3.0 and hackney to ~> 4.0) across all umbrella apps and libraries. Because Hackney 4.x introduces much stricter socket and header handling, this PR also includes several code fixes to resolve the regressions and test failures caused by the bump.

Dependency Upgrades:

Updated httpoison to 3.0.0 and hackney to 4.5.2.

Added a global override for tzdata ~> 1.1.3 to resolve sub-dependency version conflicts.

Centralized the HTTPoison/Hackney dependency declarations as overrides to prevent umbrella conflicts.

Hackney Strictness Regressions Fixed:

Connection Pooling: Added pool: false to the Hackney options in the Secrets, FDO Rendezvous, and Events clients. This prevents {:error, :invalid_chunk_size} crashes that occur when Hackney 4.x attempts to reuse a Keep-Alive socket for rapid sequential requests (e.g., during secret rotation tests).

Header Validation: Commented out the explicit Content-Length: 0 header in Astarte.FDO.Rendezvous.send_hello/0. Hackney 4.x strictly enforces explicit headers, which was causing 400 Bad Request errors from the FDO server when sending a 1-byte CBOR payload.

Graceful Error Handling:

Added a catch-all clause for bare :error returns in Astarte.PairingWeb.FallbackController. This ensures that upstream HTTP client failures now gracefully return a 502 Bad Gateway instead of crashing the pipeline with a 500 FunctionClauseError.

Code Quality:

Added @dialyzer {:nowarn_function, stream_next: 1} to modules using HTTPoison.Base to silence warnings generated by the newly injected macros in HTTPoison 3.0.


noaccOS (changes requested):
looks like it's failing to chunk the `rotate` request
even though we don't use it in production, did you investigate why it is reliably failing on the new hackney version?

apps/astarte_data_updater_plant/mix.exs
```diff
@@ -104,7 +104,10 @@ defmodule Astarte.DataUpdaterPlant.Mixfile do
       {:typedstruct, "~> 0.5"},
       {:credo, "~> 1.7", only: [:dev, :test], runtime: false},
       {:mix_audit, "~> 2.1", only: [:dev, :test], runtime: false},
-      {:hkdf, "~> 0.3.0"}
+      {:hkdf, "~> 0.3.0"},
+      {:httpoison, "~> 3.0", override: true},
```
noaccOS, line 109: 1. why do we need override?
2. why do we need to override all 3 dependencies
3. why are we pinning tzdata to that patch version?

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fallback_controller.ex
```diff
@@ -102,6 +102,15 @@ defmodule Astarte.PairingWeb.FallbackController do
     |> render(:missing_ownership_voucher)
   end
 
+  # Catch unhandled errors gracefully
+  def call(conn, :error) do
+    conn
+    # Or :unprocessable_entity depending on your spec
```
noaccOS: 🤖

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fallback_controller.ex
```diff
@@ -102,6 +102,15 @@ defmodule Astarte.PairingWeb.FallbackController do
...
+    |> put_status(:bad_gateway)
```
noaccOS: ???

libs/astarte_fdo/lib/rendezvous/rendezvous.ex
```diff
@@ -30,8 +30,8 @@ defmodule Astarte.FDO.Rendezvous do
 
   def send_hello do
     headers = [
-      {"Content-Type", "application/cbor"},
-      {"Content-Length", "0"}
+      {"Content-Type", "application/cbor"}
+      # {"Content-Length", "0"}
```
noaccOS: either we remove this (maybe motivating it) or we don't, let's not comment things out

libs/astarte_fdo/test/astarte_fdo/rendezvous/rendezvous.ex
```diff
@@ -30,8 +30,8 @@ defmodule Astarte.FDO.Rendezvous do
 
   def send_hello do
     headers = [
-      {"Content-Type", "application/cbor"},
-      {"Content-Length", "0"}
+      {"Content-Type", "application/cbor"}
```
noaccOS, line 33: looks like we have duplicated functions @mizzet1

noaccOS (changes requested):

libs/astarte_fdo/lib/rendezvous/rendezvous.ex
eddbbt: As told irl (maybe to someone else at this point), payload size is not always 0 and the new hackney version requires a strict correspondence with the real payload, when not explicitly set  it will calculate automatically the size and fill the field

I'll add a comment

noaccOS: more than a comment, maybe it would be better to make it explicit in the commit message so that it remains in the history. reading a "unset content-length" comment in the code will just be weird in a couple months when we've forgotten why it's there

apps/astarte_data_updater_plant/mix.exs
eddbbt: - cause sub sub dependecies might use another patch/minor, not compatible with the others
- see above
- see above

overridden and pinned versions are minimum common, consider that hackney update happened not so long ago

noaccOS: did we try to make dependencies work in subdipendencies? eg updating the dependency in cfxxl or doing mix deps.update of dependencies which require old versions? 

in general overriding is powerful but should be a last resort as we're making  dependencies use versions of httpoison/hakney/tzdata they were never designed to work with, so they may break in unexpected ways just like astarte did

frossq (approved):

mizzet1 (approved):

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fallback_controller.ex
noaccOS: this was not resolved. bad gateway is not correct as this is an application error. we may customize the error but without further information it should still be a 500

noaccOS (changes requested):

noaccOS (approved):

noaccOS (commented):
actually, some tests are still failing sparingly. also there is this warning in DUP, which does not trigger a test failure but it's still kind of bad. I'd rather not merge this until these issues are resolved

```
13:28:44.8	|ERROR| GenServer :tzdata_release_updater terminating
** (FunctionClauseError) no function clause matching in :hackney.body/1
    (hackney 4.5.2) /home/runner/work/astarte/astarte/apps/astarte_data_updater_plant/deps/hackney/src/hackney.erl:702: :hackney.body(<<31, 139, 8, 0, 0, 0, 0, 0, 2, 3, 236, 90, 203, 114, 220, 70, 150, 213, 214, 248, 138, 108, 201, 49, 34, 29, 245, 0, 234, 73, 210, 150, 218, 36, 69, 81, 178, 41, 145, 65, 82, 118, 143, 59, 58, 20, 9, 32, 171, 10, ...>>)
    (tzdata 1.1.3) lib/tzdata/http_client/hackney.ex:10: Tzdata.HTTPClient.Hackney.get/3
    (tzdata 1.1.3) lib/tzdata/data_loader.ex:11: Tzdata.DataLoader.download_new/1
    (tzdata 1.1.3) lib/tzdata/data_builder.ex:11: Tzdata.DataBuilder.load_and_save_table/0
    (tzdata 1.1.3) lib/tzdata/release_updater.ex:48: Tzdata.ReleaseUpdater.poll_for_update/0
    (tzdata 1.1.3) lib/tzdata/release_updater.ex:19: Tzdata.ReleaseUpdater.handle_info/2
    (stdlib 5.1) gen_server.erl:1077: :gen_server.try_handle_info/3
    (stdlib 5.1) gen_server.erl:1165: :gen_server.handle_msg/6
    (stdlib 5.1) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
Last message: :check_if_time_to_update
State: []  module=gen_server function=error_info/8
```

noaccOS (changes requested):

noaccOS:
let's try hackney git: https://github.com/benoitc/hackney/pull/912

noaccOS (approved):
🗾

PR #2059, opened 2026-06-30
fix(housekeeping): Allow resuming partially completed realm deletions

noaccOS:
Previously we would error when the keyspace was already deleted but the
`realm` entry was not removed.

Make the delete operation idempotent instead.


frossq (approved):

frossq:
Changelog entry required?

noaccOS:
> Changelog entry required?

I am still hoping for release please

PR #2063, opened 2026-07-01
chore(dup): add ExchangeFailed control flow logic

mizzet1:
This PR aim to add the logic to send an ExchangedFailed message back to the device when an error in the Key Agreement handshake occours


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -463,11 +555,30 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
       {:ack, :ok, final_state}
     else
       {:error, reason} ->
-        # TODO: Implement failedexchange message
-        Logger.warning("keyAgreement/1 failed, ignoring for now: #{inspect(reason)}")
+        Logger.warning("[keyAgreement/1] Processing failed: #{inspect(reason)}")
 
-        # Fallback
-        {:ack, :ok, state}
+        {exchange_failed_reason, exchange_failed_msg} = classify_exchange_resp_error(reason)
```
noaccOS: can we just return the errors from the failed `with` match instead?

```elixir
  else
    {:error, reason, message} ->
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -360,7 +368,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
   end
 
   defp perform_key_agreement(new_state, init_exchange, payload) do
-    with {:ok, state_after_receive} <-
+    with {:handshake_started, _} = state_after_receive <-
```
noaccOS: I find it was clearer before, as the success state was abstracted by `HandshakeState` and was not up to the caller.

I suggest changing the semantics of the returned value instead: `:ok` means a successful transition while `:error` (or another atom if you prefer) means there was something wrong, even if we did transition to the new state. so when we transition to `failed` we effectively return an `error` instead. this makes it more idiomatic to write `with` blocks matching on `ok`, and knowing when to respond with exchange failed

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -462,12 +497,39 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
+            exchange_failed_reason,
+            exchange_failed_msg
+          )
+
+        # `HandshakeState.transition/2` accepts `{:error, reason}` from *any*
+        # current state and always replies with `{:error, {:failed, reason}}`
+        # (never `{:error, :invalid_transition}`) — trust that documented
+        # contract with a plain match instead of speculatively handling an
+        # outcome it cannot produce.
+        {:error, failed_key_state} =
```
noaccOS: Let's have at most one else clause for each `with` block
Which means having homogeneous assignments and error handling

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -512,10 +576,95 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
 
       {:error, reason} ->
         Logger.error("[keyAgreement/1] State machine transition failed: #{inspect(reason)}")
-        {:error, reason}
+        {exchange_failed_reason, exchange_failed_msg} = classify_exchange_resp_error(reason)
+        {:error, exchange_failed_reason, exchange_failed_msg}
```
noaccOS: I'd say let's have the transition return the reason and message itself so we can avoid all these extra calls. ideally I'd remove `classify_exchange_resp_error` as it creates context from nothing so it doesn't add value

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
```diff
@@ -240,6 +240,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Impl do
       |> TimeBasedActions.execute_time_based_actions(timestamp)
       |> Core.Device.set_device_disconnected(timestamp)
       |> Map.put(:last_seen_message, timestamp)
+      |> Map.put(:encrypted_endpoints_key, :uninitialized)
```
noaccOS: do we have to uninitialize the key when the device disconnects?

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
mizzet1: It was a way to avoid the Elixir code smells, but will go for this cleaner way

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/impl.ex
mizzet1: this was done with the goal of resetting handshake state, including the key, on device disconnection.
However, by looking again at the flow, it makes more sense to perform the key reset on set_device_disconnected rather than in handle_disconnecting, because this way it will also trigger in case of Core.Error.handle_error calls that happen if the exchange fails

mizzet1: I think it also make more sense to maybe put it in :reset ov :uninitilized

eddbbt (approved):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -512,10 +566,69 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
+      [:astarte, :data_updater_plant, :control_handler, :key_agreement_device_failed],
+      %{payload_size: byte_size(payload)},
+      %{realm: state.realm, reason: reason}
+    )
+
+    final_state = %{
+      state
+      | encrypted_endpoints_key: HandshakeState.fail(state.encrypted_endpoints_key, reason),
+        total_received_msgs: state.total_received_msgs + 1,
+        total_received_bytes: state.total_received_bytes + byte_size(payload)
```
noaccOS: should this also add path bytes?

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -462,12 +497,19 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
     else
-      {:error, reason} ->
-        # TODO: Implement failedexchange message
-        Logger.warning("keyAgreement/1 failed, ignoring for now: #{inspect(reason)}")
+      {:error, reason, message} ->
+        Logger.warning("[keyAgreement/1] Processing failed: #{inspect(reason)} - #{message}")
 
-        # Fallback
-        {:ack, :ok, state}
+        :ok = send_exchange_failed(state.realm, state.device_id, 0, reason, message)
```
noaccOS: this is a little misleading as the function always return ok
I suggest having the function return whatever it needs to return instead and avoid strong matches here

noaccOS (changes requested):

noaccOS (approved):

PR #2064, opened 2026-07-02
fix: ensure devices are eventually deleted

mizzet1:
This PR aim to fix a bug where devices are stuck in deletion state, by adding a deletion scheduler procedure that runs every 30 minutes for devices that don't have all three deletion acks checked

### DIFF
--- a/apps/astarte_realm_management/lib/astarte_realm_management/device_removal/scheduler.ex
+++ b/apps/astarte_realm_management/lib/astarte_realm_management/device_removal/scheduler.ex
@@ -27,8 +27,10 @@ defmodule Astarte.RealmManagement.DeviceRemoval.Scheduler do
 
   use GenServer
 
+  alias Astarte.Core.Device
   alias Astarte.RealmManagement.DeviceRemoval.DeviceRemover
   alias Astarte.RealmManagement.DeviceRemoval.Queries
+  alias Astarte.RealmManagement.RPC.DataUpdaterPlant.Client, as: DevicesRPC
 
   require Logger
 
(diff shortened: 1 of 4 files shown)


apps/astarte_realm_management/lib/astarte_realm_management/device_removal/scheduler.ex
```diff
@@ -73,6 +75,26 @@ defmodule Astarte.RealmManagement.DeviceRemoval.Scheduler do
...
+    Enum.each(devices, fn %{device_id: device_id, realm_name: realm_name} ->
+      encoded_device_id = Device.encode_device_id(device_id)
+
+      _ =
+        Logger.debug(
+          "Re-sending start_device_deletion for pending device #{encoded_device_id}",
+          realm: realm_name
+        )
+
+      DevicesRPC.start_device_deletion_rpc(realm_name, device_id)
```
noaccOS: this is a `call`, which means that if the first process crashes (database error on one device?), the following processes are not executed
here we already have a task supervisor for that reason, so I suggest spawning tasks there instead


```suggestion
      Task.Supervisor.start_child(DeviceRemoverSupervisor, fn -> DevicesRPC.start_device_deletion_rpc(realm_name, device_id) end)
```

noaccOS (changes requested):
a feature is not complete without appropriate tests 🤓

apps/astarte_realm_management/lib/astarte_realm_management/device_removal/scheduler.ex
mizzet1: naruhodo

apps/astarte_realm_management/test/astarte_realm_management/device_removal/scheduler_sync_test.exs
```diff
@@ -57,6 +59,44 @@ defmodule Astarte.RealmManagement.DeviceRemoval.SchedulerSyncTest do
...
+        for device_id <- non_ackd_device_ids do
+          assert_receive {:rpc_called, ^device_id}
+        end
+
+        for device_id <- ackd_devices do
+          refute_receive {:rpc_called, ^device_id}
+        end
+
+        (ackd_deletions ++ non_ackd_deletions)
+        |> Enum.each(&Repo.delete!(&1, prefix: Realm.keyspace_name(realm)))
```
frossq, line 95: Could be in a `on_exit` call

frossq (approved):

apps/astarte_realm_management/test/astarte_realm_management/device_removal/scheduler_sync_test.exs
noaccOS: unfortunately `check all`s are a bitch, we need to clean after every run (so that each run starts in a clean state) and not only once per test

noaccOS (approved):

PR #2066, opened 2026-07-03
feat(appengine): Encrypt value before publishing

nedimtokic:
If endpoint has encrypted flag set to true, encrypt value before publishing on broker.
Added transport encryption before publishing data on broker.
Changed DB schema to store %Symmetric{} key with algorithm, and updated functions for insert and read the key.

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
@@ -208,7 +208,6 @@ defmodule Astarte.AppEngine.API.Device do
            ),
          {:ok, value} <- InterfaceValue.cast_value(mapping.value_type, raw_value),
          :ok <- validate_value_type(mapping.value_type, value),
-         wrapped_value = wrap_to_bson_struct(mapping.value_type, value),
          interface_type = interface_descriptor.type,
          reliability = mapping.reliability,
          publish_opts = build_publish_opts(interface_type, reliability),
@@ -217,8 +216,9 @@ defmodule Astarte.AppEngine.API.Device do
            maybe_apply_transport_encryption(
              realm_name,
              mapping,
-             wrapped_value,
-             device_id
+             value,
+             device_id,
+             mapping.value_type
            ),
          :ok <-
            ensure_publish(
@@ -264,18 +264,25 @@ defmodule Astarte.AppEngine.API.Device do
     end
   end
 
-  defp maybe_apply_transport_encryption(realm_name, %Mapping{} = mapping, value, device_id) do
+  defp maybe_apply_transport_encryption(
+         realm_name,
+         %Mapping{} = mapping,
+         value,
+         device_id,
+         expected_types
+       ) do
     encrypted_endpoints = extract_encrypted_endpoints([mapping])
 
     maybe_apply_transport_encryption_with_endpoints(
       realm_name,
       encrypted_endpoints,
       value,
-      device_id
+      device_id,
+      expected_types
     )
   end
 
-  defp maybe_apply_transport_encryption(realm_name, mappings, value, device_id)
+  defp maybe_apply_transport_encryption(realm_name, mappings, value, device_id, expected_types)
        when is_list(mappings) do
     encrypted_endpoints = extract_encrypted_endpoints(mappings)
 
@@ -283,7 +290,8 @@ defmodule Astarte.AppEngine.API.Device do
       realm_name,
       encrypted_endpoints,
       value,
-      device_id
+      device_id,
+      expected_types
     )
   end
 
@@ -291,34 +299,20 @@ defmodule Astarte.AppEngine.API.Device do
          realm_name,
          encrypted_endpoints,
          value,
-         device_id
+         device_id,
+         expected_types
        )
        when is_list(encrypted_endpoints) do
     should_encrypt_transport_payload? = encrypted_endpoints != []
 
     if should_encrypt_transport_payload? do
-      encoded_device_id = Device.encode_device_id(device_id)
-
-      with {:ok, device_status} <- get_device_status!(realm_name, encoded_device_id),
-           :ok <- check_shared_secret(device_status.shared_secret) do
-        encrypt_aes_gcm_binary(value, device_status.shared_secret)
+      with {:ok, shared_secret} <- Queries.retrieve_shared_secret(realm_name, device_id) do
+        bson_value = Cyanide.encode!(%{v: value})
+        EncryptedMessages.encrypt(bson_value, shared_secret.k, shared_secret.alg)
       end
     else
-      value
+      wrap_to_bson_struct(expected_types, value)
     end
-  end
-
-  defp check_shared_secret(nil), do: {:error, :device_not_ready_for_encryption}
-  defp check_shared_secret(_), do: :ok
-
-  defp encrypt_aes_gcm_binary(value, shared_secret) do
-    plaintext = normalize_transport_payload(value)
-    EncryptedMessages.encrypt(plaintext, shared_secret, :aes_256_gcm)
-  end
-
-  defp normalize_transport_payload(value) do
-    # Used for object aggregated interfaces
-    :erlang.term_to_binary(value)
   end
 
   # Helper to calculate TTL and build DB options
@@ -532,7 +526,6 @@ defmodule Astarte.AppEngine.API.Device do
            Map.new(mappings_by_key, fn {k, %Mapping{value_type: t}} -> {k, t} end),
          {:ok, value} <- InterfaceValue.cast_value(expected_types, raw_value),
          :ok <- validate_value(mappings_by_key, value),
-         wrapped_value = wrap_to_bson_struct(expected_types, value),
          reliability = extract_aggregate_reliability(mappings),
          interface_type = interface_descriptor.type,
          publish_opts = build_publish_opts(interface_type, reliability),
@@ -541,8 +534,9 @@ defmodule Astarte.AppEngine.API.Device do
            maybe_apply_transport_encryption(
              realm_name,
              mappings,
-             wrapped_value,
-             device_id
+             value,
+             device_id,
+             expected_types
            ),
          :ok <-
            ensure_publish(
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
@@ -42,6 +42,7 @@ defmodule Astarte.AppEngine.API.Device do
   alias Astarte.DataAccess.Mappings
   alias Astarte.Secrets
   alias Astarte.Secrets.Core
+  alias Astarte.Secrets.EncryptedMessages
   alias Ecto.Changeset
 
   require Logger
@@ -212,13 +213,20 @@ defmodule Astarte.AppEngine.API.Device do
          reliability = mapping.reliability,
          publish_opts = build_publish_opts(interface_type, reliability),
          interface_name = interface_descriptor.name,
+         publish_value =
+           maybe_apply_transport_encryption(
+             realm_name,
+             mapping,
+             wrapped_value,
+             device_id
+           ),
          :ok <-
            ensure_publish(
              realm_name,
              device_id,
              interface_name,
              path,
-             wrapped_value,
+             publish_value,
              publish_opts
            ) do
       opts = build_database_opts(realm_name, mapping)
@@ -256,6 +264,63 @@ defmodule Astarte.AppEngine.API.Device do
     end
   end
 
+  defp maybe_apply_transport_encryption(realm_name, %Mapping{} = mapping, value, device_id) do
+    encrypted_endpoints = extract_encrypted_endpoints([mapping])
+
+    maybe_apply_transport_encryption_with_endpoints(
+      realm_name,
+      encrypted_endpoints,
+      value,
+      device_id
+    )
+  end
+
+  defp maybe_apply_transport_encryption(realm_name, mappings, value, device_id)
+       when is_list(mappings) do
+    encrypted_endpoints = extract_encrypted_endpoints(mappings)
+
+    maybe_apply_transport_encryption_with_endpoints(
+      realm_name,
+      encrypted_endpoints,
+      value,
+      device_id
+    )
+  end
+
+  defp maybe_apply_transport_encryption_with_endpoints(
+         realm_name,
+         encrypted_endpoints,
+         value,
+         device_id
+       )
+       when is_list(encrypted_endpoints) do
+    should_encrypt_transport_payload? = encrypted_endpoints != []
+
+    if should_encrypt_transport_payload? do
+      encoded_device_id = Device.encode_device_id(device_id)
+
+      with {:ok, device_status} <- get_device_status!(realm_name, encoded_device_id),
+           :ok <- check_shared_secret(device_status.shared_secret) do
+        encrypt_aes_gcm_binary(value, device_status.shared_secret)
+      end
+    else
+      value
+    end
+  end
+
+  defp check_shared_secret(nil), do: {:error, :device_not_ready_for_encryption}
+  defp check_shared_secret(_), do: :ok
+
+  defp encrypt_aes_gcm_binary(value, shared_secret) do
+    plaintext = normalize_transport_payload(value)
+    EncryptedMessages.encrypt(plaintext, shared_secret, :aes_256_gcm)
+  end
+
+  defp normalize_transport_payload(value) do
+    # Used for object aggregated interfaces
+    :erlang.term_to_binary(value)
+  end
+
   # Helper to calculate TTL and build DB options
   defp build_database_opts(realm_name, mapping) do
     realm_max_ttl = Queries.fetch_datastream_maximum_storage_retention(realm_name)
@@ -282,14 +347,7 @@ defmodule Astarte.AppEngine.API.Device do
       |> DateTime.to_unix(:microsecond)
       |> Kernel.*(10)
 
-    encrypted_endpoints =
-      case Map.get(mapping, :encrypted) do
-        true ->
-          [Map.get(mapping, :endpoint)]
-
-        _ ->
-          []
-      end
+    encrypted_endpoints = extract_encrypted_endpoints([mapping])
 
     case maybe_encrypt_value(
            desc.aggregation,
@@ -479,13 +537,20 @@ defmodule Astarte.AppEngine.API.Device do
          interface_type = interface_descriptor.type,
          publish_opts = build_publish_opts(interface_type, reliability),
          interface_name = interface_descriptor.name,
+         publish_value =
+           maybe_apply_transport_encryption(
+             realm_name,
+             mappings,
+             wrapped_value,
+             device_id
+           ),
          :ok <-
            ensure_publish(
              realm_name,
              device_id,
              interface_name,
              path,
-             wrapped_value,
+             publish_value,
              publish_opts
            ) do
       handle_object_value_storage(%{
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device_status.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/device_status.ex
@@ -42,7 +42,6 @@ defmodule Astarte.AppEngine.API.Device.DeviceStatus do
     field :previous_interfaces, {:array, :map}
     field :groups, {:array, :string}
     field :deletion_in_progress, :boolean, default: false
-    field :shared_secret, :binary
   end
 
   @doc false
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
@@ -243,7 +243,8 @@ defmodule Astarte.AppEngine.API.Device.Queries do
     :exchanged_bytes_by_interface,
     :groups,
     :old_introspection,
-    :inhibit_credentials_request
+    :inhibit_credentials_request,
+    :shared_secret
   ]
 
   defp truncate_datetime(nil), do: nil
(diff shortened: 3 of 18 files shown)


noaccOS (approved):
Code is good, would it be possible to add tests at this stage?

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -212,6 +212,15 @@ defmodule Astarte.AppEngine.API.Device do
...
+         {:ok, {value, encrypted_dek}} <-
+           maybe_encrypt_value(
+             interface_descriptor.aggregation,
+             path,
+             value,
+             encrypted_endpoints,
+             realm_name
+           ),
          :ok <-
            ensure_publish(
```
frossq, line 224: It seems this `ensure_publish` is still sending plaintext data (as in `wrap_to_bson_struct` output). Better to revisit this logic when we agree with embedded team on how to encode data to be send encrypted (possibly through CBOR). The rest is convincing 🔥

nedimtokic: Actually, I just realized that we have special encryption for transport. I'll have to fix this. I agree.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +264,96 @@ defmodule Astarte.AppEngine.API.Device do
...
+  end
+
+  defp normalize_encryption_path(path) do
+    "/" <> String.trim(path, "/")
+  end
+
+  defp with_trailing_slash("/"), do: "/"
+  defp with_trailing_slash(path), do: path <> "/"
+
+  defp encrypt_aes_gcm_binary(value, shared_secret) do
```
frossq: I fear we need to go through CBOR.Encrypt0 here

noaccOS: `COSE.Messages.Encrypt0`*

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +265,89 @@ defmodule Astarte.AppEngine.API.Device do
...
+      realm_name,
+      encrypted_endpoints,
+      path,
+      value,
+      device_id
+    )
+  end
+
+  defp maybe_apply_transport_encryption(realm_name, mappings, path, value, device_id)
+       when is_list(mappings) do
```
noaccOS, line 286: isn't mapping always a single mapping? it comes from `Mapping.from_db_result`

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +265,89 @@ defmodule Astarte.AppEngine.API.Device do
...
+       when is_list(encrypted_endpoints) do
+    normalized_path = normalize_encryption_path(path)
+
+    should_encrypt_transport_payload? =
+      Enum.any?(encrypted_endpoints, fn endpoint ->
+        normalized_endpoint = normalize_encryption_path(endpoint)
+
+        normalized_endpoint == normalized_path or
+          String.starts_with?(normalized_endpoint, with_trailing_slash(normalized_path))
+      end)
```
noaccOS: I'd double check the intention here but I think we should encrypt
- objects when one of the endpoints is encrypted (even if not in the current object)
- individual when its endpoint is encrypted

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +265,89 @@ defmodule Astarte.AppEngine.API.Device do
...
+
+    if should_encrypt_transport_payload? do
+      encoded_device_id = Base.url_encode64(device_id, padding: false)
```
noaccOS:
```suggestion
      encoded_device_id = Astarte.Core.Device.encode_device_id(device_id)
```

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +265,89 @@ defmodule Astarte.AppEngine.API.Device do
...
+
+      with {:ok, %DeviceStatus{shared_secret: shared_secret}} <-
+             get_device_status!(realm_name, encoded_device_id),
+           false <- is_nil(shared_secret),
+           {:ok, encrypted_binary} <- encrypt_aes_gcm_binary(value, shared_secret) do
+        {:ok, encrypted_binary}
+      else
+        true -> {:error, :device_not_ready_for_encryption}
+        {:error, reason} -> {:error, reason}
+        _ -> {:error, :encryption_failed}
```
noaccOS: having more than 1 `else` clause in `with` blocks is a red flag, because it's a lot more effort when trying to understand where each value could possibly be generated, which makes both maintenance and debugging harder than they need to be

prefer instead wrapping each match in its own private function so that you can have a single `else` match (or avoid it entirely)

```elixir
with {:ok, device_status} <- get_device_status!(realm_name, encoded_device_id),
     :ok <- check_shared_secret(device_status.shared_secret) do
  encrypt_aes_gcm_binary(value, device_status.shared_secret) 
end

# ...

defp check_shared_secret(nil), do: {:error, :device_not_ready_for_encryption}
defp check_shared_secret(_), do: :ok
```

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +265,89 @@ defmodule Astarte.AppEngine.API.Device do
...
+  defp with_trailing_slash(path), do: path <> "/"
+
+  # TODO: Astarte.Secrets.encrypt_with_dek can replace this function
+  defp encrypt_aes_gcm_binary(value, shared_secret) do
+    plaintext = normalize_transport_payload(value)
+
+    encrypted_binary = EncryptedMessages.encrypt(plaintext, shared_secret, :aes_256_gcm)
+    {:ok, encrypted_binary}
+  rescue
+    _ -> {:error, :encryption_failed}
```
noaccOS: if we get here the values are valid (we passed `validate_value_type` and whatnot)
so if we raise it's a programming error
no `:ok | :error` wrap is going to save us there, it's best to just let it crash so that we can immediately assess what happened

noaccOS (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +265,89 @@ defmodule Astarte.AppEngine.API.Device do
...
+  end
+
+  defp normalize_encryption_path(path) do
+    "/" <> String.trim(path, "/")
+  end
+
+  defp with_trailing_slash("/"), do: "/"
+  defp with_trailing_slash(path), do: path <> "/"
+
+  # TODO: Astarte.Secrets.encrypt_with_dek can replace this function
```
frossq: Unfortunately not

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: For aggregate objects it is list of %Mapping{}

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Changed.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Removed rescue.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Leftover. Removed.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Changed.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: If there is any encrypted endpoint, encryption is performed now.

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -45,74 +47,79 @@ defmodule Astarte.AppEngine.API.Device.DeviceReadingV2Test do
...
 
-    property "returns the value (from root) ", context do
-      %{
-        realm_name: realm_name,
-        interfaces: interfaces,
-        device: device
-      } = context
-
-      valid_interfaces_for_update = interfaces |> Enum.filter(&(&1.ownership == :server))
+    # property "returns the value (from root) ", context do
```
noaccOS: why were these tests commented out?

also tip: if you need to disable tests you can use `@tag skip: "broken because xyz"`

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +264,64 @@ defmodule Astarte.AppEngine.API.Device do
...
+         value,
+         device_id
+       )
+       when is_list(encrypted_endpoints) do
+    should_encrypt_transport_payload? = encrypted_endpoints != []
+
+    if should_encrypt_transport_payload? do
+      encoded_device_id = Device.encode_device_id(device_id)
+
+      with {:ok, device_status} <- get_device_status!(realm_name, encoded_device_id),
```
frossq: A log around here to notify failed preparation of message could help future debugging

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +264,64 @@ defmodule Astarte.AppEngine.API.Device do
...
+
+  defp encrypt_aes_gcm_binary(value, shared_secret) do
+    plaintext = normalize_transport_payload(value)
+    encrypted_binary = EncryptedMessages.encrypt(plaintext, shared_secret, :aes_256_gcm)
+    {:ok, encrypted_binary}
+  end
+
+  defp normalize_transport_payload(value) do
+    # Used for object aggregated interfaces
+    :erlang.term_to_binary(value)
```
frossq: Would this elixir-specific binary construction break the CBOR decryption at device side?

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
frossq: Dead tests? 😿

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: `maybe_apply_transport_encryption` is in with block in `update_individual_interface_values`, and if function returns error it will be logged "Error while writing to interface."

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
nedimtokic: Leftover. Thanks.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Not sure about this.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +264,64 @@ defmodule Astarte.AppEngine.API.Device do
...
+      {:ok, value}
+    end
+  end
+
+  defp check_shared_secret(nil), do: {:error, :device_not_ready_for_encryption}
+  defp check_shared_secret(_), do: :ok
+
+  defp encrypt_aes_gcm_binary(value, shared_secret) do
+    plaintext = normalize_transport_payload(value)
+    encrypted_binary = EncryptedMessages.encrypt(plaintext, shared_secret, :aes_256_gcm)
```
noaccOS: `EncryptedMessages.encrypt/3` currently expects a COSE key, while the shared secret currently is the raw binary.
We could change the `encrypt` function to accept the raw binary and wrap it in a `Symmetric` key instead, or store the shared secret as the symmetric key itself (in which case the current appengine code would be good but we'd still need to change dup/data access)

I don't have a strong opinion on the approach here. Maybe I slightly prefer the second approach as we could remove the `key_type` from the `encrypt` signature (and use `session_key.alg` instead)

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +264,64 @@ defmodule Astarte.AppEngine.API.Device do
...
+    {:ok, encrypted_binary}
```
noaccOS: if we never return error let's just return the value itself


```suggestion
    EncryptedMessages.encrypt(plaintext, shared_secret, :aes_256_gcm)
```

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -267,7 +275,15 @@ defmodule Astarte.AppEngine.API.Device.DeviceReadingV2Test do
       assert %{payload: payload, topic_tokens: topic_tokens, qos: qos} = args
       assert topic_tokens == expected_token
       assert qos == expected_qos
-      assert {:ok, %{"v" => ^expected_published_value}} = Cyanide.decode(payload)
+
+      assert {:ok, %{"v" => published_value}} = Cyanide.decode(payload)
+
+      if encrypted_path? do
+        assert is_binary(published_value)
+        refute published_value == expected_published_value
```
noaccOS: can't we do something like


```suggestion
        assert decrypt_value(published_value, device.shared_secret) == expected_published_value
```

instead?

noaccOS (changes requested):

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
nedimtokic: Tests are returned, they are passing.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Changed.

apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
nedimtokic: Done by passing key into `update_interface` fn.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/state.ex
```diff
@@ -114,6 +114,6 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.State do
     field :last_datastream_maximum_retention_refresh, non_neg_integer(), default: 0
     field :capabilities, Capabilities.t(), default: %Capabilities{}
     field :encrypted_endpoints_key, HandshakeState.t(), default: :uninitialized
-    field :shared_secret, binary()
+    field :shared_secret, struct() | nil
```
noaccOS: `Symmetric.t() | nil`

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device_status.ex
```diff
@@ -42,7 +42,7 @@ defmodule Astarte.AppEngine.API.Device.DeviceStatus do
     field :previous_interfaces, {:array, :map}
     field :groups, {:array, :string}
     field :deletion_in_progress, :boolean, default: false
-    field :shared_secret, :binary
+    field :shared_secret, struct() | nil
```
noaccOS: actually this line should be removed, we don't really want to return the secret when we query for the device

libs/astarte_data_access/lib/astarte_data_access/database/migrations/realm/0025_add_shared_secret_to_device.ex
```diff
@@ -23,7 +23,7 @@ defmodule Astarte.DataAccess.Database.Migrations.Realm.AddSharedSecretToDevice d
 
   def change do
     alter table(:devices) do
-      add :shared_secret, :binary
+      add :shared_secret, struct() | nil
```
noaccOS:
```suggestion
      add :shared_secret, :session_key
```

but we really should add a new migration which changes the type of the field (or deletes the field and creates a new one)

libs/astarte_data_access/lib/astarte_data_access/devices/device.ex
```diff
@@ -68,6 +68,6 @@ defmodule Astarte.DataAccess.Devices.Device do
     field :protocol_revision, :integer
     field :total_received_bytes, :integer
     field :total_received_msgs, :integer
-    field :shared_secret, :binary
+    field :shared_secret, struct() | nil
```
noaccOS: this should use embeddedtype instead, look at to2session

noaccOS (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Changed DB schema to store %Symmetric{} key with algorithm, and updated functions for insert and read the key.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
noaccOS: I think that (for object aggregates at least) we agreed that the value -> binary conversion would be bson
I'd double check with joshua though

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -923,4 +923,30 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
+      from d in DatabaseDevice,
+        prefix: ^keyspace_name,
+        select: d.shared_secret
+
+    opts = [consistency: Consistency.device_info(:read), error: :device_not_found]
+
+    Repo.fetch(query, device_id, opts)
+  end
+
+  # TODO This is copied from DUP
```
noaccOS, line 944: maybe also a `@doc false` if we're not using it in the code?

noaccOS (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
nedimtokic: It is only used for tests.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Changed encoding to BSON.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -923,4 +923,31 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
+    keyspace_name = Realm.keyspace_name(realm)
+
+    query =
+      from d in DatabaseDevice,
+        prefix: ^keyspace_name,
+        select: d.shared_secret
+
+    opts = [consistency: Consistency.device_info(:read), error: :device_not_found]
+
+    Repo.fetch(query, device_id, opts)
```
noaccOS: maybe it would be more ergonomic to handle nil shared secret here instead?


```suggestion
    case Repo.fetch(query, device_id, opts) do
      {:ok, nil} -> {:error, :device_not_ready_for_encryption}
      {:ok, shared_secret} -> {:ok, shared_secret}
      error -> error
    end
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/state.ex
```diff
@@ -114,6 +114,6 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.State do
     field :last_datastream_maximum_retention_refresh, non_neg_integer(), default: 0
     field :capabilities, Capabilities.t(), default: %Capabilities{}
     field :encrypted_endpoints_key, HandshakeState.t(), default: :uninitialized
-    field :shared_secret, binary()
+    field :shared_secret, Symmetric.t()
```
noaccOS, line 118: dialyzer is failing, missing alias?

noaccOS (approved):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -256,6 +264,61 @@ defmodule Astarte.AppEngine.API.Device do
...
+         value,
+         device_id,
+         expected_types
+       )
+       when is_list(encrypted_endpoints) do
+    should_encrypt_transport_payload? = encrypted_endpoints != []
+
+    if should_encrypt_transport_payload? do
+      with {:ok, shared_secret} <- Queries.retrieve_shared_secret(realm_name, device_id),
+           :ok <- check_shared_secret(shared_secret) do
```
noaccOS: this is now unnecessary as it is already returned by the `retrieve_shared_secret` call

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/state.ex
noaccOS: still relevant

noaccOS (changes requested):

noaccOS (approved):
:sushi:

PR #2068, opened 2026-07-06
refactor(dup): make HashOk return the associated sequence number

mizzet1:
This PR aim to refactor the HashOk message structure of the Key Agreement protocol so the returned value is the associated SecretHash sequence number

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
@@ -401,7 +401,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
   end
 
   defp process_secret_hash(
-         %{encrypted_endpoints_key: {:established, %{shared_secret: shared_secret, alg: alg}}} =
+         %{encrypted_endpoints_key: {:established, %{shared_secret: shared_secret}}} =
            state,
          seq_num,
          secret_hash_msg,
@@ -409,7 +409,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
          _timestamp
        ) do
     state
-    |> confirm_secret_hash(secret_hash_msg, shared_secret, alg)
+    |> confirm_secret_hash(secret_hash_msg, shared_secret)
     |> case do
       {:ok, new_key_state} ->
         final_state = %{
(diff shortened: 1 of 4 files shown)


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -412,27 +420,54 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
 
   defp process_secret_hash(state, _seq_num, _secret_hash_msg, _payload, _timestamp) do
     Logger.warning("[keyAgreement/2] No shared secret established.")
-    # TODO: implement ExchangeFailed message
+
+    :ok =
+      send_exchange_failed(
+        state.realm,
+        state.device_id,
+        0,
```
frossq: Probably we'd want to send back the right seqnum also here

mizzet1: thank you good catch

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -563,6 +712,64 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
...
 
+  defp send_exchange_failed(realm, device_id, seq_num, reason, error_msg) do
+    topic = "#{realm}/#{Device.encode_device_id(device_id)}/control/keyAgreement/4"
+
+    {:ok, exchange_failed} = ExchangeFailed.new(seq_num, reason, error_msg)
+    payload = ExchangeFailed.cbor_encode(exchange_failed)
+
+    publish_start = System.monotonic_time()
+
+    case VMQPlugin.publish(topic, payload, 2) do
```
frossq, line 693: In the doc it is written "All Key Agreement messages are published at QoS **1**"   :o

mizzet1: We confirmed with embedded team that we want exactly once delivery for these messages, so qos 2 is correct here
Will update both the draft doc and put that info in the rfc too

frossq (approved):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/key_agreement/hash_ok.ex
```diff
@@ -23,28 +23,41 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.KeyAgreement.HashOk do
...
+    field :seq_num, non_neg_integer()
+  end
+
+  @doc """
+  Builds a new `%HashOk{}` given the sequence number of the associated
+  `SecretHash` message being acknowledged.
+  """
+  @spec new(non_neg_integer()) :: t()
+  def new(seq_num) when is_integer(seq_num) and seq_num >= 0 do
+    %HashOk{seq_num: seq_num}
```
noaccOS: in general it's best practice to avoid `new` functions when all they do is populate struct fields
it's preferable for the caller to use `%HashOk{seq_num: seq_num}` instead as it's clearer

noaccOS (changes requested):

noaccOS (approved):

PR #2094, opened 2026-07-16
chore(dup): configure log level through env var at runtime

mizzet1:
This PR aim to add configuration of Data Updater Plant log level through a newly defined env var that can be set at runtime to dinamically change the log level of DUP


apps/astarte_data_updater_plant/config/runtime.exs
```diff
@@ -0,0 +1,10 @@
+import Config
+
+if config_env() == :prod do
+  log_level =
+    System.get_env("DATA_UPDATER_PLANT_LOG_LEVEL", "info")
```
noaccOS: 1. we usually use the astarte prefix for dup variables `ASTARTE_DATA_UPDATER_PLANT_*`
2. IMO to avoid handling 6 separate vars, I'd say a single `ASTARTE_LOG_LEVEL` is easier to handle. what do you think?

apps/astarte_data_updater_plant/config/runtime.exs
```diff
@@ -0,0 +1,10 @@
...
+    |> String.downcase()
+    |> String.to_existing_atom()
```
noaccOS: this technically allows setting almost any string as a log level. I'd suggest restricting the user to only https://logger.hexdocs.pm/Logger.html#level/0 allowed levels

```elixir
allowed_levels = ["emergency", "alert", "critical", "error", "warning", "warn", "notice", "info", "debug", "all", "none"]
level = System.get_env("DATA_UPDATER_PLANT_LOG_LEVEL", "info")
if level not in allowed_levels, do: raise ~s[Invalid value for ASTARTE_LOG_LEVEL: "#{level}"]
level = String.to_existing_atom(level)
```

apps/astarte_data_updater_plant/config/runtime.exs
```diff
@@ -0,0 +1,10 @@
+import Config
+
+if config_env() == :prod do
+  log_level =
+    System.get_env("DATA_UPDATER_PLANT_LOG_LEVEL", "info")
+    |> String.downcase()
```
noaccOS: I don't think we allow case insensitive settings for other "enums", I'd just not add this

noaccOS (changes requested):

noaccOS (approved):

apps/astarte_data_updater_plant/config/runtime.exs
```diff
@@ -0,0 +1,23 @@
...
+    "warn",
+    "notice",
+    "info",
+    "debug",
+    "all",
+    "none"
+  ]
+
+  level = System.get_env("DATA_UPDATER_PLANT_LOG_LEVEL", "info")
+  if level not in allowed_levels, do: raise(~s[Invalid value for ASTARTE_LOG_LEVEL: "#{level}"])
```
noaccOS: wait the name wasn't updated consistently

noaccOS (changes requested):

apps/astarte_data_updater_plant/config/config.exs
```diff
@@ -22,6 +22,8 @@ config :logger, :console,
     :tag
   ]
 
+config :logger, level: :info
```
noaccOS: this is literally overriden in each environment, I suggest defaulting to `debug` here and removing the change to `dev`, `test` and `ci`

noaccOS (changes requested):

noaccOS (approved):

PR #2097, opened 2026-07-20
fix(pairing): Implement Service Info chunks

nedimtokic:
This PR fixes the FDO `service_info_end` flow in Pairing API and extends controller coverage for the device ServiceInfo exchange.

The main change is that `service_info_end` now handles the three protocol cases explicitly:

- device sends more chunks, so the payload is accumulated and the owner replies with an empty OwnerServiceInfo
- device yields with an empty ServiceInfo map, so the owner continues with the next chunk
- device sends the final chunk, so the device is registered and the owner ServiceInfo response is built and returned


apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
```diff
@@ -432,24 +432,56 @@ defmodule Astarte.PairingWeb.FDOOnboardingController do
...
+       when map_size(service_info) == 0 do
+    ServiceInfo.build_owner_service_info(realm_name, session, device_service_info)
+  end
+
+  defp build_owner_service_info_response(
+         realm_name,
+         session,
+         %DeviceServiceInfo{is_more_service_info: false} = device_service_info
+       ) do
+    device_id = generate_device_id(device_service_info)
```
noaccOS: mhhhh this is only considering the data present in the last chunk
I think we should instead handle the owner service info so that the last chunk is merged with the rest of the service info before proceeding
may need a little refactor on the `astarte_fdo` side as well

noaccOS (changes requested):

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
nedimtokic: Called function to merge and store last device service info. Device is generate now using complete and decoded data (since device service info is CBOR encoded).

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
```diff
@@ -432,26 +432,72 @@ defmodule Astarte.PairingWeb.FDOOnboardingController do
...
+           Session.add_device_service_info(session, realm_name, service_info),
+         decoded_service_info <-
+           Session.decode_device_service_info(session.device_service_info),
+         decoded_device_service_info <-
+           %DeviceServiceInfo{
+             is_more_service_info: false,
+             service_info: decoded_service_info
+           },
+         device_id <- generate_device_id(decoded_device_service_info),
+         encoded_device_id <- Device.encode_device_id(device_id),
```
noaccOS: let's use `=` when matches can't fail


```suggestion
         decoded_service_info =
           Session.decode_device_service_info(session.device_service_info),
         decoded_device_service_info =
           %DeviceServiceInfo{
             is_more_service_info: false,
             service_info: decoded_service_info
           },
         device_id = generate_device_id(decoded_device_service_info),
         encoded_device_id = Device.encode_device_id(device_id),
```

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
```diff
@@ -432,26 +432,72 @@ defmodule Astarte.PairingWeb.FDOOnboardingController do
...
+        encoded_device_id,
+        credentials_secret
+      )
+    end
+  end
+
   defp generate_device_id(%DeviceServiceInfo{service_info: %{{"devmod", "sn"} => %{value: sn}}}),
     do: UUID.uuid5(:oid, sn, :raw)
 
+  defp generate_device_id(%DeviceServiceInfo{
```
noaccOS: why can we parse the service info either with atom or string keys?

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
```diff
@@ -432,26 +432,72 @@ defmodule Astarte.PairingWeb.FDOOnboardingController do
...
+       ) do
+    with {:ok, session} <-
+           Session.add_device_service_info(session, realm_name, service_info),
+         decoded_service_info <-
+           Session.decode_device_service_info(session.device_service_info),
+         decoded_device_service_info <-
+           %DeviceServiceInfo{
+             is_more_service_info: false,
+             service_info: decoded_service_info
+           },
```
noaccOS: I suggest changing the `generate_device_id` private function if we changed the scope in which we use it, instead of wrapping the data ourselves

noaccOS (changes requested):

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
nedimtokic: Changed. Also, moved `add_device_service_info` fn before decoding so we decode entire message.

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
nedimtokic: Changed to use only string keys.

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
nedimtokic: Changed, to match only service info serial number.

apps/astarte_pairing/lib/astarte_pairing_web/controllers/fdo_onboarding_controller.ex
```diff
@@ -432,26 +432,65 @@ defmodule Astarte.PairingWeb.FDOOnboardingController do
...
+    ServiceInfo.build_owner_service_info(realm_name, session, device_service_info)
+  end
+
+  defp build_owner_service_info_response(
+         realm_name,
+         session,
+         %DeviceServiceInfo{is_more_service_info: false, service_info: service_info}
+       ) do
+    {:ok, session} =
+      Session.add_device_service_info(session, realm_name, service_info)
```
noaccOS: I would advise against strong matches when the data depends on external sources (such as data from the device)
The single `with` was better IMO

libs/astarte_fdo/lib/service_info.ex
```diff
@@ -59,31 +57,23 @@ defmodule Astarte.FDO.ServiceInfo do
           service_info: service_info
         }
       )
-      when is_empty(service_info) do
```
noaccOS, line 62: are we still using `is_empty`?

noaccOS (approved):

eddbbt (approved):

libs/astarte_fdo/lib/service_info.ex
nedimtokic: Nope, we have  `map_size(service_info) == 0 ` in controller.

PR #2100, opened 2026-07-21
test(pairing): fix flakiness

mizzet1:
This PR aim to fix a flaky test in Astarte Pairing by adding sequential namespace creation before async tests start to avoid race condition


noaccOS (approved):
🤖

PR #2104, opened 2026-07-24
feat(dup): Encrypt device shared secret

nedimtokic:
When device and Astarte exchange shared secret, it is encrypted using realm KEK and then stored


libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -223,6 +223,72 @@ defmodule Astarte.Secrets do
...
+
+        :error
+    end
+  end
+
+  @doc """
+  Decrypts the provided vault ciphertext using OpenBao Transit Engine.
+  """
+  @spec decrypt_with_key(String.t(), String.t(), list()) :: {:ok, binary()} | :error
+  def decrypt_with_key(key_name, ciphertext, options \\ []) do
```
noaccOS: I think it would be better to expose something like `encrypt_with_kek` instead, which accepts the realm name and uses that to decrypt the cipher, instead of pushing that responsibility on to the user

The same applies to the namespace, I would like that the caller is not required to use or know of namespaces

The takeaway is that this is supposed to be an abstraction over our functionality not over openbao's apis

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -223,6 +223,72 @@ defmodule Astarte.Secrets do
     Core.decrypt_with_dek(blob, dek)
   end
 
+  @doc """
+  Encrypts the provided plaintext using OpenBao Transit Engine with the given key.
+  Returns the vault ciphertext.
+  """
+  @spec encrypt_with_key(String.t(), binary(), list()) :: {:ok, String.t()} | :error
+  def encrypt_with_key(key_name, plaintext, options \\ []) do
```
noaccOS: Same as decrypt

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
```diff
@@ -935,9 +936,23 @@ defmodule Astarte.AppEngine.API.Device.Queries do
...
+      error ->
+        error
+    end
+  end
+
+  defp decrypt_shared_secret(realm, %Symmetric{k: key_material} = shared_secret)
+       when is_binary(key_material) do
+    with {:ok, kek} <- Secrets.fetch_realm_kek(realm),
+         {:ok, decrypted_key} <-
+           Secrets.decrypt_with_key(kek.name, key_material, namespace: kek.namespace) do
```
noaccOS: Nit: right now it's ok but given we have the goal of putting these functions on data access eventually :tm: it would be great to decouple database access from openbao's access

So my suggestion (now or during a refactor later) would be to have another fetch which sits on top of database access and secrets access and puts the two together

noaccOS (changes requested):

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
frossq: This function seems very similar to the `decrypt` function below (can be merged?)

frossq (approved):

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
nedimtokic: Changed function to accept realm name and fetch KEK before decrypting.

nedimtokic: We have different request body and now paremeters.

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
nedimtokic: Changed.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
nedimtokic: If we move `fetch_realm_kek` and `decrypt_with_key` into data_access library, it could contain changes that not related to shared secret encryption.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -574,12 +575,21 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
   end
 
   defp persist_shared_secret(realm, device_id, %Symmetric{} = symmetric_key) do
-    case Queries.save_shared_secret(realm, device_id, symmetric_key) do
-      :ok ->
-        :ok
+    with {:ok, encrypted_secret} <- encrypt_shared_secret(realm, symmetric_key) do
```
noaccOS, line 578: `encrypt_shared_secret` does not return a 3 element tuple, so this function returns `:ok | {:error, reason} | {:error, reason, message}`

I think the caller expects either 2 or 3 element tuples

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -223,6 +223,79 @@ defmodule Astarte.Secrets do
     Core.decrypt_with_dek(blob, dek)
   end
 
+  @doc """
+  Encrypts the provided plaintext using OpenBao Transit Engine with the given key.
+  Returns the vault ciphertext.
+  """
+  @spec encrypt_with_kek(String.t(), binary(), list()) :: {:ok, String.t()} | :error
+  def encrypt_with_kek(realm_name, plaintext, options \\ []) do
+    {:ok, key} = fetch_realm_kek(realm_name)
```
noaccOS: Let's not use hard matches here, this is a fallible operation.

instead, prefer something like

```elixir
with {:ok, key} <- fetch_realm_kek(realm_name) do
  client_opts = options |> Keyword.take([:token]) |> Keyword.put(:namespace, key.namespace)
  encrypt(key.name, plaintext, opts)
end
```

and something similar for decryption. for decryption you already have the `decrypt` function, for encrypt you could add a private or `Core` function instead.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/queries.ex
noaccOS: I'd rather keep data_access a database only library and secrets a vault only library, and instead add another library on top which orchestrates their interaction

noaccOS (changes requested):

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
nedimtokic: Existing `decrypt` function has different request body.

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -223,19 +223,56 @@ defmodule Astarte.Secrets do
...
-  def decrypt(key_name, ciphertext, options \\ []) do
+  def decrypt(key_name, ciphertext, options \\ [])
+
+  def decrypt(key_name, "vault:v1:" <> _ = ciphertext, options) do
+    do_decrypt(key_name, ciphertext, options)
+  end
+
+  def decrypt(key_name, ciphertext, options) do
+    ciphertext = "vault:v1:" <> Base.encode64(ciphertext)
+    do_decrypt(key_name, ciphertext, options)
```
noaccOS: I see now what you meant. these two are indeed separate functions, one expects the body to be already serialized while the other accepts a "ciphertext" and serializes it itself.

My suggestion here would be to not have a single function do two separate things. Instead, we could have a `decrypt` function in core which is responsible for performing the call to openbao, the `decrypt` function here serializes the body while the `decrypt_with_kek` calls the core function instead of this one

nedimtokic: Moved decrypt` request` to the Core module and called it in `decrypt_with_kek` and `decrypt` functions in `AstarteSecrets` module.

noaccOS (approved):
Code looks good just needs green ci

PR #2103, opened 2026-07-22
chore(dup): decrypt payloads incoming on encrypted endpoints

frossq:
Decrypt CBOR-encrypted payloads sent from devices on encrypted endpoints, using the shared secret previously derived and found in DUP state.

Ensure that the key status gets set to 'established' when the shared key is correctly retrieved from db.

Added tests for payloads decryption and related errors.
Test setup now also imports shared secret into DUP state.

#### Additional documentation e.g. usage docs, diagrams, reviewer notes, etc.:

---

<details>
<summary><i>Thanks for sending a pull request! If this is your first time, here are some tips for you:</i></summary>

1. You can take a look at our [developer guide] for an introduction on Astarte development!
2. Make sure to read [CONTRIBUTING.md] and [CODE_OF_CONDUCT.md]
3. If the PR is unfinished or you're actively working on it, mark it as draft

When fixing existing issues, use [github's syntax to link your pull request] to it

> `fixes #<issue number>`

We also have a syntax to signal dependencies to other open pull requests

> `depends on #<pr number>`
> `depends on https://github.com/...`

In case of stacked PRs, you may add the PR number in the last commit's title instead:

> ```mermaid
> gitGraph
>     commit id: "Current master"
>     branch feat1
>     checkout feat1
>     commit id: "feat: add something"
>     commit id: "feat: add something else (#100)"
>     branch feat2
>     checkout feat2
>     commit id: "refactor: do something"
>     commit id: "fix: solve issue"
>     commit id: "feat: add a feature (#101)"
>     branch feat3
>     checkout feat3
>     commit id: "feat: feat without pr number"
> ```

</details>

[github's syntax to link your pull request]: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#about-linked-issues-and-pull-requests
[developer guide]: https://docs.astarte-platform.org/astarte/snapshot/001-dev_guide.html
[CONTRIBUTING.md]: ../CONTRIBUTING.md
[CODE_OF_CONDUCT.md]: ../CODE_OF_CONDUCT.md

PR #2112, opened 2026-07-31
chore(DUP): Refactor encrypted endpoints

nedimtokic:
- Remove unrelated comment in AppEngine related to encrypted endpoints
- Update interface in encrypted endpoints tests in DUP to support parameterized paths
- Update DUP tests to replace generic interface path with some value
- Ensure DUP handle paths with parameters for individual interfaces

### DIFF
--- a/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
+++ b/apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
@@ -134,12 +134,12 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
     encrypted_endpoints =
       state.mappings
       |> Map.values()
-      |> Enum.filter(& &1.encrypted)
+      |> Enum.filter(&(&1.interface_id == context.interface_id and &1.encrypted))
       |> Enum.map(& &1.endpoint)
 
     case maybe_encrypt_value(
            interface_descriptor.aggregation,
-           path,
+           mapping.endpoint,
            value,
            encrypted_endpoints,
            realm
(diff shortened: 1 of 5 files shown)


apps/astarte_data_updater_plant/test/support/data_query_helper.ex
```diff
@@ -35,19 +35,34 @@ defmodule Astarte.DataUpdaterPlant.DataQueryHelper do
...
           binary(),
+          binary(),
           String.t(),
           :individual_datastream | :property
         ) :: list(map())
-  def query_endpoint_data(interface, endpoint_path, device_id, keyspace, :property) do
+  def query_endpoint_data(
+        interface,
+        endpoint_path,
+        _resolved_path,
```
noaccOS: let's call these `endpoint` and `path` as we do throughout the codebase

apps/astarte_data_updater_plant/test/support/data_query_helper.ex
```diff
@@ -35,19 +35,34 @@ defmodule Astarte.DataUpdaterPlant.DataQueryHelper do
...
+        device_id,
+        keyspace,
+        :property
+      ) do
     interface_id = CQLUtils.interface_id(interface.name, interface.major_version)
     endpoint_id = CQLUtils.endpoint_id(interface.name, interface.major_version, endpoint_path)
```
noaccOS: I think this should also use the endpoint and not the path:

`def endpoint_id(interface_name, interface_major, endpoint)`

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
noaccOS, line 134: this pr should probably be based on top of #2103 which has some fixes

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
nedimtokic: Filtered endpoints by interface ID now. Thanks!

noaccOS (approved):

Issue #2118, opened 2026-08-06
Wrong error for interface not installed in Astarte


joshuachp:
When a device has an interface not installed in Astarte the error returned will be `resend_interface_properties_failed` instead of `interface_loading_failed`.

Here the relative Astarte console log:

<img width="1274" height="555" alt="Image" src="https://github.com/user-attachments/assets/000c87e1-b2e1-4900-a600-7bd59a4662c1" />

And log lines:

```
level=warning ts=2026-08-06T07:36:53.543+00:00 msg="Failed io.edgehog.devicemanager.apps.AvailableDeviceRequests interface loading." realm=cleaos device_id=2hWlA3sRRbOiEnzUZKJ4tA module=Astarte.DataUpdaterPlant.DataUpdater.Core.Device function=resend_all_properties/1
```

noaccOS:
fixed by #2119. included in 1.3.3 and the latest 1.4 snapshots

PR #2119, opened 2026-08-06
fix(dup): stop remapping interface loading failed error

mizzet1:
When resending properties to a device whose introspection includes an interface that isn't installed in Astarte, the device error trigger fired resend_interface_properties_failed instead of the actual cause, interface_loading_failed

This PR fixes this [issue](http://github.com/astarte-platform/astarte/issues/2118) by making Astarte correctly report interface_loading_failed.


apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/control_handler.ex
```diff
@@ -343,15 +343,15 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.ControlHandler do
 
         {:ok, new_state}
 
-      {:error, :sending_properties_to_interface_failed} ->
-        # Track interface send failure
+      {:error, :interface_loading_failed} ->
+        # Track interface loading failure
         :telemetry.execute(
           [:astarte, :data_updater_plant, :control_handler, :properties_resend],
```
noaccOS, line 349: this is inconsistent with the actual problem reported. instead, we should allow both errors to be emitted by `resend_all_properties`, and differenciate between
- interface not found
- error while sending data to the device

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/device.ex
```diff
@@ -424,9 +424,12 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.Device do
...
         {:cont, {:ok, new_state}}
       else
-        {:error, :interface_loading_failed} ->
+        {:error, :interface_loading_failed} = error ->
           Logger.warning("Failed #{interface} interface loading.")
-          {:halt, {:error, :sending_properties_to_interface_failed}}
+          {:halt, error}
+
+        {:error, :sending_properties_to_interface_failed} = error ->
+          {:halt, error}
```
noaccOS: 🤖  dialyzer error 🤖 

here we can just _not_ be assertive on the error and we're good to go though


```suggestion
        {:error, reason} = error ->
Logger.warning("Error while sending properties to device: #{inspect(reason)} on interface #{interface}")
          {:halt, error}
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/device.ex
```diff
@@ -437,14 +440,28 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.Device do
...
+             encoded_device_id,
+             interface_descriptor,
+             mapping
+           ) do
+        :ok ->
+          {:cont, :ok}
+
+        {:error, reason} ->
+          Logger.warning(
+            "Failed sending properties for #{interface_descriptor.name} interface: #{inspect(reason)}."
```
noaccOS: nit: I find it a bit weird that the name is before `interface`, I read it as if the reason is the interface (though it might be a me issue)


```suggestion
            "Failed sending properties for interface #{interface_descriptor.name}: #{inspect(reason)}."
```

noaccOS (changes requested):

noaccOS (approved):
🚀 thank you

PR #2120, opened 2026-08-06
fix(appengine): Handle parameters on encrypted endpoints

nedimtokic:
Payload is encrypted for parameterized endpoints.

### DIFF
--- a/apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
+++ b/apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
@@ -33,6 +33,7 @@ defmodule Astarte.AppEngine.API.Device.DeviceReadingV2Test do
   alias Astarte.AppEngine.API.Device.Queries, as: AppEngineDeviceQueries
 
   alias Astarte.Core.Device, as: CoreDevice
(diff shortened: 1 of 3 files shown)


apps/astarte_appengine_api/test/astarte_appengine_api/device/device_reading_v2_test.exs
```diff
@@ -346,22 +349,24 @@ defmodule Astarte.AppEngine.API.Device.DeviceReadingV2Test do
...
 
-  defp encrypted_mapping_path?(interface_to_update, path) do
-    normalized_path = normalize_path(path)
+  defp encrypted_mapping_path?(interface_to_update, path, realm_name) do
+    {:ok, interface_descriptor} =
+      InterfaceQueries.fetch_interface_descriptor(
+        realm_name,
+        interface_to_update.name,
+        interface_to_update.major_version
+      )
```
noaccOS: why do you need the descriptor? doesn't `interface_to_update` have the aggregation set?

nedimtokic: Correct, this is leftover. Fixed.

noaccOS (approved):

PR #2123, opened 2026-08-07
refactor(dup): Move encryption logic into library

nedimtokic:
The device payload is encrypted using the Astarte Secrets library. AppEngine generates a new DEK for each encryption request, while DUP retrieves the required DEK from the DEKCache for encryption.


libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -214,6 +214,41 @@ defmodule Astarte.Secrets do
...
+  @spec encrypt_value(
+          :individual | :object,
+          String.t(),
+          term(),
+          [String.t()],
+          %{plaintext: binary(), ciphertext: String.t()}
+        ) :: {term(), String.t() | nil}
+
+  def encrypt_value(:individual, path, value, encrypted_endpoints, dek_entry) do
+    case path in encrypted_endpoints do
```
noaccOS: I would like this function to not expect the caller to provide a list of encrypted endpoints
instead, it should expect the caller to provide the list of mappings being updated, and then decide whether to encrypt the data based on that. For more context check notion or ask @eddbbt

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -214,6 +214,41 @@ defmodule Astarte.Secrets do
...
+        %{plaintext: plaintext_dek, ciphertext: ciphertext_dek} = dek_entry
+        encrypted_value = do_encrypt_individual_value(value, plaintext_dek)
+        {encrypted_value, ciphertext_dek}
+
+      false ->
+        {value, nil}
+    end
+  end
+
+  def encrypt_value(:object, _path, obj_value, encrypted_endpoints, dek_entry) do
```
noaccOS: the `:individual` path is really a `maybe_encrypt` while the `:object` path always encrypts. let's decide on one and be consistent for both paths

noaccOS (changes requested):

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
nedimtokic: Yes, but also we do not want to generate DEK if we do not have encrypted endpoint ?

noaccOS: The dek should always be present regardless

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
nedimtokic: Changed, `object` path has `case` now.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -141,7 +141,7 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
            interface_descriptor.aggregation,
            mapping.endpoint,
            value,
-           encrypted_endpoints_on_current_interface,
+           state.mappings,
```
noaccOS: `state.mappings` has all mappings, we only want those needed for the current data

https://app.notion.com/p/seco-mind/perf-avoid-cycling-through-state-mappings-for-encrypted-endpoints-3b4c537dba7f80ceb824cfde44a45a70?source=copy_link

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -767,63 +763,13 @@ defmodule Astarte.AppEngine.API.Device do
...
-            acc_obj,
-            endpoint,
-            do_encrypt_individual_value(Map.get(acc_obj, endpoint), plaintext_dek)
-          )
-        end)
-
-      {obj_value_with_encryption, ciphertext_dek}
+  # Retrieve the current realm DEK and delegate endpoint-level encryption logic to Secrets.
+  defp maybe_encrypt_value(aggregation, path, value, mappings, realm) do
+    with {:ok, dek_entry} <- generate_new_dek(realm) do
```
noaccOS: oh now I see appengine does indeed generate a separate dek for all data.
how about we move `DataEncryptionKeyCache` to secrets as well? That way it could handle the dek by itself, without pushing the dek responsibility onto dup and appengine.

you may have astarte_secrets become an otp application or just requiring dup and appengine to start the cache in their supervision trees I don't have a strong opinion on that

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -214,6 +214,56 @@ defmodule Astarte.Secrets do
...
+          [String.t()],
+          %{plaintext: binary(), ciphertext: String.t()}
+        ) :: {term(), String.t() | nil}
+
+  # if value == nil we are trying to unset a property, be it encrypted or not: return nil as it is
+  def maybe_encrypt_value(_, _, nil, _, _) do
+    {nil, nil}
+  end
+
+  def maybe_encrypt_value(:individual, path, value, mappings, dek_entry) do
```
noaccOS: we can probably have a single generic `maybe_encrypt_value` and delegate the specialization to a `do_encrypt_value(aggregation, ...)`, as everything else is (or rather should, the `case` here needs to be updated) equivalent

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -185,71 +187,11 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
...
-    Enum.reduce(encrypted_endpoints, obj_value, fn endpoint, acc_obj ->
-      endpoint = Path.basename(endpoint)
-
-      Map.put(
-        acc_obj,
-        endpoint,
-        do_encrypt_individual_value(Map.get(acc_obj, endpoint), dek)
-      )
-    end)
+  defp encrypted_endpoints_from_mappings(mappings) do
```
noaccOS: next step would probably be removing this
in data_access we only use it to check if we need to encrypt an object value, but we can just as well check if the encrypted dek is nil or not

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -132,16 +137,13 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
...
-      state.mappings
-      |> Map.values()
-      |> Enum.filter(&(&1.interface_id == context.interface_id and &1.encrypted))
-      |> Enum.map(& &1.endpoint)
+      encrypted_endpoints_from_mappings(current_interface_mappings)
 
-    case maybe_encrypt_value(
+    case Secrets.maybe_encrypt_value(
            interface_descriptor.aggregation,
            mapping.endpoint,
```
noaccOS: I suggest passing the whole `interface_descriptor` and `mapping`

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -214,6 +215,55 @@ defmodule Astarte.Secrets do
...
+
+  defp do_encrypt_value(:individual, _path, value, _encrypted_endpoints, dek) do
+    do_encrypt_individual_value(value, dek)
+  end
+
+  defp do_encrypt_value(:object, _path, obj_value, encrypted_endpoints, dek) do
+    do_encrypt_object_mappings(encrypted_endpoints, obj_value, dek)
+  end
+
+  defp encryption_needed?(:individual, path, encrypted_endpoints), do: path in encrypted_endpoints
```
noaccOS: this way here we can just check for `mapping.encrypted`

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -280,4 +330,30 @@ defmodule Astarte.Secrets do
...
+    value_binary = :erlang.term_to_binary(value)
+    {:ok, encrypted_value} = encrypt_with_dek(value_binary, dek)
+    encrypted_value
+  end
+
+  defp encrypted_endpoints_from_mappings(mappings) do
+    mappings
+    |> Map.values()
+    |> Enum.filter(& &1.encrypted)
+    |> Enum.map(& &1.endpoint)
```
noaccOS, line 352: and this can stop at the first mapping found

```suggestion
    |> Enum.any?(& &1.encrypted)
```

and only needs to be called for object aggregates not for all values

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -660,18 +671,41 @@ defmodule Astarte.AppEngine.API.Device do
...
-          acc ++ [Map.get(mapping, :endpoint)]
+  # For individual aggregation we already have the selected endpoint mapping, so we can avoid
+  # scanning all mappings and build a one-element map.
+  defp mappings_for_current_interface(:individual, mapping, _all_mappings, _interface_id) do
+    %{mapping.endpoint_id => mapping}
+  end
 
-        _ ->
+  defp mappings_for_current_interface(:object, _mapping, all_mappings, interface_id) do
+    all_mappings
```
noaccOS: can we also use the result from the EndpoinstAutomaton to get the mappings in AppEngine as we do in dup?

noaccOS (changes requested):

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
nedimtokic: Removed now. Function is used by transport encryption only.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
nedimtokic: Done.

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
nedimtokic: Function is removed now.

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
nedimtokic: Called only for object aggregates, but we need to filter all encrypted endpoints so can encrypt all of them.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Done.

libs/astarte_secrets/test/astarte_secrets/data_encryption_key_cache_test.exs
```diff
@@ -16,11 +16,11 @@
 # limitations under the License.
 #
 
-defmodule Astarte.DataUpdaterPlant.DataEncryptionKeyCacheTest do
+defmodule Astarte.Secrets.DataEncryptionKeyCacheTest do
```
noaccOS, line 19: maybe should be moved to astarte_secrets?

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
noaccOS: The current logic is still looping all interfaces instead of using the automaton

noaccOS (changes requested):

libs/astarte_secrets/test/astarte_secrets/data_encryption_key_cache_test.exs
nedimtokic: Done.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -404,22 +408,34 @@ defmodule Astarte.AppEngine.API.Device do
...
+      )
+
     with {:guessed, guessed_endpoints} <-
            EndpointsAutomaton.resolve_path(path, interface_descriptor.automaton),
-         :ok <- check_object_aggregation_prefix(path, guessed_endpoints, mappings) do
+         :ok <- check_object_aggregation_prefix(path, guessed_endpoints, mappings_by_endpoint_id) do
+      mapping =
+        Enum.find(mappings, fn %Mapping{endpoint_id: guessed_endpoint_id} ->
+          guessed_endpoint_id in guessed_endpoints
+        end)
```
noaccOS: this still needs to cycle all the mapping. let's use

```elixir
interface_mappings = Map.take(mappings, guessed_endpoints) |> Map.values()
```

instead, and verify that `Enum.count(guessed_endpoints) == Enum.count(interface_mappings)`

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -660,18 +677,49 @@ defmodule Astarte.AppEngine.API.Device do
...
   end
 
-  defp extract_encrypted_endpoints(mappings) do
-    Enum.reduce(mappings, [], fn mapping, acc ->
-      case Map.get(mapping, :encrypted) do
-        true ->
-          acc ++ [Map.get(mapping, :endpoint)]
+  # For individual aggregation we already have the selected endpoint mapping, so we can avoid
+  # scanning all mappings and build a one-element list.
+  defp mappings_for_current_interface(
```
noaccOS: this function should be deleted

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -360,6 +303,17 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
 
   defp validate_required_mappings(_context, _interface_descriptor, _mappings, _value), do: :ok
 
+  defp select_data_mapping(
```
noaccOS: this function has a generic name but its function is very much non-generic
If you _need_ the mapping to be encrypted maybe the logic should be reworked a bit.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/interface.ex
```diff
@@ -231,23 +230,12 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.Interface do
...
-          # all mappings (this is enforced by Realm Management when the interface is installed)
-
-          endpoint_id =
-            CQLUtils.endpoint_id(
-              interface_descriptor.name,
-              interface_descriptor.major_version,
-              ""
-            )
+             :ok <- check_object_aggregation_prefix(path, guessed_endpoints, mappings) do
+          all_mappings = Enum.map(guessed_endpoints, &Map.fetch!(mappings, &1))
```
noaccOS:
```suggestion
          all_mappings = Map.take(mappings, guessed_endpoints) |> Map.values()
```

your current implementation crashes if a mapping is not found, instead the error should be handled gracefully

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
```diff
@@ -214,6 +215,62 @@ defmodule Astarte.Secrets do
...
   end
 
+  @doc """
+  Conditionally encrypts a value according to interface aggregation and encrypted endpoints.
+
+  The DEK entry is expected to already be fetched by the caller (for example via an app-level cache).
+  This keeps the secrets library independent from app-specific DEK retrieval strategies.
+  """
+  @spec maybe_encrypt_value(
+          map(),
```
noaccOS: If we're adding specs let's be thorough

`InterfaceDescriptor.t(), Mapping.t(), term(), [Mapping.t()], String.t()`

why can the (so called) `realm_name` be a string or a ` %{plaintext: binary(), ciphertext: String.t()}`?

noaccOS (changes requested):

libs/astarte_secrets/lib/astarte_secrets/astarte_secrets.ex
nedimtokic: Changed. _%{plaintext: binary(), ciphertext: String.t()} _ is leftover from previous version.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/interface.ex
nedimtokic: Done.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
nedimtokic: Function is removed, since we do not have to filer encrypted endpoints here.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Done.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: mappings is not a list, and contains only mappings for single interface.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -404,22 +401,27 @@ defmodule Astarte.AppEngine.API.Device do
...
       Enum.into(mappings, %{}, fn mapping ->
         {mapping.endpoint_id, mapping}
       end)
 
     with {:guessed, guessed_endpoints} <-
            EndpointsAutomaton.resolve_path(path, interface_descriptor.automaton),
-         :ok <- check_object_aggregation_prefix(path, guessed_endpoints, mappings) do
+         :ok <- check_object_aggregation_prefix(path, guessed_endpoints, mappings_by_endpoint_id) do
+      all_mappings =
+        Enum.filter(mappings, fn %Mapping{endpoint_id: guessed_endpoint_id} ->
```
noaccOS, line 412: filter runs in O(n). Let's use `Map.take` instead!

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -660,16 +663,14 @@ defmodule Astarte.AppEngine.API.Device do
...
+  end
 
-        _ ->
-          acc
-      end
-    end)
+  defp encrypted_endpoints_from_mappings(mappings) do
+    mappings
+    |> Enum.filter(& &1.encrypted)
+    |> Enum.map(& &1.endpoint)
```
noaccOS: If I'm not mistaken, this is only used for `should_encrypt_transport_payload?`
No reason to return the whole list of endpoints, we can just make this return the boolean itself

 
```suggestion
    |> Enum.any?(& &1.encrypted)
```

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
```diff
@@ -59,17 +59,26 @@ defmodule Astarte.DataUpdaterPlant.DataUpdater.Core.DataHandler do
...
          :ok <- validate_path(context),
          {:ok, interface_descriptor, context} <- maybe_handle_cache_miss(context),
          :ok <- can_write_on_interface?(context, interface_descriptor.ownership),
-         {:ok, mapping} <- resolve_path(context, interface_descriptor),
+         {:ok, current_interface_mappings} <-
+           resolve_path(context, interface_descriptor),
          {:ok, context} <- maybe_decrypt_payload(context, interface_descriptor),
          {:ok, {value, value_timestamp, _metadata}} <- decode_bson_payload(context),
-         :ok <- validate_value(context, interface_descriptor, mapping, value) do
+         mapping = hd(current_interface_mappings),
```
noaccOS: I don't love this because to me it looks like you're making up the fact that the first mapping is what's important for the value

what about having `resolve_path` (private function, not from the core interface module) return `{:ok, {mapping, current_interface_mappings}}` instead? that way we delegate to the responsibility of finding the correct mapping to it, which is also where it make sense to replace the endpoint id for the mapping

noaccOS (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: mappings is a list. And here in App Engine mappings contains only mappings for single interface.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Removed function.

apps/astarte_data_updater_plant/lib/astarte_data_updater_plant/data_updater/core/data_handler.ex
nedimtokic: Done.

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
```diff
@@ -555,6 +555,8 @@ defmodule Astarte.AppEngine.API.Device do
         raw_value: raw_value,
         value: value,
         mappings: mappings,
+        mapping: hd(current_interface_mappings),
```
noaccOS: this is still using the first mapping as a special mapping which is what I wanted to avoid with my previous comment

noaccOS (changes requested):

apps/astarte_appengine_api/lib/astarte_appengine_api/device/device.ex
nedimtokic: Previous was in DUP. Now I done it in App Engine also.

noaccOS (approved):

PR #2126, opened 2026-08-07
Add vmq plugin to monorepo

mizzet1:
This PR brings `astarte_vmq_plugin` into the monorepo under `apps/astarte_vmq_plugin/`, with its full commit history intact, and adds it into the monorepo's CI/CD.

The import was done with `git subtree add --prefix=apps/astarte_vmq_plugin`, which produces a single merge commit with two parents, the monorepo's own `master` tip and `astarte_vmq_plugin`'s `master` tip 

Lint and test CI now goes through the shared `astarte-apps-build-workflow.yaml`, same as every other app in `apps/`. 

For the Docker side, since `astarte_vmq_plugin` builds a full VerneMQ binary from source rather than doing a `mix release`, a dedicated snapshot/release publish workflows was added instead, mirroring how `astarte-dashboard` is already handled as the repo's other non-standard build. 
Both publish `astarte/vernemq`, which `docker-compose.yml` was already pulling as a published image.

No other app depends on `astarte_vmq_plugin` as a mix dependency.


.github/workflows/publish-vmq-plugin-release-to-dockerhub-workflow.yaml
```diff
@@ -0,0 +1,106 @@
+name: Publish vmq_plugin release Docker images
```
noaccOS, line 1: ok for now, eventually it will probably be best to have a single workflow with everything instead, it's getting a little too spaghetti

.github/workflows/publish-vmq-plugin-release-to-dockerhub-workflow.yaml
```diff
@@ -0,0 +1,106 @@
...
+    permissions:
+      packages: write
+      contents: read
+      attestations: write
+      id-token: write
+    strategy:
+      fail-fast: true
+      matrix:
+        platform:
+          - os: ubuntu-22.04
```
noaccOS: let's use 24.04, as we'll do everywhere after #2125 is merged

noaccOS (changes requested):
maybe missing codecov components changes?

apps/astarte_vmq_plugin/.credo.exs
```diff
@@ -0,0 +1,26 @@
+%{
```
noaccOS: I suggest removing this otherwise we're not using the global config

apps/astarte_vmq_plugin/.tool-versions
```diff
@@ -0,0 +1,2 @@
+elixir 1.15.7-otp-26
```
noaccOS: let's remove this file as we're using the root tool-versions instead

apps/astarte_vmq_plugin/codecov.yaml
```diff
@@ -0,0 +1,4 @@
+coverage:
```
noaccOS: I'm not sure how this behaves on ci, I'd suggest removing it and adding vmq plugin to the current `codecod.yaml` configuration instead

apps/astarte_vmq_plugin/.dockerignore
```diff
@@ -0,0 +1,25 @@
+# Standard Elixir gitignore to be applied to all apps
```
noaccOS, line 1: eventually this file will need to be removed, up to you if you want to do it in this pr

apps/astarte_vmq_plugin/.gitignore
```diff
@@ -0,0 +1,20 @@
+# The directory Mix will write compiled artifacts to.
```
noaccOS: eventually this file will need to be removed, up to you if you want to do it in this pr

noaccOS (changes requested):
the scope for vmq plugin should also be added to `committed.toml`

I would suggest using `vmq_plugin` as all our multi-word scopes use underscores

noaccOS (approved):

nedimtokic (approved):

PR #2127, opened 2026-08-07
chore(config): increase coverage

noaccOS:
https://github.com/parroty/excoveralls/issues/59#issuecomment-2616559261

depends on #2125


noaccOS:
integrated in #2125

PR #2139, opened 2026-08-14
Add astarte_core to monorepo

mizzet1:
This PR brings astarte_core into the monorepo under libs/astarte_core/, with its full commit history intact, and adds it into the monorepo's CI/CD.

The import was done with git subtree add --prefix=libs/astarte_core, which produces a single merge commit with two parents, the monorepo's own master tip and astarte_core's master tip.

Lint and test CI now goes through the shared astarte-libs-build-workflow.yaml, same as every other library in libs/.


libs/astarte_events/mix.exs
```diff
@@ -64,7 +64,7 @@ defmodule Astarte.Events.Mixfile do
       {:mimic, "~> 2.3", only: [:dev, :test]},
       {:dialyxir, "~> 1.0", only: [:dev, :test], runtime: false},
       {:astarte_data_access, path: "../astarte_data_access"},
-      {:astarte_core, github: "astarte-platform/astarte_core", override: true},
+      {:astarte_core, path: "../astarte_core", override: true},
```
noaccOS: is `override: true` still needed? I think we should now have ownership of the whole tree

noaccOS (approved):

PR #2144, opened 2026-08-19
chore: Update exandra

nedimtokic:
- Use `exandra` fork until PR is accepted on the original repository.
- Run mix deps.clean --unused to remove unused deps.


apps/astarte_pairing/mix.exs
```diff
@@ -97,8 +97,7 @@ defmodule Astarte.Pairing.Mixfile do
       {:astarte_secrets, path: astarte_lib("astarte_secrets"), override: true},
       {:bcrypt_elixir, "~> 2.2"},
       {:ecto, "~> 3.13", override: true},
```
noaccOS: we can drop this

apps/astarte_vmq_plugin/mix.exs
```diff
@@ -91,9 +91,8 @@ defmodule Astarte.VMQ.Plugin.Mixfile do
       {:dialyxir, "~> 1.4", only: [:dev, :ci, :test], runtime: false},
       {:credo, "~> 1.7", only: [:dev, :test], runtime: false},
       {:mix_audit, "~> 2.1", only: [:dev, :test], runtime: false},
-      {:xandra, "~> 0.14"},
+      {:xandra, "~> 0.20"},
```
noaccOS, line 94: oh yeah vmq plugin is using plain xandra
might need to fix that for consistency (not in this pr just a thought)

libs/astarte_data_access/mix.lock
```diff
@@ -7,17 +7,17 @@
...
   "earmark_parser": {:hex, :earmark_parser, "1.4.44", "f20830dd6b5c77afe2b063777ddbbff09f9759396500cdbe7523efd58d7a339c", [:mix], [], "hexpm", "4778ac752b4701a5599215f7030989c989ffdc4f6df457c5f36938cc2d2a2750"},
-  "ecto": {:hex, :ecto, "3.13.5", "9d4a69700183f33bf97208294768e561f5c7f1ecf417e0fa1006e4a91713a834", [:mix], [{:decimal, "~> 2.0", [hex: :decimal, repo: "hexpm", optional: false]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: true]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "df9efebf70cf94142739ba357499661ef5dbb559ef902b68ea1f3c1fabce36de"},
+  "ecto": {:hex, :ecto, "3.13.6", "352135b474f91d1ab99a1b502171d207e9db60421c9e3d0ecab4c7ab96b24d14", [:mix], [{:decimal, "~> 2.0 or ~> 3.0", [hex: :decimal, repo: "hexpm", optional: false]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: true]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "8afa059bc16cd2c94739ec0a11e3e5df69d828125119109bef35f20a21a76af2"},
```
noaccOS: weird in other apps we have ecto 3.14
would it be possible to have uniform versions? just so that if we do find an issue somewhere we don't have to consider distinct ecto behavior (looking at you `:label`)

noaccOS (changes requested):

libs/astarte_data_access/mix.lock
nedimtokic: Updated with `mix deps.update ecto_sql.`

noaccOS (approved):
Thanks!

PR #2146, opened 2026-08-20
fix: route AppEngine room events across replicas

### LINKED ISSUE #2077: AppEngine Channels: with more than one AppEngine replica, room events are delivered to the wrong instance and the volatile trigger is erroneously deleted
## Summary

When AppEngine API runs with **more than one replica**, Astarte Channels (Rooms) deliver only the first few events to a connected WebSocket client and then stop entirely, while the connection stays open and the data is correctly persisted in the database. Worse, the **volatile trigger** backing the room is **actively deleted**, so the loss becomes permanent until the client re-issues `watch`.

The root cause is architectural and confirmed in the source (identical on `v1.0.4` and current `master`): the rooms-events AMQP queue is a **single shared queue** consumed by all AppEngine instances (competing consumers), while the `Room` process is registered in a **node-local `Registry`**. When RabbitMQ delivers an event to an instance that does not host the room, dispatch fails and the code deletes the volatile trigger.

## Observed behavior

1. Client connects to the Channels socket, joins `rooms:<realm>:<name>`, and pushes a `watch` for a device/interface/path. The `watch` reply is `ok`.
2. The device produces data (verified end-to-end: values are correctly persisted and readable via the REST API).
3. The WebSocket receives only the **first 0–2 events**, then nothing more. The connection stays open (no disconnect).
4. The behavior is **non-deterministic** across attempts, and two simultaneous connections watching the same trigger receive **different subsets** of events (e.g. one gets 2, the other 0).

## Code evidence

Modules are identical between `v1.0.4` and `master` for this path.

1. **Single shared queue + constant routing key** — `apps/astarte_appengine_api/lib/astarte_appengine_api/config.ex`:
   - `rooms_events_queue_name` defaults to `"astarte_rooms_events"` (a constant, with no per-node/per-instance suffix);
   - `rooms_events_routing_key!/0` returns the constant `"astarte_rooms"`.

2. **Every instance declares and consumes the same queue** — `apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/amqp_client.ex` performs `Queue.declare(... rooms_events_queue_name ...)` followed by `Basic.consume(...)` on that constant name. With N AppEngine instances this is a classic RabbitMQ **competing-consumers** setup: each event is delivered to exactly one instance.

3. **The room is registered in a node-local Registry** — `apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/room.ex` calls `Registry.register(Registry.AstarteRooms, {:parent_trigger_id, room_uuid}, [])`, and `apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/master_supervisor.ex` declares `{Registry, keys: :unique, name: Registry.AstarteRooms}`. This Registry is local to the BEAM node (not `:global`; no `libcluster`/`horde`/`swarm` in `mix.exs`).

4. **Destructive dispatch on the wrong instance** — `apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/events_dispatcher.ex`:
   ```elixir
   {:room_pid, []} ->
     Logger.warn("Dispatch: unexisting room for event #{inspect(simple_event)}.", tag: "unexisting_room_for_event")
     :telemetry.execute([:astarte, :appengine, :channels, :event_discarded], %{}, %{realm: realm})
     DataUpdaterPlant.delete_volatile_trigger(realm, device_id, simple_trigger_id)
     {:error, :no_room_for_event}
   ```
   When the event lands on an instance that does not host the room, the code assumes the room no longer exists and **deletes the volatile trigger**.

Because the publish routing key (`"astarte_rooms"`) carries no room/instance identity (`parent_trigger_id`/`simple_trigger_id` are only in the protobuf payload and AMQP headers), the broker cannot route the event to the instance that holds the room. With N replicas each event reaches the correct instance with probability ~1/N, and the **first miss deletes the trigger**, which explains the "first 0–2 events, then permanent silence" pattern.

## Affected versions

- Reproduced on **AppEngine API 1.0.4**.
- The relevant Rooms modules are **unchanged on `master`**.

## Relation to #699

#699 covers the mirror problem for a **replicated Data Updater Plant** (the AppEngine→DUP RPC used to install/delete volatile triggers). This report is the complementary direction: a **replicated AppEngine** and the DUP→AppEngine rooms-events path via the `astarte_rooms_events` queue. The RPC-side considerations in #699 do not address this queue.

## Impact

- Astarte Channels are effectively unusable with more than one AppEngine replica: realtime delivery silently degrades and volatile triggers are removed.
- Single-replica AppEngine works correctly (confirmed by falling back to one replica).

## Possible directions (for maintainers to evaluate)

- A per-instance queue (e.g. queue name suffixed with an instance/node identifier) so each instance only receives events for rooms it hosts; or
- A distributed process registry (`:global` / `pg` / Horde) with cluster-aware dispatch, so any instance can locate the room that owns the event; or
- At minimum, avoid deleting the volatile trigger in the `no_room_for_event` branch unless it can be established that no instance hosts the room (so a misrouted event does not permanently break a healthy subscription).

## Workaround

Run AppEngine API with a **single replica** when using Astarte Channels.

## Note on reproducibility

The effect is non-deterministic and depends on which replica RabbitMQ delivers each event to; on a single-replica deployment it does not manifest at all.

mvanhorn:

Replace the local Rooms registry child in `MasterSupervisor` with a unique, auto-membered `Horde.Registry`, retaining its supervision ordering with `RoomsSupervisor`. Update Room registration and `:via` naming, room-existence checks, and event dispatch lookup to consistently use the Horde registry so any AMQP consumer replica can resolve and call the Room process on its owning node. Keep the shared queue and existing volatile-trigger lifecycle: `EventsDispatcher` should still delete a trigger when the cluster-wide lookup proves that no Room owns the parent trigger, and when an existing Room reports that the simple trigger is no longer watched.

AppEngine replicas currently compete on one Rooms AMQP queue, but each replica can only find Room processes in its node-local `Registry.AstarteRooms`. An event consumed on a replica other than the Room owner is therefore classified as targeting a dead room, discarded, and followed by deletion of its still-live volatile trigger. This makes Channels delivery nondeterministic and permanently stops the watch after the first misrouted event. The repository already uses auto-membered Horde registries for cross-node RPC discovery, providing an established cluster-aware registration pattern for Rooms.

Fixes #2077

---

<details>

> `fixes #<issue number>`

> `depends on #<pr number>`

> ```mermaid
> gitGraph
>     branch feat1
>     checkout feat1
>     branch feat2
>     checkout feat2
>     branch feat3
>     checkout feat3
> ```

</details>

### DIFF
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/events_dispatcher.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/events_dispatcher.ex
@@ -25,6 +25,9 @@ defmodule Astarte.AppEngine.API.Rooms.EventsDispatcher do
   alias Astarte.AppEngine.API.RPC.DataUpdaterPlant
   alias Astarte.Core.Triggers.SimpleEvents.SimpleEvent
 
+  @room_lookup_attempts 3
+  @room_lookup_retry_interval_ms 100
+
   require Logger
 
   def dispatch(serialized_simple_event) do
@@ -49,8 +52,7 @@ defmodule Astarte.AppEngine.API.Rooms.EventsDispatcher do
       end
 
     with {:room_pid, [{pid, _}]} <-
-           {:room_pid,
-            Registry.lookup(Registry.AstarteRooms, {:parent_trigger_id, parent_trigger_id})},
+           {:room_pid, lookup_room({:parent_trigger_id, parent_trigger_id})},
          :ok <- Room.broadcast_event(pid, simple_trigger_id, device_id, timestamp, event) do
       :ok
     else
@@ -101,6 +103,23 @@ defmodule Astarte.AppEngine.API.Rooms.EventsDispatcher do
         )
 
         {:error, :dispatch_error}
+    end
+  end
+
+  defp lookup_room(parent_trigger_key, attempts \\ @room_lookup_attempts)
+
+  defp lookup_room(parent_trigger_key, 1) do
+    Horde.Registry.lookup(Registry.AstarteRooms, parent_trigger_key)
+  end
+
+  defp lookup_room(parent_trigger_key, attempts) do
+    case Horde.Registry.lookup(Registry.AstarteRooms, parent_trigger_key) do
+      [] ->
+        Process.sleep(@room_lookup_retry_interval_ms)
+        lookup_room(parent_trigger_key, attempts - 1)
+
+      room ->
+        room
     end
   end
 end
--- a/apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/rooms_supervisor.ex
+++ b/apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/rooms_supervisor.ex
@@ -32,11 +32,20 @@ defmodule Astarte.AppEngine.API.Rooms.RoomsSupervisor do
   end
 
   def start_room(realm, room_name) do
-    DynamicSupervisor.start_child(__MODULE__, {Room, realm: realm, room_name: room_name})
+    case DynamicSupervisor.start_child(__MODULE__, {Room, realm: realm, room_name: room_name}) do
+      :ignore ->
+        case Horde.Registry.lookup(Registry.AstarteRooms, room_name) do
+          [{pid, _value}] -> {:ok, pid}
+          [] -> {:error, :room_not_registered}
+        end
+
+      result ->
+        result
+    end
   end
 
   def room_started?(room_name) do
-    case Registry.lookup(Registry.AstarteRooms, room_name) do
+    case Horde.Registry.lookup(Registry.AstarteRooms, room_name) do
       [] ->
         false
 
--- a/apps/astarte_appengine_api/test/astarte_appengine_api_web/channels/rooms_channel_test.exs
+++ b/apps/astarte_appengine_api/test/astarte_appengine_api_web/channels/rooms_channel_test.exs
@@ -588,10 +588,11 @@ defmodule Astarte.AppEngine.APIWeb.RoomsChannelTest do
       refute_broadcast "new_event", %{"device_id" => @device_id, "event" => _event}
     end
 
-    test "a transient registry miss does not uninstall a live room trigger", %{
-      socket: socket,
-      room_process: room_process
-    } = context do
+    test "a transient registry miss does not uninstall a live room trigger",
+         %{
+           socket: socket,
+           room_process: room_process
+         } = context do
       :ok = Mimic.set_mimic_private(context)
       Mimic.verify_on_exit!(context)
 
--- a/apps/astarte_appengine_api/test/astarte_appengine_api_web/channels/rooms_channel_test.exs
+++ b/apps/astarte_appengine_api/test/astarte_appengine_api_web/channels/rooms_channel_test.exs
@@ -23,6 +23,7 @@ defmodule Astarte.AppEngine.APIWeb.RoomsChannelTest do
   alias Astarte.AppEngine.API.Groups
   alias Astarte.AppEngine.API.Rooms.EventsDispatcher
   alias Astarte.AppEngine.API.Rooms.Room
+  alias Astarte.AppEngine.API.Rooms.RoomsSupervisor
   alias Astarte.AppEngine.API.Utils
   alias Astarte.AppEngine.APIWeb.RoomsChannel
   alias Astarte.AppEngine.APIWeb.UserSocket
@@ -175,6 +176,20 @@ defmodule Astarte.AppEngine.APIWeb.RoomsChannelTest do
       assert {:ok, _reply, socket} = join(socket, "rooms:#{@realm}:#{@authorized_room_name}")
       assert socket.assigns[:room_name] == "#{@realm}:#{@authorized_room_name}"
       assert Room.clients_count("#{@realm}:#{@authorized_room_name}") == 1
+    end
+
+    test "joining the same room resolves one cluster-wide room process", %{socket: first_socket} do
+      room_name = "#{@realm}:#{@authorized_room_name}"
+      token = JWTTestHelper.gen_channels_jwt_all_access_token()
+      {:ok, second_socket} = connect(UserSocket, %{"realm" => @realm, "token" => token})
+
+      assert {:ok, _reply, _first_socket} = join(first_socket, "rooms:#{room_name}")
+      assert {:ok, _reply, _second_socket} = join(second_socket, "rooms:#{room_name}")
+
+      assert [{room_process, _}] = Horde.Registry.lookup(Registry.AstarteRooms, room_name)
+      assert {:ok, ^room_process} = RoomsSupervisor.start_room(@realm, room_name)
+      assert Room.clients_count(room_name) == 2
+      assert Process.alive?(room_process)
     end
   end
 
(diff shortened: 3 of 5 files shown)


apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/events_dispatcher.ex
```diff
@@ -103,4 +105,21 @@ defmodule Astarte.AppEngine.API.Rooms.EventsDispatcher do
         {:error, :dispatch_error}
     end
   end
+
+  defp lookup_room(parent_trigger_key, attempts \\ @room_lookup_attempts)
```
noaccOS, line 109: I don't get why you'd need multiple lookup attempts here. Can you give an example of a failure encountered with the singular lookup which is fixed using this logic?

apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/rooms_supervisor.ex
```diff
@@ -32,11 +32,20 @@ defmodule Astarte.AppEngine.API.Rooms.RoomsSupervisor do
   end
 
   def start_room(realm, room_name) do
-    DynamicSupervisor.start_child(__MODULE__, {Room, realm: realm, room_name: room_name})
+    case DynamicSupervisor.start_child(__MODULE__, {Room, realm: realm, room_name: room_name}) do
```
noaccOS, line 35: We can avoid the double lookup if you keep the `{:error, {:already_started, pid}}` as return value. I think it doesn't really matter for dynamic supervisor, both situation should just *not add the child*:

> If the child process start function returns `:ignore`, then no child is added to the supervision tree and this function returns `:ignore` too.
> 
> If the child process start function returns an error tuple or an erroneous value, or if it fails, the child specification is discarded and this function returns `{:error, error}` where error is the error or erroneous value returned from child process start function, or failure reason if it fails.

https://elixir.hexdocs.pm/DynamicSupervisor.html#start_child/2

apps/astarte_appengine_api/test/astarte_appengine_api_web/channels/rooms_channel_test.exs
```diff
@@ -176,6 +177,20 @@ defmodule Astarte.AppEngine.APIWeb.RoomsChannelTest do
       assert socket.assigns[:room_name] == "#{@realm}:#{@authorized_room_name}"
       assert Room.clients_count("#{@realm}:#{@authorized_room_name}") == 1
     end
+
+    test "joining the same room resolves one cluster-wide room process", %{socket: first_socket} do
```
noaccOS, line 181: this is not really testing anything. I think that if you remove the `Horde.` this test also passes on master.
the registry does not have multiple members

apps/astarte_appengine_api/test/astarte_appengine_api_web/channels/rooms_channel_test.exs
```diff
@@ -573,6 +588,59 @@ defmodule Astarte.AppEngine.APIWeb.RoomsChannelTest do
       refute_broadcast "new_event", %{"device_id" => @device_id, "event" => _event}
     end
 
+    test "a transient registry miss does not uninstall a live room trigger", %{
+      socket: socket,
+      room_process: room_process
+    } = context do
+      :ok = Mimic.set_mimic_private(context)
```
noaccOS, line 596: this is the default behavior


```suggestion
```

apps/astarte_appengine_api/test/astarte_appengine_api_web/channels/rooms_channel_test.exs
```diff
@@ -573,6 +588,59 @@ defmodule Astarte.AppEngine.APIWeb.RoomsChannelTest do
...
+
+      ref = push(socket, "watch", watch_payload)
+      assert_broadcast("watch_added", _)
+      assert_reply(ref, :ok, %{})
+
+      %{room_uuid: room_uuid, watch_id_to_request: watch_map} = :sys.get_state(room_process)
+      [simple_trigger_id | _] = Map.keys(watch_map)
+      parent_trigger_key = {:parent_trigger_id, room_uuid}
+
+      Mimic.expect(Horde.Registry, :lookup, 2, fn Registry.AstarteRooms, ^parent_trigger_key ->
```
noaccOS, line 621: I have found expectations on horde registry processes to be quite flaky, it would be better to rework your code to expose a `@doc false` public function with does the lookup and place an expectation on that instead, _if you really need this retry logic_

apps/astarte_appengine_api/test/astarte_appengine_api_web/channels/rooms_channel_test.exs
```diff
@@ -599,11 +667,20 @@ defmodule Astarte.AppEngine.APIWeb.RoomsChannelTest do
...
+
       existing_room_serialized_event =
         %{@simple_event | parent_trigger_id: room_uuid, simple_trigger_id: simple_trigger_id}
         |> SimpleEvent.encode()
 
-      assert :ok = EventsDispatcher.dispatch(existing_room_serialized_event)
+      dispatcher_task =
+        Task.async(fn -> EventsDispatcher.dispatch(existing_room_serialized_event) end)
+
+      assert :ok = Task.await(dispatcher_task)
```
noaccOS, line 684: what

noaccOS (changes requested):

mvanhorn:
Pushed a rebase onto current master that clears the four failing checks.

- Formatting: `rooms_channel_test.exs` was not formatter-clean. `mix format --check-formatted` now passes for `apps/astarte_appengine_api`.
- check-fast-forward: the branch is rebased onto master and fast-forwardable again.
- DCO: `fix: route AppEngine room events across replicas` was missing its `Signed-off-by`. Both commits now carry one.

No behaviour change in this push beyond the formatting of that test file.

noaccOS:
> No behaviour change in this push beyond the formatting of that test file.

Thanks for fixing the ci @mvanhorn but if you ignore the review the pr will not move forward :sushi:

mvanhorn:
Sorry for the delay on the review itself.

On the horde registry expectation - I want to check I'm reading you right before I rework it. The "if you really need this retry logic" reads like you suspect the retry isn't warranted at all. Two options as I see them:

1. Drop lookup_room entirely and call Horde.Registry.lookup directly, which removes the flaky expectation along with the retry.
2. Keep the retry but expose it as a @doc false function and put the expectation on that instead, as you suggested.

I lean toward 1 if you're confident the room is always registered by the time dispatch runs - the retry was there to paper over a startup race I saw, but I may have been chasing a test artifact. Which would you prefer?

apps/astarte_appengine_api/lib/astarte_appengine_api/rooms/events_dispatcher.ex
noaccOS: let's keep the conversation here @mvanhorn to avoid cluttering the page.

> The "if you really need this retry logic" reads like you suspect the retry isn't warranted at all.

I am not aware of failures on the current master, so I figured you added the retry logic because you did encounter an error? 

If you didn't, then I wouldn't bother modifying the logic at all
If you did, we can think about a proper solution (eg waiting for the room to be ready before handling other messages) instead, I don't think a retry logic is the best approach here

mvanhorn:
Honest answer: no, I did not hit this on master. The retry was defensive, added because the Horde registry lookup looked racy to me on a cold room, not because I observed a failure. You are right to push on it, and I should have said so when you first asked rather than defending the shape of the retry.

Given that, I do not think this PR should land. `@room_lookup_attempts` plus a `Process.sleep` is a workaround for a race I never actually reproduced, and your point about the tests holds too: the `rooms_channel_test` cases I added pass on master with `Horde.` removed, so they were not testing distribution at all.

Happy to close it. If a real lookup race does turn up later, the readiness-wait approach you suggested sounds like the right shape rather than retrying, and I would rather come back with a reproduction first.

Thanks for the review time on something that turned out not to be needed.

eddbbt:
Closing this PR as the underlying race condition was never actually reproduced and the proposed logic was purely speculative.
 
We cannot merge defensive workarounds, like arbitrary Process.sleep calls and blind retries, for hypothetical issues, especially since the added tests were not actually evaluating the distribution as intended. To keep the codebase maintainable, we need to ensure that issues are explicitly reproduced and solutions are verified prior to submission.
 
If actual reproduction steps for a lookup race are documented in the future, we can revisit the readiness-wait approach.

PR #2156, opened 2026-08-28
feat(fdo): add ownership voucher deletion

mizzet1:
This PR aim to add the delete ownership voucher feature for FDO in Astarte


apps/astarte_pairing/lib/astarte_pairing_web/controllers/ownership_voucher_controller.ex
```diff
@@ -196,4 +236,10 @@ defmodule Astarte.PairingWeb.OwnershipVoucherController do
 
   defp ensure_ownership_voucher_parameter(_params),
     do: {:error, :missing_ownership_voucher}
+
+  defp decode_guid(guid_str) do
+    {:ok, UUID.string_to_binary!(guid_str)}
+  rescue
+    ArgumentError -> {:error, :not_found}
```
noaccOS: personally not a fan of `rescue`, but this is very limited in scope so I'd accept it as it is. In general you can:

```suggestion
    case Ecto.UUID.dump(guid_str) do
      {:ok, _} = ok -> ok
      :error -> {:error, :not_found}
    end
```

apps/astarte_pairing/lib/astarte_pairing_web/controllers/ownership_voucher_controller.ex
```diff
@@ -196,4 +236,10 @@ defmodule Astarte.PairingWeb.OwnershipVoucherController do
 
   defp ensure_ownership_voucher_parameter(_params),
     do: {:error, :missing_ownership_voucher}
+
+  defp decode_guid(guid_str) do
+    {:ok, UUID.string_to_binary!(guid_str)}
```
noaccOS: are all guids parsed correctly by UUID.string_to_binary?
The function is expected to work for valid uuids, I think the guid could be any 16 byte identifier

libs/astarte_fdo/lib/ownership_voucher/ownership_voucher.ex
```diff
@@ -36,6 +40,32 @@ defmodule Astarte.FDO.OwnershipVoucher do
     Queries.list_ownership_vouchers(realm_name)
   end
 
+  @doc """
+  Deletes an ownership voucher.
+  """
+  def delete(realm_name, guid) do
+    with {:ok, voucher_cbor} <- Queries.get_ownership_voucher(realm_name, guid) do
+      revoke_rendezvous_registration(realm_name, guid, voucher_cbor)
```
noaccOS: if we fail to revoke it we should not delete it from the database. let's add it a a with clause instead

libs/astarte_fdo/test/astarte_fdo/ownership_voucher_test.exs
```diff
@@ -26,8 +27,12 @@ defmodule Astarte.FDO.OwnershipVoucherTest do
   alias Astarte.FDO.Core.PublicKey
   alias Astarte.FDO.Helpers
   alias Astarte.FDO.OwnershipVoucher
+  alias Astarte.FDO.TO0
+  alias Astarte.Secrets
   alias COSE.Messages.Sign1
 
+  setup :verify_on_exit!
```
noaccOS: not needed when you `use` Mimic


```suggestion
```

astarte-dashboard/src/FdoVouchersPage.tsx
```diff
@@ -142,10 +145,14 @@ const VoucherRow = ({ voucher, onView }: VoucherRowProps): React.ReactElement =>
...
     </td>
     <td className="text-end">
-      <Button variant="outline-secondary" size="sm" onClick={onView}>
+      <Button variant="outline-secondary" size="sm" onClick={onView} className="me-2">
         <Icon icon="documentation" className="me-1" />
         View details
       </Button>
+      <Button variant="outline-danger" size="sm" onClick={onDelete}>
+        <Icon icon="delete" className="me-1" />
+        Delete
```
noaccOS, line 154: I'll have someone else review the UI ☕

noaccOS (changes requested):
this should be a feat because you allow an additional action to the users of astarte

AliouneGaye21 (approved):

noaccOS (approved):

PR #2166, opened 2026-09-02
refactor: init umbrella skeleton

noaccOS:
for now no app is using the umbrella configuration, but this serves
as a skeleton for future work


frossq (approved):
☔

PR #2169, opened 2026-09-02
refactor: make existing apps/ build part of the umbrella

noaccOS:
- remove per-app `mix.lock`
- use upstream `typedstruct` dependency as latest published version does
  not compile properly in umbrellas
- for now, config/ is kept separate for each app, with the exception of
  events consumer for trigger engine which would fail dialyzer if not
  replicated top level.
- move rel/env.sh.eex to umbrella root
- remove outdated rel/config.exs from apps


docker.log
frossq, line 1: File needed? 🤔

noaccOS: no lmao it was for debugging the docker build

eddbbt (approved):

frossq (approved):
🏖️

PR #2178, opened 2026-09-04
Feat/generators test suite support

shinnokdisengir:

---

<details>

> `fixes #<issue number>`

> `depends on #<pr number>`

> ```mermaid
> gitGraph
>     branch feat1
>     checkout feat1
>     branch feat2
>     checkout feat2
>     branch feat3
>     checkout feat3
> ```

</details>


libs/astarte_generators/lib/astarte/common/generators/uuid.ex
```diff
@@ -0,0 +1,38 @@
...
+  use ExUnitProperties
+
+  @typedoc "A raw RFC 4122 UUID."
+  @type t :: <<_::128>>
+
+  @doc """
+  Generates a random RFC 4122 version 4 UUID.
+  """
+  @spec uuid() :: StreamData.t(t())
+  def uuid do
```
noaccOS: do we need this? I think in astarte we used `repeatedly(Ecto.UUID.bingenerate/0)`, but if you prefer to have this ad hoc implementation it works for me

libs/astarte_generators/lib/astarte/core/generators/triggers/trigger.ex
```diff
@@ -0,0 +1,86 @@
...
+        policy: policy
+      }
+    end
+  end
+
+  defp simple_trigger_ids(trigger_uuid),
+    do:
+      uuid()
+      |> filter(&(&1 != trigger_uuid))
+      |> uniq_list_of(min_length: 1, max_length: 8)
```
noaccOS: technically correct, but practically we only support **one** simple trigger for each trigger

libs/astarte_generators/test/astarte/common/generators/uuid_test.exs
```diff
@@ -0,0 +1,38 @@
...
+
+  @moduletag :common
+  @moduletag :uuid
+
+  describe "UUID generator" do
+    property "generates RFC 4122 version 4 UUIDs" do
+      check all uuid <- uuid() do
+        <<_prefix::48, version::4, _middle::12, variant::2, _suffix::62>> = uuid
+
+        assert byte_size(uuid) == 16 and version == 4 and variant == 2
```
noaccOS: what about

```suggestion
        assert {:ok, _} = Ecto.UUID.cast(uuid)
```

it's simpler and we leave the validation to ecto

libs/astarte_generators/test/astarte/core/generators/triggers/simple_configs/scope_test.exs
```diff
@@ -0,0 +1,47 @@
...
+
+  import Astarte.Core.Generators.Triggers.SimpleConfigs.Scope
+
+  alias Astarte.Core.Device
+  alias Astarte.Core.Group
+
+  describe "triggers scope generator" do
+    property "generates the any-device scope" do
+      check all scope <- trigger_scope(type: :any_device) do
+        assert %{device_id: "*", group_name: nil} = scope
```
noaccOS: while valid, from the [doc](https://docs.astarte-platform.org/astarte/latest/060-triggers.html#device-triggers):

> `device_id` can be used to pass a specific Device ID to restrict the trigger to a single device. `*` is also accepted as `device_id` to maintain backwards compatibility and it is considered equivalent to no `device_id` specified.

I suggest using the non-legacy representation where _both_ `device_id` and `group_name` are not present in the config

libs/astarte_generators/test/astarte/core/generators/triggers/trigger_test.exs
```diff
@@ -0,0 +1,79 @@
...
+          simple_triggers_uuids: simple_triggers_uuids,
+          action: action,
+          name: name,
+          policy: policy
+        } = trigger
+
+        action = Jason.decode!(action)
+        %URI{scheme: scheme} = action |> Map.fetch!("http_url") |> URI.parse()
+
+        assert trigger == trigger |> Trigger.encode() |> Trigger.decode() and
```
noaccOS: if this assertion becomes `false` we only see `expected true` on the whole condition
using separate assertion would help isolate where the problem is as the single failing assertion would be visible

noaccOS (approved):
just one nit, but otherwise ready

noaccOS (approved):

PR #2187, opened 2026-09-08
refactor(hk): use Umbrella configuration

nedimtokic:
Move Housekeeping configuration to the centralized Umbrella config


noaccOS (approved):
☂️

PR #2188, opened 2026-09-08
refactor(pairing): use Umbrella config

nedimtokic:
Move Pairing config into centralized Umbrella config.


config/config.exs
```diff
@@ -102,4 +104,25 @@ config :astarte_housekeeping, Astarte.HousekeepingWeb.Endpoint,
...
+config :astarte_fdo, :endpoint, Astarte.PairingWeb.Endpoint
+
+config :astarte_pairing, Astarte.PairingWeb.AuthGuardian,
+  allowed_algos: ["ES256", "ES384", "ES512", "PS256", "PS384", "PS512", "RS256", "RS384", "RS512"]
+
+config :mime, :types, %{
+  "application/cbor" => ["cbor"]
+}
+
+config :astarte_rpc, :astarte_services, [:astarte_realm_management]
```
noaccOS: this should be included in `config/runtime.exs` instead, in the `case service do`

noaccOS (changes requested):

noaccOS (approved):
🌂

PR #2189, opened 2026-09-08
chore(fdo): improve validations for Ownership Voucher upload/delete requests

frossq:
Trying to delete an OV registered in a realm for which the user is not authenticated results in 404 error

Also, reject installation of OVs with duplicated GUID ad Queries level

#### Additional documentation e.g. usage docs, diagrams, reviewer notes, etc.:

---

<details>
<summary><i>Thanks for sending a pull request! If this is your first time, here are some tips for you:</i></summary>

1. You can take a look at our [developer guide] for an introduction on Astarte development!
2. Make sure to read [CONTRIBUTING.md] and [CODE_OF_CONDUCT.md]
3. If the PR is unfinished or you're actively working on it, mark it as draft

When fixing existing issues, use [github's syntax to link your pull request] to it

> `fixes #<issue number>`

We also have a syntax to signal dependencies to other open pull requests

> `depends on #<pr number>`
> `depends on https://github.com/...`

In case of stacked PRs, you may add the PR number in the last commit's title instead:

> ```mermaid
> gitGraph
>     commit id: "Current master"
>     branch feat1
>     checkout feat1
>     commit id: "feat: add something"
>     commit id: "feat: add something else (#100)"
>     branch feat2
>     checkout feat2
>     commit id: "refactor: do something"
>     commit id: "fix: solve issue"
>     commit id: "feat: add a feature (#101)"
>     branch feat3
>     checkout feat3
>     commit id: "feat: feat without pr number"
> ```

</details>

[github's syntax to link your pull request]: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#about-linked-issues-and-pull-requests
[developer guide]: https://docs.astarte-platform.org/astarte/snapshot/001-dev_guide.html
[CONTRIBUTING.md]: ../CONTRIBUTING.md
[CODE_OF_CONDUCT.md]: ../CODE_OF_CONDUCT.md

PR #2196, opened 2026-09-10
feat(fdo): Allow specifying device_id on ownership voucher upload

noaccOS:
Register the device as soon as the ownership voucher is loaded, with `nil` credentials_secret.
During to2, the device credentials are created.


libs/astarte_fdo/lib/ownership_voucher/load_request.ex
```diff
@@ -90,6 +97,35 @@ defmodule Astarte.FDO.OwnershipVoucher.LoadRequest do
...
+        add_error(changeset, field, "is not a valid base64 encoded 128 bits id")
+    end
+  end
+
+  def validate_introspection(field, introspection) when is_map(introspection) do
+    Enum.reduce(introspection, [], fn
+      {interface_name, %{"major" => major, "minor" => minor}}, acc
+      when is_integer(major) and is_integer(minor) ->
+        if major < 0 or minor < 0 do
+          [{field, "has negative versions in interface #{interface_name}"}]
```
frossq:
```suggestion
          [{field, "has negative versions in interface #{interface_name}"} | acc]
```
probably

apps/astarte_pairing/lib/astarte_pairing_web/api_spec/schemas/ownership_voucher.ex
```diff
@@ -61,7 +65,7 @@ defmodule Astarte.PairingWeb.ApiSpec.Schemas.OwnershipVoucher do
               description: "Optional PEM-encoded replacement public key."
             }
           },
-          required: [:ownership_voucher, :key_name, :key_algorithm]
+          required: [:hw_id, :ownership_voucher, :key_name, :key_algorithm]
```
frossq, line 68: Are we planning to make mandatory declaring the hw_id when uploading the voucher? No more 'autogeneration' of device id?

libs/astarte_data_access/lib/astarte_data_access/device.ex
```diff
@@ -170,38 +190,23 @@ defmodule Astarte.DataAccess.Device do
...
     fetch_opts = [
       prefix: keyspace,
       consistency: Consistency.device_info(:read),
       error: :device_not_found
     ]
 
-    Repo.delete!(unconfirmed_device, delete_opts)
-    Repo.fetch(Device, device_id, fetch_opts)
+    with {:ok, device} <- Repo.fetch(Device, device_id, fetch_opts) do
+      # Removes TTL from credentials_secret
```
frossq, line 207: Can it happen that the credentials secret have already expired at this point?

libs/astarte_fdo/lib/ownership_voucher/load_request.ex
noaccOS: Fair enough! I moved this over from the pairing's device registration request, but this indeed looks like a bug. Who'd have thought that using reduce is error prone!

noaccOS: this also allows version 0.0 which is not valid

apps/astarte_pairing/lib/astarte_pairing_web/api_spec/schemas/ownership_voucher.ex
noaccOS: exactly. we're making this api similar to our default registration request. api clients and dashboards can have their own ways to autogenerate ids

libs/astarte_data_access/lib/astarte_data_access/device.ex
noaccOS: yes! added the check to see if credentials_secret is `nil`.
also reworked the order of operations to only confirm the device if everything else works

libs/astarte_fdo/lib/ownership_voucher/load_request.ex
```diff
@@ -90,6 +103,76 @@ defmodule Astarte.FDO.OwnershipVoucher.LoadRequest do
     |> decode_replacement_fields()
   end
 
+  defp ensure_voucher_not_already_claimed(%{valid?: false} = changeset, _, _, _), do: changeset
+
+  defp ensure_voucher_not_already_claimed(changeset, realm_name, guid, voucher) do
```
frossq, line 108: Does this address the 'not overwrite a voucher' feature to be implemented in PR https://github.com/astarte-platform/astarte/pull/2189?

frossq (approved):

mizzet1 (approved):

PR #2197, opened 2026-09-11
refactor(fdo): move FDO lib into Umbrella

nedimtokic:
Move FDO library from `libs` directory into Umbrella `apps` directory.

Move config in Umbrella, and delete unused config files.
Adapt CI for new lib path. 
Remove `Astarte.Generators.Interface ` module since it has name collision.


.github/workflows/astarte-libs-workflow.yml
```diff
@@ -52,11 +52,6 @@ jobs:
     with:
       lib: "astarte_generators"
     secrets: inherit
-  astarte_fdo:
```
noaccOS, line 55: we should still run its tests in ci. I think the time has come to restructure the apps workflow like the libs workflow (ie: test everything when we push on apps/**)

noaccOS (changes requested):

.github/workflows/astarte-apps-workflow.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
```
noaccOS, line 29: I think this is missing vmq_plugin?

noaccOS (changes requested):

.github/workflows/astarte-apps-workflow.yaml
nedimtokic: Added now.

noaccOS (approved):

PR #2202, opened 2026-09-15
feat(dashboard): add device id and initial introspection to voucher upload 

mizzet1:
This PR aim to add the missing device id and initial introspection fields  to the dashboard  voucher upload page, which is now required after changes done in  #2196


noaccOS (approved):
I'll wait for Ali's review before merging
also, this should be a dashboard commit not an fdo one

AliouneGaye21 (approved):

PR #2203, opened 2026-09-16
fix(dup): ensure RPC server is always available to clients

noaccOS:
Resolved the issue where a temporary disconnection and reconnection of data_updater_plant to the cluster would make the RPC server inaccessible.


mizzet1 (approved):

nedimtokic (changes requested):
Realm Management mix.lock needs update ?

nedimtokic (approved):

PR #2211, opened 2026-09-18
feat(fdo): bind devices to their ownership voucher guid

mizzet1:
Devices onboarded through FDO are known to the rendezvous server by the GUID in
their ownership voucher, but nothing tied that GUID to the Astarte device.

With this PR:

- On voucher upload the device is registered with the voucher's GUID.
- Deleting a device now deletes its voucher, before the device row itself: a
  missing device then always implies a missing voucher


nedimtokic (approved):

libs/astarte_data_access/lib/astarte_data_access/devices/device.ex
```diff
@@ -48,6 +48,7 @@ defmodule Astarte.DataAccess.Devices.Device do
     field :first_credentials_request, DateTimeMs
     field :first_registration, DateTimeMs
     field :groups, Exandra.Map, key: :string, value: UUID
+    field :guid, UUID
```
noaccOS: at the database level this is a blob not a uuid


```suggestion
    field :guid, :binary
```

libs/astarte_data_access/lib/astarte_data_access/fdo/queries.ex
```diff
@@ -127,11 +127,21 @@ defmodule Astarte.DataAccess.FDO.Queries do
...
   def delete_ownership_voucher(guid) do
     keyspace = Realm.astarte_keyspace_name()
+    consistency = Consistency.device_info(:write)
+    opts = [prefix: keyspace, consistency: consistency]
 
     %OwnershipVoucher{
       guid: guid
     }
-    |> Repo.delete(prefix: keyspace)
+    |> Repo.delete(opts)
```
noaccOS, line 136: at this point we've changed all references to this function to drop the _voucher, what about doing it here directly? that way we also don't need to add the new function


```suggestion
    |> Repo.delete(opts)
    |> case do
      {:ok, _voucher} -> :ok
      error -> error
    end
```

CHANGELOG.md
```diff
@@ -7,10 +7,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+
+- [fdo] Devices are now bound to the `guid` of their ownership voucher when the voucher is
+  uploaded
```
noaccOS: this is not a user facing feature


```suggestion
```

noaccOS (changes requested):
code is mostly good, one nit:
`guid` may be a bit too generic as it's just a global identifier, this field right here is supposed to be the guid for fdo so I'd suggest using naming which better links it to the fdo voucher.

some option would be fdo_guid or ownership_voucher_guid

noaccOS (approved):

PR #2213, opened 2026-09-21
refactor(rpc): move Astarte RPC into Umbrella 

nedimtokic:
Move Astarte RPC library into Umbrella apps


apps/astarte_appengine_api/mix.exs
```diff
@@ -111,7 +111,7 @@ defmodule Astarte.AppEngine.API.Mixfile do
       # Test section
       {:excoveralls, "~> 0.15", only: :test},
       {:mox, "~> 1.3", only: :test},
-      {:mimic, "~> 2.3", only: :test},
+      {:mimic, "~> 2.3", only: [:test, :dev]},
```
eddbbt, line 114: dev?

apps/astarte_fdo/mix.exs
```diff
@@ -67,15 +67,15 @@ defmodule Astarte.FDO.MixProject do
...
       {:cose, github: "secomind/cose-elixir"},
       {:excoveralls, "~> 0.15", only: :test},
       {:dialyxir, "~> 1.4", only: [:dev, :test], runtime: false},
       {:castore, "~> 1.0"},
-      {:mimic, "~> 2.3", only: :test},
-      {:stream_data, "~> 1.1", only: :test},
-      {:astarte_generators, path: "../../libs/astarte_generators", only: :test},
+      {:mimic, "~> 2.3", only: [:test, :dev]},
+      {:stream_data, "~> 1.1", only: [:test, :dev]},
+      {:astarte_generators, path: "../../libs/astarte_generators", only: [:test, :dev]},
```
eddbbt, line 78: same here, why dev

apps/astarte_appengine_api/mix.exs
nedimtokic: Astarte RPC has dependency `{:mimic, "~> 2.3", only: [:test, :dev]}.` Since, we moved Astarte RPC into Umbrella we need to align `mimic` dependency in all services.

apps/astarte_fdo/mix.exs
nedimtokic: Astarte RPC has dependency `{:mimic, "~> 2.3", only: [:test, :dev]}.` Since, we moved Astarte RPC into Umbrella we need to align `mimic` dependency in all services.

noaccOS (approved):

